PR #20, opened 2022-01-19
add .clang-format

pvelesko:


linehill:
Didn’t we settle on using LLVM code style? I’m wondering why the file does not have a single line: `BasedOnStyle: LLVM`.

pvelesko:
@linehill Because clang-tidy doesn't automatically fix everything. For example, it can automatically change parameters from snake_case to CamelCase but then inside the body of the function the old snake_case arguments remain... There's a lot of refactoring to do so my current approach is this:

1. Introduce .clang-tidy rules one-by-one
2. Fix lint errors by hand using symbol reference renaming (I'm using VSCode)
3. Compile code to make sure nothing gets broken
4. Run tests
5. Introduce additional rules

Furthermore, if you just specify LLVM then naming conventions are not enforced as they are described https://llvm.org/docs/CodingStandards.html#style-issues 

Also, there are a lot of issues regarding performance like dead stores, etc so I am focusing solely on naming right now.

pvelesko:
@linehill  I just realised you meant clang-format not clang-tidy. 
To answer your question it's pretty much the same reason - I am making changes in increments. The configuration listed is the one I got from my IDE. It's the same as LLVM style with one exception of pointers being left aligned. 

Personally I find right-aligned pointers to be less readable:
```
void* func(){} // returning a type of void pointer
void *func(){}
```
@linehill I'll change it to LLVM style, but don't you agree that the latter is more readable than the former or is that just me?

This is still in progress and you can assume the same for every PR until I request reviews.

Kerilk:
> Personally I find right-aligned pointers to be less readable:
> 
> ```
> void* func(){} // returning a type of void pointer
> void *func(){}
> ```
> 
> @linehill I'll change it to LLVM style, but don't you agree that the latter is more readable than the former or is that just me?

Well you can write something like this:

`int a, *f(), *b;`
so I tend to right align pointers as `int* a, b;` would be very misleading.

pvelesko:
> Well you can write something like this: int a, *f(), *b;, so I tend to right align pointers as int* a, b; would be very misleading.

Ok this makes sense then. @Kerilk

PR #50, opened 2022-05-05
add graphs placeholders

pvelesko:


samples/CMakeLists.txt
```diff
@@ -91,12 +91,12 @@ set(SAMPLES
...
     11_device
     hipStreamSemantics
-    hipMultiThreadAddCallback
+    #hipMultiThreadAddCallback
     hipInfo
     hipSymbol
     hip-cuda
     printf
-    texture
+    #texture
```
linehill, line 99: This line was commented out in the `add graphs placeholders` patch. Was this intentional?

pvelesko: Yes. It looked to me that these samples are the same as the unit tests since at least one of them had the same name.

pvelesko:
find ./ -name "hipTextureObjFetchVector.cc"
./HIP/tests/catch/unit/texture/hipTextureObjFetchVector.cc
./samples/texture/hipTextureObjFetchVector.cc

Seems like only this file is mirroring a unit test name. I can rename and
re-enable?

linehill:
You can remove the ./samples/texture/hipTextureObjFetchVector.cc one. The test case was borrowed from the HIP-Common. I recall that the HIP-Common test was not added to the catch framework when I did the borrow it.

PR #56, opened 2022-05-24
Implement Non-Immediate Queues

pvelesko:


Kerilk:
@pvelesko Could you describe this change reasoning and and scope?
Thanks

pvelesko:
> @pvelesko Could you describe this change reasoning and and scope?
> 
> Thanks

Bugs in the runtime when using immediate queues.

PR #83, opened 2022-07-26
Refactor CMake - Add hip::host hip::device export targets

pvelesko:
Allows for building HIP codes using clang and CMake without needing to use hipcc.


pvelesko:
@franz merged your changes from main. Please check that this is working/failing as expected

franz:
running `cmake -DLLVM_USE_INTERGRATED_SPIRV=ON` with LLVM 15 produces

```
CMake Error at CMakeLists.txt:104 (message):
  -DLLVM_USE_INTERGRATED_SPIRV=On is not supported for LLVM 13 and less
```

...because the condition is reversed (in merge commit). Swap `if(CLANG_VERSION_LESS_14)` for `if(NOT CLANG_VERSION_LESS_14)` (CMakeLists line 101) and it works. Otherwise seems OK, tests (dont) pass as expected.

PR #98, opened 2022-08-09
Reduce (spurious) warning messages

### LINKED ISSUE #94: Spurious (harmless) warnings?
I get these warnings (with default logging settings) dumped out when building tests:
```
CHIP warning [TID 54662] [1660033951.029736425] : Ignoring alignment. Using hardcoded value 0x1000
CHIP warning [TID 54662] [1660033951.029825009] : Usigned zeMallocHost instead of zeMallocShared due to outstanding bug
CHIP warning [TID 54662] [1660033951.211279355] : Remaining 0 events that haven't been collected:
CHIP warning [TID 54662] [1660033951.211318094] : Remaining 0 command lists that haven't been collected:
CHIP warning [TID 54665] [1660033951.237032246] : Ignoring alignment. Using hardcoded value 0x1000
CHIP warning [TID 54665] [1660033951.237127111] : Usigned zeMallocHost instead of zeMallocShared due to outstanding bug
CHIP warning [TID 54665] [1660033951.441705646] : Remaining 0 events that haven't been collected:
CHIP warning [TID 54665] [1660033951.441741497] : Remaining 0 command lists that haven't been collected:
```

linehill:
... by:

* demoting a "Remaining # events..." messages to debug level as this seems to be an implementation detail which probably is not valuable info for users.
* removing a Level Zero workaround for allocation alignment issue that does not seems to to be around anymore (tested on Intel Compute Runtime 22.30.23789).

Resolves #94.

### DIFF
--- a/src/backend/Level0/CHIPBackendLevel0.cc
+++ b/src/backend/Level0/CHIPBackendLevel0.cc
@@ -1389,8 +1389,6 @@ void *CHIPContextLevel0::allocateImpl(size_t Size, size_t Alignment,
                                       CHIPHostAllocFlags Flags) {
 
   void *Ptr = 0;
-  logWarn("Ignoring alignment. Using hardcoded value 0x1000");
-  Alignment = 0x1000; // TODO Where/why
 
   ze_device_mem_alloc_flags_t DeviceFlags =
       ZE_DEVICE_MEM_ALLOC_FLAG_BIAS_CACHED;
--- a/src/backend/Level0/CHIPBackendLevel0.cc
+++ b/src/backend/Level0/CHIPBackendLevel0.cc
@@ -1240,15 +1240,17 @@ void CHIPBackendLevel0::uninitialize() {
   StaleEventMonitor->Stop = true;
   StaleEventMonitor->join();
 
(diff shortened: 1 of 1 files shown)


pvelesko:
`0x1000` is also used in `hipHostMalloc` : 

````
hipError_t hipHostMalloc(void **Ptr, size_t Size, unsigned int Flags) {
  CHIP_TRY
  CHIPInitialize();
  NULLCHECK(Ptr);
  if (Size == 0) {
    *Ptr = nullptr;
    RETURN(hipSuccess);
  }

  auto FlagsParsed = CHIPHostAllocFlags(Flags);

  void *RetVal = Backend->getActiveContext()->allocate(
      Size, 0x1000, hipMemoryType::hipMemoryTypeHost, FlagsParsed);
  ERROR_IF((RetVal == nullptr), hipErrorMemoryAllocation);

  *Ptr = RetVal;
  RETURN(hipSuccess);
  CHIP_CATCH
}
````

Looks strange it being just a raw value. Maybe we should add some sort of a CMake define?

LGTM

pvelesko (approved):

PR #103, opened 2022-08-12
Attempt to fix #101

### LINKED ISSUE #101: samples/hip-cuda/FloydWarshall uses a lot of memory with Level0 backend
seems like a memory leak of some kind, the test uses >20G memory with Level0 backend, with OpenCL backend it runs in ~135M or so.

linehill:
I saw that CHIPStaleEventMonitorLevel0::monitor() was not cleaning up
stale command list (aka. they are memory leaked) which I suspected to
be the cause for memory hogging in tests like FloydWarshall. Attempted
to fix the issue by:

* changing event call method get() -> peek() in
  CHIPQueueLevel0::executeCommandList to "fix" reference counting. I
  saw in the logs that reference counter was incremented two times but
  only ever decreased by one once.
* resetting event state when the reference count decresed to zero so
  they could be reused. Without this the recycled events were not put
  into Backend->Events lists so the monitor never picked them up.

Also, fixed CHIPStaleEventMonitorLevel0::monitor() not to delete used
events which are belonging to event pools. Apparently, LZEventPool
manages them.

With the fixes above, the memory hogging issues are cleared in the
FloydWarshall and BitonicSort test cases at least on my
machine. However, other problems pops out. I see following tests
failing occasionally:

* FastWalshTransform
* hip_async_binomial
* cuda-bandwidthTest
* cuda-FDTD3d

The failures seems to pop up more often when the tests are ran without
debug printing.

### DIFF
--- a/src/CHIPBackend.hh
+++ b/src/CHIPBackend.hh
@@ -506,9 +506,8 @@ class CHIPContext;
 class CHIPDevice;
 
 class CHIPEvent {
-  public:
-  bool TrackCalled = false;
 protected:
+  bool TrackCalled_ = false;
   event_status_e EventStatus_;
   CHIPEventFlags Flags_;
   std::vector<CHIPEvent *> DependsOnList;
@@ -538,14 +537,7 @@ public:
     }
   }
   void trackImpl();
-  void track() {
-    std::lock_guard<std::mutex> Lock(Mtx);
-    //std::call_once(TrackCalled, &CHIPEvent::trackImpl, this);
-    if (!TrackCalled) {
-      trackImpl();
-      TrackCalled = true;
-    }
-  }
+  void track();
   CHIPEventFlags getFlags() { return Flags_; }
   std::mutex Mtx;
   std::string Msg;
@@ -556,11 +548,6 @@ public:
              (void *)this, Msg.c_str(), *Refc_, *Refc_ - 1, Reason);
     if (*Refc_ > 0) {
       (*Refc_)--;
-      if (*Refc_ == 0) {
-        Msg = "INVALID";
-        TrackCalled = false;
-        EventStatus_ = EVENT_STATUS_INIT;
-      }
     } else {
       logError("CHIPEvent::decreaseRefCount() called when refc == 0");
     }
--- a/src/CHIPBackend.hh
+++ b/src/CHIPBackend.hh
@@ -507,7 +507,7 @@ class CHIPDevice;
 
 class CHIPEvent {
 protected:
-  std::once_flag TrackCalled;
+  bool TrackCalled = false;
   event_status_e EventStatus_;
   CHIPEventFlags Flags_;
   std::vector<CHIPEvent *> DependsOnList;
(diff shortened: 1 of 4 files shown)


src/CHIPBackend.hh
```diff
@@ -548,6 +555,11 @@ public:
              (void *)this, Msg.c_str(), *Refc_, *Refc_ - 1, Reason);
     if (*Refc_ > 0) {
       (*Refc_)--;
+      if (*Refc_ == 0) {
+        Msg = "INVALID";
```
pvelesko, line 559: Might as well make this an assert since refcount should never be 0

src/backend/Level0/CHIPBackendLevel0.cc
```diff
@@ -555,7 +556,8 @@ void CHIPStaleEventMonitorLevel0::monitor() {
         // Add the most course-grain lock here in case event destructor is not
         // thread-safe
         std::lock_guard Lock(Backend->Mtx);
-        delete E;
+        if (!E->EventPool) // Delete event if not owned by an event pool.
+          delete E;
```
pvelesko, line 560: This should probably go away entirely. Event creation/destruction is managed by pools now. all other events are managed by the user

src/backend/Level0/CHIPBackendLevel0.cc
```diff
@@ -1119,12 +1121,13 @@ void CHIPQueueLevel0::executeCommandList(ze_command_list_handle_t CommandList) {
...
 
+    logDebug("assoc event {} w/ cmdlist", (void *)LastCmdListEvent);
     ((CHIPBackendLevel0 *)Backend)
         ->EventCommandListMap[(CHIPEventLevel0 *)LastCmdListEvent] =
         CommandList;
 
-    Status = zeCommandListAppendBarrier(CommandList, LastCmdListEvent->get(), 0,
-                                        nullptr);
+    Status = zeCommandListAppendBarrier(CommandList, LastCmdListEvent->peek(),
+                                        0, nullptr);
```
pvelesko, line 1130: If you change this from `get()` to `peek()` the refcount for this event is 0 when we begin tracking it. Once the event fires, it will try to decrement it from 0 to -1 which is bad.

src/CHIPBackend.hh
```diff
@@ -537,7 +537,14 @@ public:
...
   void trackImpl();
-  void track() { std::call_once(TrackCalled, &CHIPEvent::trackImpl, this); }
+  void track() {
+    std::lock_guard<std::mutex> Lock(Mtx);
+    //std::call_once(TrackCalled, &CHIPEvent::trackImpl, this);
+    if (!TrackCalled) {
+      trackImpl();
+      TrackCalled = true;
+    }
+  }
```
pvelesko, line 547: Right, this change was necessary when we moved to reusing events. Probably what was happening was that events would get recycled and not get tracked due to `call_once` protection

pvelesko (commented):
I'll take a look at this today.

pvelesko:
@linehill please check

linehill:
`CHIP_BE=level0 ctest --timeout 120` gives:

```
The following tests FAILED:
	 65 - cuda-simpleCallback (Subprocess aborted)
	 73 - cuda-qrng (Failed)
```

And with `CHIP_BE=level0 ctest --timeout 120 -j8` following tests occasionally fail:
* stream (SEGFAULT)
* PrintfDynamic (Failed)
* FloydWarshall (Timeout)

Beside the above I'm not seeing gigabytes of memory hogging.

linehill:
The cuda-simpleCallback is also failing on OpenCL. The test is multithreaded and there is already a disabled multithreaded callback HIP test so we probably want to disable this as well.

franz:
@linehill  yeah i have already disabled `cuda-simpleCallback`, it will be in upcoming PR.

pvelesko:
The only samples that fail for me are: 
```
	346 - activemask (Failed)
	415 - cuda-simpleCallback (Timeout)
	423 - cuda-qrng (Failed)
```

the callback issue is known - I am working on it.

PR #106, opened 2022-08-12
Resolve multi-thread callbacks

pvelesko:


pvelesko:
Using this test (modified to run on 32 threads instead of 1000 for debug pursposes) 

```
/home/pvelesko/space/CHIP-SPV/build/catch/unit/stream/hipStreamACb_MultiThread Unit_hipStreamAddCallback_MultipleThreads
```

it currently either passes, segfaults, or sometimes hangs. At 1000 threads it seems to just hang. Not sure what is going on

pvelesko:
image

at 32 threads, seems to work just segfault at uninit.  
at 1000 threads no hangs but the segfault above

PR #108, opened 2022-08-15
Fix to hipcc when used from build directory

linehill:
Fix hipcc fails to compile HIP programs for CHIP-SPV when the compiler is used from the build directory and CHIP-SPV installation step has not been run.

Fixed by creating separate versions of .hipInfo for build and install directory.


franz:
Tested, works from build directory. Thanks @linehill

PR #111, opened 2022-08-16
Small README.md updates + a flaky_tests file

pjaaskel:


README.md
```diff
@@ -118,6 +99,7 @@ Compiling a HIP application with CHIP-SPV will allow you to execute HIP code on
 You can find various HIP applications here: <https://github.com/CHIP-SPV/hip-testsuite>
 
 ```bash
+export HIP_PLATFORM=spirv
```
linehill, line 102: Should not be needed. `hipcc` looks for `$HIP_PATH/lib/.hipInfo` which defines `HIP_PLATFORM=spirv` and other important variables. 

(If HIP_PATH is not defined it defaults to <path-to-hipcc>/..).

franz:
@pjaaskel I've added your README changes to my PR. Please push directly into that branch (& notify me) if you have more documentation updates. WRT the issue not requiring both backends, there's a fix in that PR. WRT the flaky tests, `cat ../flaky_tests` assumes that the build is subdir of source. I've instead added a custom target named "check" (like in pocl) that skips failing tests with a regexp. If you run into more failing tests, please add them to the regexp (i'll add some more too).

PR #120, opened 2022-08-23
Implement support for Priority Queues

pvelesko:
* OpenCL breaks correctness when enabled
* OpenCL complains about the extension pragma not being supported?
* Level Zero works but doesn't seem to pass verification tests using events. There might be a bug in the stream semantics as `hipEvent` unit test also fails sometimes.


pvelesko:
Level Zero passes all unit tests on Iris but Arcticus fails to verify with events

pvelesko:
OpenCL passes completely, Level Zero still sometimes fails most likely due to #128

PR #130, opened 2022-08-31
Release Candidate 0.9

pvelesko:
@franz I followed your release checklist:

Created a commit bumping the version to 1.0-pre and merged into main.
Pulled main, `git checkout HEAD~1` followed by `git checkout -b Release-0.9` and made this PR the purpose of which is to include any last changes necessary for a stable release while the main branch is now pointing to 1.0-pre

Let me know if this looks good


pvelesko:
@franz @pjaaskel @Kerilk @linehill 
All of the non-flaky tests pass on Iris and Arcticus.
Out of the 709 tests, 670 are non-flaky giving us a pass rate of 95%.

The tests that do fail have been documented on the wiki https://github.com/CHIP-SPV/chip-spv/wiki/Release-testing-of-chip-spv-0-9 

modules were created on JLSE 
```
module use /home/pvelesko/local/modulefiles
module load HIP/clang14/chip-spv-0.9 # RelWithDebInfo
module load HIP/clang14/chip-spv-0.9-debug #Debug
```
Using these modules HIP-Examples were successfully compiled.

CMakeLists.txt
```diff
@@ -508,10 +488,10 @@ install(
...
 
 # Generate .hipInfo
-file(WRITE "${PROJECT_BINARY_DIR}/lib/.hipInfo_install" ${_hipInfo_install})
-file(WRITE "${PROJECT_BINARY_DIR}/lib/.hipInfo" ${_hipInfo_build})
-install(FILES ${PROJECT_BINARY_DIR}/lib/.hipInfo_install
-  DESTINATION ${LIB_INSTALL_DIR} RENAME .hipInfo)
+file(WRITE "${PROJECT_BINARY_DIR}/bin/.hipInfo_install" ${_hipInfo_install})
+file(WRITE "${PROJECT_BINARY_DIR}/bin/.hipInfo" ${_hipInfo_build})
+install(FILES ${PROJECT_BINARY_DIR}/bin/.hipInfo_install
+  DESTINATION ${BIN_INSTALL_DIR} RENAME .hipInfo)
```
linehill, line 494: .hipInfo should be placed under <install-dir>/lib/. HIPAMD install their hipInfo in there and hippcc reads the file from it.

linehill: How do we know there are no other users that rely on the on the .hipInfo to be found under lib/ directory?

pvelesko: .hipInfo is only used by hipcc and hipconfig where I made the adjustment as well. As soon as we merge this Release, I will  merge the HIPCC branch which makes the switch from Perl based hipcc to C++ based hipcc. The C++ hipcc does not use the .hipInfo for AMD/NVIDIA paths so we are free to put it anywhere we want and it makes more sense to have it in the same dir the as the hipcc executable.

linehill: If you meant that the .hipInfo has been private to the hipcc and hipconfig then it’s fine to relocate the file.

> ... it makes more sense to have it in the same dir the as the hipcc executable.

Usually regular files are not installed under bin/ and lib/. Perhaps, its place is under share/?

pvelesko: > Perhaps, its place is under share/?

Agree.

PR #145, opened 2022-09-19
Fix #144

### LINKED ISSUE #144: Unable to build with clang-15 on Ubuntu 20.04 due to dependency on `icpx`
I'm trying to build chip-spv on my Intel laptop (Core i7-1165G7, Ubuntu 20.04). I've built the spirv-llvm-translator using clang-15 and it seems to work. However, while building chip-spv, it fails with
```
[ 61%] Building CXX object samples/sycl_hip_interop/sycl_hip_interop_driver/CMakeFiles/sycl_chip_interop.dir/sycl_chip_interop.cpp.o
cd /home/v99/sources/chip-spv/build/samples/sycl_hip_interop/sycl_hip_interop_driver && icpx   -Wno-format-extra-args -fsycl -Wno-deprecated-declarations -std=c++17 -MD -MT samples/sycl_hip_interop/sycl_hip_interop_driver/CMakeFiles/sycl_chip_interop.dir/sycl_chip_interop.cpp.o -MF CMakeFiles/sycl_chip_interop.dir/sycl_chip_interop.cpp.o.d -o CMakeFiles/sycl_chip_interop.dir/sycl_chip_interop.cpp.o -c /home/v99/sources/chip-spv/samples/sycl_hip_interop/sycl_hip_interop_driver/sycl_chip_interop.cpp
/bin/sh: 1: icpx: not found
```
I tried to source my `/opt/intel/oneapi/setvars.sh`, but then it fails because of non-matching clang versions (Intel OneAPI 2022 appears to be using LLVM/Clang 11).
```
Unknown attribute kind (70) (Producer: 'LLVM15.0.1' Reader: 'LLVM 11.0.0')
clang: error: hipspv-link command failed with exit code 1 (use -v to see invocation)
```
Is there some workaround for this? Is icpx necessary for building/using chip-spv?

pvelesko:
* find_library for OpenCL and Level Zero would cache by default so if you run CMake without oneapi in your path it would fail to find OpenCL/LeveZero and cache that value. If you load oneapi then it would use the cached value and fail. Fixed by forcing find_library to `NO_CACHE` 
* If no modules are loaded and oneapi is installed in the default location, we would find icpx and enable the compilation of icpx dependent samples. For these samples, we would assume that icpx is in PATH. Fixed by using absolute path for icpx
* Related to previous error, various dependent library paths would not be found during link time. Fixed by extracting these paths during the search for icpx.


pjaaskel:
Looks OK. I'm not sure how to test this though. Can you make a PR against the release branch as this should fix things?

pvelesko:
>  Can you make a PR against the release branch as this should fix things?

That branch was merged into main and the PR was automatically closed. When I re-opened the PR you said that it was not necessary - "There is no need for a PR here in this use case for release branches." 

.. So if we merge this fix into the Release-0.9 branch then how will it make its way into main?

pjaaskel:
I meant PRs for merging the release branches to main in my comment. Those are unnecessary if we just merge them with git (not github) occasionally.

So, fixes from release branches end up in main via 'git merge'. In this case I asked to send a PR to merge it to the release branch as it's a fix.

PR #154, opened 2022-09-22
Fix #153

### LINKED ISSUE #153: Two samples are failing
Observed two failures while running "**make check**" on my setup. My configuration is ATS + intel_compute_runtime_475 + clang 14

        578 - hip_async_binomial (SEGFAULT)
        597 - hipNormalizedFloat (Failed)

While running it standalone, below is the failure output observed for hipNormalizedFloat. Looked further and found data type 'char' and 'short' subtests are failing with exact same indexes not matching every time. If I comment out char and short cases rest of the test passes without any issue.

**Failed output**
<my_server>:~/src/chip-spv/build/samples/texture> ./hipNormalizedFloat
Error at [0][5]: Expected '-0.566929'. Got '-0.566929'
Error at [3][2]: Expected '-0.818898'. Got '-0.818898'
Error at [6][6]: Expected '-0.283465'. Got '-0.283465'
Error at [7][2]: Expected '0.818898'. Got '0.818898'

pvelesko:


src/CHIPBackend.hh
```diff
@@ -1536,6 +1536,21 @@ public:
...
+
+  /**
+   * @brief Get the smallest max allocation size for this context by checking
+   * the max allocation size across all devices in this context.
+   *
+   * @return size_t
+   */
+  size_t getMaxAllocSize() {
+    size_t AllocSize = std::numeric_limits<size_t>::max();
+    for (auto Dev : getDevices()) {
```
Sarbojit2019, line 1548: Why do you need to scan through all the device? I would assume current device's max limit should be enough to check.

pvelesko: Because allocations take place inside a context and a context can be created for multiple devices. I assume it's possible that these devices have different properties? 

```
  // Load devices to device vector
  Status = zeDeviceGet(ZeDriver, &DeviceCount, nullptr);
  CHIPERR_CHECK_LOG_AND_THROW(Status, ZE_RESULT_SUCCESS, hipErrorTbd);
  ZeDevices.resize(DeviceCount);
  Status = zeDeviceGet(ZeDriver, &DeviceCount, ZeDevices.data());
  CHIPERR_CHECK_LOG_AND_THROW(Status, ZE_RESULT_SUCCESS, hipErrorTbd);

  const ze_context_desc_t CtxDesc = {ZE_STRUCTURE_TYPE_CONTEXT_DESC, nullptr,
                                     0};

  ze_context_handle_t ZeCtx;
  Status = zeContextCreateEx(ZeDriver, &CtxDesc, DeviceCount, ZeDevices.data(),
                             &ZeCtx);
  CHIPERR_CHECK_LOG_AND_THROW(Status, ZE_RESULT_SUCCESS, hipErrorTbd);
```

PR #160, opened 2022-09-27
Implement hiprtc API for basic runtime compilation cases

linehill:
Implement hiprtc API for basic runtime compilation cases

Supported:

* Invoking extern "C" `__global__` kernels.

Unsupported or missing features:

* Compile options (currently ignored).
* Invoking non-extern "C" `__global__` kernels (required `hiprtcAddNameExpression()` and `hiprtcGetLoweredName()` functions are not implemented yet).

### DIFF
--- a/include/hip/spirv_hip_fp16.h
+++ b/include/hip/spirv_hip_fp16.h
@@ -39,8 +39,8 @@ THE SOFTWARE.
 typedef _Float16 _hip_f16;
 typedef _Float16 _hip_f16_2 __attribute__((ext_vector_type(2)));
 #else
-typedef api_half _hip_f16;
-typedef api_half2 _hip_f16_2;
+typedef short _hip_f16;
+typedef short _hip_f16_2 __attribute__((ext_vector_type(2)));
 #endif
 
 struct __half_raw {
--- a/src/CHIPBackend.cc
+++ b/src/CHIPBackend.cc
@@ -1451,17 +1451,6 @@ CHIPQueue *CHIPBackend::findQueue(CHIPQueue *ChipQueue) {
   return *QueueFound;
 }
 
-bool CHIPBackend::eraseProgram(const CHIPProgram *ToErase) {
-  auto ErasePos =
-      std::remove_if(ChipPrograms_.begin(), ChipPrograms_.end(),
-                     [&](const std::unique_ptr<CHIPProgram> &Prog) -> bool {
-                       return Prog.get() == ToErase;
-                     });
-  bool WasErased = ErasePos != ChipPrograms_.end();
-  ChipPrograms_.erase(ErasePos, ChipPrograms_.end());
-  return WasErased;
-}
-
 // CHIPQueue
 //*************************************************************************************
 CHIPQueue::CHIPQueue(CHIPDevice *ChipDevice, CHIPQueueFlags Flags, int Priority)
--- a/src/CHIPBackend.cc
+++ b/src/CHIPBackend.cc
@@ -51,7 +51,7 @@ static void queueVariableInfoShadowKernel(CHIPQueue *Q, CHIPModule *M,
                                           const CHIPDeviceVar *Var,
                                           void *InfoBuffer) {
   assert(M && Var && InfoBuffer);
-  auto *K = M->getKernel(std::string(ChipVarInfoPrefix) + Var->getName());
+  auto *K = M->getKernelByName(std::string(ChipVarInfoPrefix) + Var->getName());
   assert(K && "Module is missing a shadow kernel?");
   void *Args[] = {&InfoBuffer};
   queueKernel(Q, K, Args);
@@ -62,7 +62,7 @@ static void queueVariableBindShadowKernel(CHIPQueue *Q, CHIPModule *M,
   assert(M && Var);
   auto *DevPtr = Var->getDevAddr();
   assert(DevPtr && "Space has not be allocated for a variable.");
-  auto *K = M->getKernel(std::string(ChipVarBindPrefix) + Var->getName());
+  auto *K = M->getKernelByName(std::string(ChipVarBindPrefix) + Var->getName());
   assert(K && "Module is missing a shadow kernel?");
   void *Args[] = {&DevPtr};
   queueKernel(Q, K, Args);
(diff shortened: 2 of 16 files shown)


pjaaskel (approved):
Great work, Henry. Does this enable some tests to the test suite?

linehill:
> Does this enable some tests to the test suite?

A test is enabled by this commit: 16b2712.

pjaaskel:
@pvelesko is this OK to push? Does it suffice for libceed?

pvelesko:
I will review today or tomorrow - working on HIP submodule refactoring that's required for CHIP-SPV integration into HIPCOMMON.

pvelesko:
```
/home/pvelesko/space/CHIP-SPV/src/Utils.cc:27:10: fatal error: 'filesystem' file not found
```

I need to replace this with `experimental::filesystem` to get past compilation. @linehill

linehill:
<filesystem> should be provided by C++17 standard. AFAIK, we don't require the standard but the root CMakeLists.txt defaults to it:
```
set(CMAKE_CXX_STANDARD 17 CACHE STRING "The C++ standard to use.")
set(CMAKE_CXX_STANDARD_REQUIRED ON)
```

pvelesko:
Can you push the fix? @linehill

I'm in the process of replacing fs with experimental fs but still running into linking issues

pvelesko:
Not sure what you mean by that? I'm using the unmodified CMakeLists.txt

linehill:
<filesystem> is part of C++17. Assuming CMake compiles the CHIP-SPV with c++17 the header should be found. Could you investigate why it doesn't find it in your system?

pvelesko:
> Could you investigate why it doesn't find it in your system?

Ok I'll look into that.

linehill:
Could be caused by old libstdc++ which lacks the header. Clang does not provide it and uses what is found in the system instead, I think.

pvelesko:
Can we merge the fix then? HIPCC uses std::experimental::fs as well so this wouldn't introduce anything new.

linehill:
The fix may fail on newer standard library not having the <experimental/filesystem> header anymore. HIPCC also uses <filesystem> if the system provides it. I'm working on an alternate fix candidate.

src/CHIPBackend.cc
```diff
@@ -1445,6 +1450,17 @@ CHIPQueue *CHIPBackend::findQueue(CHIPQueue *ChipQueue) {
...
 }
 
+bool CHIPBackend::eraseProgram(const CHIPProgram *ToErase) {
+  auto ErasePos =
+      std::remove_if(ChipPrograms_.begin(), ChipPrograms_.end(),
+                     [&](const std::unique_ptr<CHIPProgram> &Prog) -> bool {
+                       return Prog.get() == ToErase;
+                     });
+  bool WasErased = ErasePos != ChipPrograms_.end();
+  ChipPrograms_.erase(ErasePos, ChipPrograms_.end());
```
pvelesko, line 1460: thread safe?

src/CHIPBackend.hh
```diff
@@ -1850,6 +1886,12 @@ public:
   /* event interop */
   virtual hipEvent_t getHipEvent(void *NativeEvent) = 0;
   virtual void *getNativeEvent(hipEvent_t HipEvent) = 0;
+
+  void addProgram(std::unique_ptr<CHIPProgram> Program) {
+    ChipPrograms_.push_back(std::move(Program));
```
pvelesko, line 1891: thread safe?

include/hip/spirv_hip_fp16.h
```diff
@@ -39,8 +39,8 @@ THE SOFTWARE.
 typedef _Float16 _hip_f16;
 typedef _Float16 _hip_f16_2 __attribute__((ext_vector_type(2)));
 #else
-typedef api_half _hip_f16;
-typedef api_half2 _hip_f16_2;
+typedef short _hip_f16;
+typedef short _hip_f16_2 __attribute__((ext_vector_type(2)));
```
pvelesko, line 43: I was confused about this part.. Where does api_half come from?

src/CHIPBackend.cc
```diff
@@ -737,6 +734,14 @@ void CHIPDevice::addModule(const std::string *ModuleStr, CHIPModule *Module) {
   this->ChipModules[ModuleStr] = Module;
 }
 
+void CHIPDevice::eraseModule(CHIPModule *Module) {
+  auto *ModSrc = Module->getModuleSource();
+  if (auto It = ChipModules.find(ModSrc); It != ChipModules.end()) {
+    delete It->second;
+    ChipModules.erase(It);
```
pvelesko, line 741: thread safe?

include/hip/spirv_hip_fp16.h
linehill: No idea where it was supposed to come from but it not defined anywhere, AFAIK.

src/CHIPBackend.cc
linehill: I have a follow up PR for removing this. I'll transfer it here.

src/CHIPBackend.hh
linehill: I'll remove this.

src/Utils.hh
```diff
@@ -24,9 +24,56 @@
 #ifndef SRC_UTILS_HH
 #define SRC_UTILS_HH
 
+#include "common.hh"
+
+#include "hip/hip_fatbin.h"
+
+#include <filesystem>
+#include <optional>
```
pvelesko, line 32: unnecessary if included in common.hh

pvelesko:
Nice work! I think we should merge #182 first as that will enable a lot more hipRTC tests.

linehill:
I'll rebase this patch when it's merged.

src/CHIPBackend.cc
linehill: Added device scope lock.

src/Utils.hh
linehill: Filesystem include is gone. Didn't see optional included in common.h.

linehill:
Updated. The patch tries to use `<experimental/filesystem>` header instead if the `<filesystem>` header is not available.

pvelesko:
Merged the HIPCOMMON rebase. Please checkout HIP/test/README there's a new section about hipRTC testing

pvelesko:
Still waiting on feedback from AMD and the rebased branch needs some additional work so go ahead and merge this when you are ready.

linehill:
Alright, thanks.

PR #172, opened 2022-09-30
Fix a test failed after switching to C++-based hipcc

linehill:
This fix requires HIPCC submodule to be updated with fixes from https://github.com/CHIP-SPV/HIPCC/pull/1.


linehill:
Oops, didn't check where the fix was needed. Well, this goes for the next milestone.

PR #174, opened 2022-10-01
Stream Semantics test

Sarbojit2019:
Added stream semantics test to identify streams can run independently.


pvelesko:
@Sarbojit2019 missing the actual test. Also, no need to add a new test - just the hipStreamSemantics test.

Sarbojit2019:
@pvelesko, I have updated the right test please take a look.

pvelesko:
Unit test passing.

Sarbojit2019:
@pvelesko, I have added few more scenarios. I have verified with the fix but majority still fails.

PR #179, opened 2022-10-04
Fix hipMemset behavior with host malloc'd pointers

pjaaskel:
AFAIU it should not just go and fill the memory as asked in that case, but return an error if it's not an UM-mapped pointer.

Fixes Unit_hipMemset_InvalidPtrTests for OpenCL.

@pvelesko  can you test with the Aurora envs and push this in if works and looks OK? Thanks!


pvelesko:
fixed formatting + rebase on main

Issue #222, opened 2022-10-26
Implementation of Math intrinsics is not portable


pjaaskel:
The current implementation maps [CUDA intrinsics](https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH__INTRINSIC__SINGLE.html#group__CUDA__MATH__INTRINSIC__SINGLE) to OpenCL [native_ functions](https://registry.khronos.org/OpenCL/sdk/1.0/docs/man/xhtml/mathFunctions.html).

The CUDA intrinsics' accuracy is [well defined](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#intrinsic-functions), whereas the OpenCL native functions can be pretty much whatever and the accuracy of the implementation cannot be queried from the device. Calling the native functions and assuming CUDA accuracy when doing so makes the generated SPIR-V inherently non-portable. The native functions could just return zero or even rand() and be compliant as far as I understand.

This problem can be demonstrated with the cuda-blackscholes test case (#142): It works (by good enough implementation accuracy) with the iGPU, but fails with the CPU. When I force it to use an accurate version of fexp, it works similarly with GPU and CPU.

This means, unfortunately, that we cannot use the native versions for implementing the intrinsics by default. I'm not sure what the best fix would be to gain the performance benefits of the faster "native" intrinsics when it's safe to do so. Probably via an SPIR-V extension that can be used to declare/demand at least similar accuracy that CUDA defines to the "native" versions or such.

pvelesko:
Could this be solved by using OCML instead of native opencl functions?

bashbaug:
I don't think it makes sense to try to bound the accuracy of the "native" built-in functions.  The "native" functions are intended to map as closely as possible to underlying hardware instructions, so by definition their error bounds will be implementation-defined.

I do think it would be interesting to consider another class of built-in functions, say "fast" built-in functions, with defined error bounds, but not necessarily a 1:1 mapping to native hardware instructions.

[...]

PR #213, opened 2022-10-24
Documentation update + half2 test + half/half2 intrinsics fixes

franz:

PR #223, opened 2022-10-27
Correctness first with the CUDA math intrinsics

pjaaskel:
OpenCL native_* functions have no accuracy quarantees whatsoever whereas the __* math intrinsics of CUDA do. Thus, we cannot use the native functions by default if we want to retain portable correctness in the produced SPIR-Vs.

Fixes cuda-blackscholes for OpenCL/CPU. See Issue #222.

While fixing this, I noticed the builtin library implementation looks a bit messy and has similar codes in two places. What is the idea and longer term plan with it @pvelesko?

### DIFF
--- a/bitcode/devicelib.cl
+++ b/bitcode/devicelib.cl
@@ -113,6 +113,14 @@ EXPORT int CL_NAME2(NAME, d)(double x) { return NAME(x); } \
 EXPORT int CL_NAME2(NAME, h)(half x) { return NAME(x); }
 //EXPORT int CL_NAME(NAME, h2)(half2 x) { return NAME(x); }
 
+
+#ifdef CHIP_NONPORTABLE_MATH_INTRISINCS
+
+// The OpenCL native_* functions have no accuracy quarantees whatsoever
+// whereas the __* math intrinsics of CUDA do. Thus, we cannot use the
+// native functions by default if we want to retain portable correctness
+// in the produced SPIR-Vs.
+
 #define DEF_OPENCL1F_NATIVE(NAME) \
 float OVLD native_##NAME(float x); \
 double OVLD native_##NAME(double x); \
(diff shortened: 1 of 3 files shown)


pvelesko:
> While fixing this, I noticed the builtin library implementation looks a bit messy and has similar codes in two places. What is the idea and longer term plan with it @pvelesko?
 
One of my milestones for next year is to get rid of OCML definitions of our own and to switch to using ROCm-Device-Libs.

include/hip/spirv_hip_devicelib.hh
```diff
@@ -520,9 +520,22 @@ DEFOPENCL1F_NATIVE(log)
...
-DEFOPENCL2F_NATIVE(powr)
+
+#else
+
+EXPORT float OVLD __recip(float x) { return 1.0f / x; }
+EXPORT double OVLD __recip(double x) { return 1.0 / x; }
+
+EXPORT float OVLD __dividef(float x, float y) { return x / y; }
+EXPORT double OVLD __divided(double x, double y) { return x / y; }
+
```
pvelesko, line 537: These should go and replace their respective placeholder under `CHIP-SPV/include/hip/devicelib/` 

I don't see these particular intrinsics defined under CUDA math API, however:

https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH__INTRINSIC__SINGLE.html#group__CUDA__MATH__INTRINSIC__SINGLE 

Nor do I see them defined under HIP instrisics:

https://github.com/ROCm-Developer-Tools/HIP/blob/develop/docs/markdown/hip_kernel_language.md#floating-point-intrinsics

include/hip/spirv_hip_devicelib.hh
```diff
@@ -520,9 +520,22 @@ DEFOPENCL1F_NATIVE(log)
 DEFOPENCL1F_NATIVE(recip)
 DEFOPENCL1F_NATIVE(sqrt)
 DEFOPENCL1F_NATIVE(rsqrt)
+DEFOPENCL2F_NATIVE(powr)
```
pvelesko, line 523: These should go and replace their respective placeholder under CHIP-SPV/include/hip/devicelib/

I don't see these particular intrinsics defined under CUDA math API, however:

https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH__INTRINSIC__SINGLE.html#group__CUDA__MATH__INTRINSIC__SINGLE

Nor do I see them defined under HIP instrisics:

https://github.com/ROCm-Developer-Tools/HIP/blob/develop/docs/markdown/hip_kernel_language.md#floating-point-intrinsics

bitcode/devicelib.cl
```diff
@@ -125,6 +133,22 @@ double OVLD native_##NAME(double x, double y); \
...
+EXPORT float CL_NAME2(NAME##_native, f)(float x) { return NAME(x); } \
+EXPORT double CL_NAME2(NAME##_native, d)(double x) { return NAME(x); }
+
+#define DEF_OPENCL2F_NATIVE(NAME) \
+float OVLD native_##NAME(float x, float y); \
+double OVLD native_##NAME(double x, double y); \
+EXPORT float CL_NAME2(NAME##_native, f)(float x, float y) { return NAME(x, y); } \
+EXPORT double CL_NAME2(NAME##_native, d)(double x, double y) { return NAME(x, y); }
+
+#endif
```
pvelesko, line 150: `devicelib.cl` contains implementations whereas all the mappings take place under hip/include/devicelib

so it would make more sense to me to have both the portable and non-portable implementations and then use the `CHIP_NONPORTABLE_MATH_INTRISINCS` define to map to the desired device function.

pvelesko (changes requested):
Please refer to https://github.com/CHIP-SPV/chip-spv/pull/213

pjaaskel:
I rebased and tried to address your concerns here. I'm not still 100% confident I understand the builtin implementation's what-goes-where.

pvelesko:
@pjaaskel The milestone after my current one deals contains a line item to improve integration with ROCm-Device-Libs so let's go ahead and merge with your changes as-is and I will rework everything later.

pvelesko (approved):

PR #239, opened 2022-10-31
Copy the latest AMD HIP fp16 implementation

pjaaskel:
Also sneak in hipGetStreamDeviceId() implementation.

These updates address some of the first issues encountered while trying to build rocPRIM with CHIP-SPV.

### DIFF
--- a/include/hip/spirv_hip_fp16.h
+++ b/include/hip/spirv_hip_fp16.h
@@ -1,5 +1,5 @@
 /*
-Copyright (c) 2015 - present Advanced Micro Devices, Inc. All rights reserved.
+Copyright (c) 2015 - 2021 Advanced Micro Devices, Inc. All rights reserved.
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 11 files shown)


pvelesko (changes requested):
Some lint/format  + failure to build

```
/usr/lib64/gcc/x86_64-suse-linux/7/../../../../x86_64-suse-linux/bin/ld: CMakeFiles/fp16.dir/haxpy-base.cpp.o: in function `__half::operator unsigned short<unsigned short, (void*)0>() const':
/tmp/CHIP-SPV/include/hip/spirv_hip_fp16.h:248: undefined reference to `__extendhfsf2'
clang-15: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [samples/fp16/CMakeFiles/fp16.dir/build.make:100: samples/fp16/fp16] Error 1
make[1]: *** [CMakeFiles/Makefile2:2438: samples/fp16/CMakeFiles/fp16.dir/all] Error 2
```

include/hip/spirv_hip_runtime.h
```diff
@@ -27,6 +27,12 @@ THE SOFTWARE.
 #define __HIP_PLATFORM_SPIRV__
 #endif
 
+#if defined(__clang__) && defined(__HIP__)
+#define __HIP_CLANG_ONLY__ 1
+#else
+#define __HIP_CLANG_ONLY__ 0
+#endif
+
```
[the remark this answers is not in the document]
pvelesko: Interesting... I wonder what the use case is for doing something like that.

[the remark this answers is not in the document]
pjaaskel: No clue. Just brutally trying to force it build without minimal (rocPRIM) changes.

pjaaskel:
I couldn't reproduce your build problem, but I fixed the failing tests (my bad for overlooking). I think it would be good to leave the mostly copied header in the original style so it's easier to diff and synchronize in the future.

pvelesko:
> I couldn't reproduce your build problem

I'll take a look on my end

pjaaskel:
I tested only with OpenCL on my local iGPU BTW, but should be a quite portable change.

pjaaskel:
Rebased.

`Unit_hipGraph_BasicFunctional` passes about 50% of the time, also `Unit_hipStreamCreate_MultistreamBasicFunctionalities` fails (likely after a timeout), but seems unrelated to this PR.

pjaaskel:
This might break using hipcc from installation path. It at least misses the include dir and perhaps some other defines. Need to investigate.

```
hipcc hip-repro.cc -o hip-repro
Searching for .hipInfo in /home/pjaaskel/local/stow/chip-spv/bin/../share
hip-repro.cc:27:10: fatal error: 'hip/hip_runtime.h' file not found
#include "hip/hip_runtime.h"
         ^~~~~~~~~~~~~~~~~~~
1 error generated.

failed to execute:/home/pjaaskel/local/bin/clang++ -L/home/pjaaskel/local/stow/chip-spv/lib -lCHIP -L/usr/lib/x86_64-linux-gnu -lOpenCL -L/usr/lib/x86_64-linux-gnu -lze_loader -Wl,-rpath,/home/pjaaskel/local/stow/chip-spv/lib  hip-repro.cc -o hip-repro
```

pjaaskel:
Nope, seems main is broken similarly. Using hipcc from install path fails with that missing include.

pvelesko:
This would be made redundant after #187 is merged. Can I close this?

pjaaskel:
Sure, if it's redundant.

pjaaskel:
Seems this is not redundant. The HIP common rebase doesn't bring in fp16 to our implementation. I'll rebase.

pvelesko:
> Seems this is not redundant. The HIP common rebase doesn't bring in updated fp16 from hipamd to our implementation. I'll rebase.

The rebase was done on latest develop. Where are you bringing in these changes from?

pjaaskel:
> > Seems this is not redundant. The HIP common rebase doesn't bring in updated fp16 from hipamd to our implementation. I'll rebase.
> 
> The rebase was done on latest develop. Where are you bringing in these changes from?

hipamd

pjaaskel:
What is the status of this?

pvelesko:
@pjaaskel it's good to go. I'm updating some scripts that check for unit test failure resolution so I want to merge that first so I can run those tests on top of this PR to see if this resolves any unit tests. I'll merge today.

include/hip/spirv_hip_fp16.h
```diff
@@ -18,1011 +18,1713 @@ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
...
-#if !defined(__HIP_NO_HALF_CONVERSIONS__)
-  template <typename T, Enable_if_t<std::is_integral<T>{}>* = nullptr>
-  __device__ operator T() const {
-    return data;
-  }
+#define __ocml_log10_f16(__X) log10_h(__X)
+#define __ocml_log10_2f16(__X) log10_2h(__X)
+
+#define __llvm_amdgcn_rcp_f16(__X) ((__half)1.0f / (__half)__X)
+#define __llvm_amdgcn_rcp_2f16(__X) ((__half2)1.0f / (__half2)__X)
```
pvelesko, line 73: Aren't `amdgcn` built-in LLVM intrinsics for AMD? Why do we have any mention of them on the SPIR-V side? @pjaaskel

pjaaskel: Just for source-level compatibility with the roc libraries which call them directly (in this case rocPRIM). IMHO we can just as well add them to avoid needing to redefine them when porting roc/hip libs which refer to them.

include/hip/spirv_hip_fp16.h
```diff
@@ -18,1011 +18,1713 @@ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
...
-  }
-
-// ACCESSORS - DEVICE ONLY
-#if !defined(__HIP_NO_HALF_CONVERSIONS__)
-  template <typename T, Enable_if_t<std::is_integral<T>{}>* = nullptr>
-  __device__ operator T() const {
-    return data;
-  }
+#define __ocml_log10_f16(__X) log10_h(__X)
+#define __ocml_log10_2f16(__X) log10_2h(__X)
```
pvelesko, line 70: What is `log10_h`? I can't find anything by this name besides this reference. 
The closest thing I see is a placeholder in `bfloat16_math.hh` 
```
// __device__​ __nv_bfloat16 	hlog10 ( const __nv_bfloat16 a )
```

is `bfloat16` and `fp16` two different things?

pvelesko: To answer my own question, yes: 

https://nhigham.com/2018/12/03/half-precision-arithmetic-fp16-versus-bfloat16/

pjaaskel: Likely we haven't implemented it. Please bear in mind that this is a direct copy from AMDHIP and we better keep it so (make minimal unnecessary modifications to it, including formatting) in order to be able to easily synch updates from upstream later.

pvelesko: Ok. It doesn't seem right that we need to be copying stuff over from hipamd. I thought the whole idea of splitting HIP-COMMON and HIP-AMD was to avoid stuff like this.

pjaaskel: I definitely agree this should be COMMON material when going forward.  I have another similar one in thepipeline, bf16.

pvelesko: Somewhat relevant answer from Rahul as to why the math headers aren't shared:

```
It was decided to include these device headers as part of each backend for example -[ https://github.com/ROCm-Developer-Tools/hipamd/blob/develop/include/hip/amd_detail/amd_device_functions.h](https://github.com/ROCm-Developer-Tools/hipamd/blob/develop/include/hip/amd_detail/amd_device_functions.h). One of the reason was usage of inline functions.
```

pjaaskel: Why inline functions prevent putting to common?

pvelesko: I don't know.

PR #252, opened 2022-11-08
Fix missing backend uninitialization causes occational crashes

### LINKED ISSUE #228: Occasional hipRTC Failure
```
Total Test time (real) =  11.86 sec

The following tests FAILED:
	  2 - TestHiprtcCPPKernels (SEGFAULT)
```
Had to run it for a while

linehill:
Missing backend uninitialization at program exit is known to cause crashes occasionally due to monitor threads being still in-flight and making references to destructed objects.

* Fix HIP programs without embedded modules (occasionally) crash at program exit.

* Make sure spdlog is destructed after the CHIP backend is uninitialized.

Fixes #228.

### DIFF
--- a/src/CHIPBackend.cc
+++ b/src/CHIPBackend.cc
@@ -1184,7 +1184,11 @@ std::string CHIPBackend::getJitFlags() {
   return Flags;
 }
 
-CHIPBackend::CHIPBackend() { logDebug("CHIPBackend Base Constructor"); };
+CHIPBackend::CHIPBackend() {
+  logDebug("CHIPBackend Base Constructor");
+  Logger = spdlog::default_logger();
+};
+
 CHIPBackend::~CHIPBackend() {
   logDebug("CHIPBackend Destructor. Deleting all pointers.");
   while (!ChipExecStack.empty())
--- a/src/CHIPBackend.hh
+++ b/src/CHIPBackend.hh
@@ -1632,6 +1632,10 @@ protected:
   CHIPContext *ActiveCtx_;
   CHIPDevice *ActiveDev_;
 
+  // Keep hold on the default logger instance to make sure that it is
+  // not destructed before the backend finishes uninitialization.
+  std::shared_ptr<spdlog::logger> Logger;
+
 public:
 #ifdef DUBIOUS_LOCKS
   std::mutex DubiousLockOpenCL;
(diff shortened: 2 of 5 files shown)


pvelesko:
SYCL-HIP interop samples segfault
```
CHIPContextLevel0::~CHIPContextLevel0() {
  logTrace("~CHIPContextLevel0() {}", (void *)this);
  zeContextDestroy(this->ZeCtx); // segfault here
}
``` 

ran into this issue in #250

pvelesko:
@linehill

src/CHIPBackend.hh
```diff
@@ -1769,6 +1773,7 @@ public:
    * @return std::vector<std::string*>&
    */
   std::vector<std::string *> &getModulesStr();
+  const std::vector<std::string *> &getModulesStr() const;
```
pvelesko, line 1776: Why do we need this

src/CHIPBackend.hh
```diff
@@ -1789,6 +1794,10 @@ public:
    * @param mod_str
    */
   void unregisterModuleStr(std::string *ModuleStr);
+  size_t getNumRegisteredModules() const {
```
pvelesko, line 1797: Why make this const? it puts a requirement on making `CHIPBackend::BackendMtx` mutable and adding a const-flavored version of `getModulesStr()`

src/CHIPBackend.cc
```diff
@@ -1346,6 +1350,9 @@ size_t CHIPBackend::getNumDevices() {
   return NumDevices;
 }
 std::vector<std::string *> &CHIPBackend::getModulesStr() { return ModulesStr_; }
+const std::vector<std::string *> &CHIPBackend::getModulesStr() const {
+  return ModulesStr_;
+}
```
pvelesko, line 1355: Seems unnecessary?

pvelesko (approved):

pvelesko:
Rebased on main and a  few comments but looks good.

src/CHIPBackend.cc
linehill: It's for calling the method via `const` qualified pointers and references.

src/CHIPBackend.hh
linehill: I intend to communicate here that the method call does not change the state of the `CHIPBackend` instance which is **enforced by the compiler**. Although the `BackendMtx` is qualified as `mutable` (for enabling locking in this method), its state change is transient - the lock state returns to the original state after the call.

src/CHIPBackend.cc
pvelesko: I meant that why have two of these? Why not get rid of the old non-const function and create a `addModulesStr()` and `removeModulesStr()` 

having a const and a non-const version of the same thing seems a bit redundant/confusing.

linehill: Removed the getModuleStr.

pvelesko:
@linehill next time please let me review before merging.

Issue #258, opened 2022-11-17
hip_sycl_interop crashes if CHIP_BE is not set


pvelesko:
It should default to something instead of segfaulting

Issue #253, opened 2022-11-08
HIP-SYCL Interop - unable to destroy CHIPContextLevel0


pvelesko:
For all 4 of these samples, if we try to destroy the context we get a segfault. I can see how that could happen when we are initializing CHIP-SPV from SYCL handles since we do not own the memory for the provided context but the other way around doesn't make sense to me since the default is for MKL to not retain the context.

Sarbojit2019:
Fix raised in PR #260

Sarbojit2019:
Fixed

PR #260, opened 2022-11-18
Fix crash seen in hip_sycl_interop samples

Sarbojit2019:


Sarbojit2019:
PR has fix for #258 and #253

samples/hip_sycl_interop/hip_sycl_interop.cpp
```diff
@@ -61,7 +61,13 @@ void VerifyResult(float *c_A, float *c_B) {
...
 
 int main() {
-  std::string envVar = std::getenv("CHIP_BE");
+  const char* val = std::getenv("CHIP_BE");
+  // In case CHIP_BE is not present
+  if (!val) {
+    std::cout << "HIP_SKIP_THIS_TEST" << std::endl;
+    exit(0);
+  }
+  std::string envVar(val);
```
pvelesko, line 70: the runtime will default to something for all other tests so we should query the runtime to get this value rather than reading the env.

pvelesko (changes requested):

pvelesko:
sycl_hip samples are still segfaulting but hip_sycl is now passing.

pvelesko (approved):

PR #263, opened 2022-11-24
Graphs

pvelesko:
HIP Graph support. 

Things in this PR:
√ HIP Graph API support with the exception of `hipGraphExecUpdate`
√ 100% pass rate for unit tests included in the version of HIP-COMMON
√ 2 Additional samples

Things not in this PR:
x Optimizations for Level Zero to take advantage of command lists.


pvelesko:
@jz10

pjaaskel:
Nice! Did you check if your implementation could utilize the OpenCL command buffer extension beneficially?

pvelesko:
> Nice! Did you check if your implementation could utilize the OpenCL command buffer extension beneficially?

@pjaaskel currently graphs are executed by issuing the regular HIP API commands neither ocl command buffers nor l0 command lists are being taken advantage of but that part is rather easy compared to getting the overall graph execution model to work. In short, it will.

I plan on merging this and then focusing on updating the HIP base version we are using since currently we only have ~12 HIP graphs tests so I'm sure there are lots of edge-case bugs.

PR #264, opened 2022-11-25
Fix compiler error with __attribute__((__noinline__))

linehill:
Fix a compiler error caused by C++ libraries (e.g. libstdc++ provided by g++-12) using `__attribute__((__noinline__))` in their code. Our definition of the `__noinline__` appears in the compilation as followed:

```
...
/usr/lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/bits/shared_ptr_base.h:196:37: error: use of undeclared identifier 'noinline'; did you mean 'inline'?
      __attribute__((__attribute__((noinline))))
...
```

For Clang <15 this issue is fixed by substituting the `__noinline__` with empty string and, therefore, rendering it ineffective, unfortunately. Fortunately, In Clang >=15 the `__noinline__` is a builtin keyword and the following use cases work in CUDA and HIP language modes:

```
__attribute__((__inline__)) void foo(...);
__inline__ void bar(...);
```
Fixes #138. 

Also, piggybacking a fix for hipcc missing headers when run in the build directory.

PR #266, opened 2022-11-30
Add missing ihipEvent_t type 

linehill:
Found out that a HIP client is using `ihipEvent_t` type, which is missing in the CHIP-SPV, so add one.

NOTE: This patch is stacked on #264.

### DIFF
--- a/src/CHIPBackend.hh
+++ b/src/CHIPBackend.hh
@@ -536,7 +536,7 @@ class CHIPQueue;
 class CHIPContext;
 class CHIPDevice;
 
-class CHIPEvent {
+class CHIPEvent : public ihipEvent_t {
 protected:
   bool UserEvent_ = false;
   bool TrackCalled_ = false;
--- a/src/CHIPBindings.cc
+++ b/src/CHIPBindings.cc
@@ -1364,7 +1364,7 @@ hipError_t hipStreamWaitEvent(hipStream_t Stream, hipEvent_t Event,
   ERROR_IF((Stream == nullptr), hipErrorInvalidResourceHandle);
   ERROR_IF((Event == nullptr), hipErrorInvalidResourceHandle);
 
-  std::vector<CHIPEvent *> EventsToWaitOn = {Event};
+  std::vector<CHIPEvent *> EventsToWaitOn = {(CHIPEvent *)Event};
   Stream->enqueueBarrier(&EventsToWaitOn);
 
   RETURN(hipSuccess);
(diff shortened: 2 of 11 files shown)


src/CHIPBackend.hh
```diff
@@ -536,7 +536,7 @@ class CHIPQueue;
 class CHIPContext;
 class CHIPDevice;
 
-class CHIPEvent {
```
pvelesko, line 539: Why didn't I come up with that.. 

Anyways, I think this is an issue because of our modified HIP headers. It didn't occur to me make CHIP classes inherit from these internal structures. iirc, `ihipEvent_t` is defined in the original HIP headers. 

Having said that, structures that start with `i` are internal and should not be used by external libraries

linehill: > Having said that, structures that start with i are internal and should not be used by external libraries

I thought that too initially, but after seeing the PyTorch code and [this](https://docs.amd.com/bundle/HIP_Supported_CUDA_API_Guide_v5.3/page/CUDA_Runtime_API_functions_supported_by_HIP.html) (which lists the `ihipEvent_t` as supported) it seems more that the clients may use the `ihipEvent_t` as a forward declaration.

pvelesko:
We should expand on this PR to undo changes in `HIP/hip_runtime_api.h` where I specifically exclude these definitions:

```
#ifndef __HIP_PLATFORM_SPIRV__
typedef struct ihipEvent_t* hipEvent_t;
typedef int hipDevice_t;
typedef struct ihipCtx_t* hipCtx_t;
typedef struct ihipStream_t* hipStream_t;
typedef struct ihipModule_t* hipModule_t;
typedef struct ihipModuleSymbol_t* hipFunction_t;
#endif
```

- [ ] Revert the above changes to `HIP/hip_runtime_api.h` (so that there is no diff with respect to HIP develop branch) 
- [ ] Add blank definitions inside `hip_to_chip.hh` like `struct ihipEvent_t {};`
- [ ] Use `static_cast` inside `CHIPBindings.cc` to cast HIP types to CHIP types 
- [ ] Check `static_cast` result for `nullptr` to make sure the cast was successful. 
- [ ] Update HIP submodule

pvelesko (changes requested):
see TODO from previous comment.

src/CHIPBindings.cc
```diff
@@ -1565,16 +1566,21 @@ hipError_t hipEventElapsedTime(float *Ms, hipEvent_t Start, hipEvent_t Stop) {
...
   }
-  if (Start->getFlags().isDisableTiming() || Stop->getFlags().isDisableTiming())
+  if (ChipEvStart->getFlags().isDisableTiming() ||
+      ChipEvStop->getFlags().isDisableTiming())
     CHIPERR_LOG_AND_THROW("One of the events has timings disabled. "
                           "Unable to return elasped time",
                           hipErrorInvalidResourceHandle);
 
-  *Ms = Start->getElapsedTime(Stop);
+  *Ms = ChipEvStart->getElapsedTime((CHIPEvent *)Stop);
```
pvelesko, line 1583: @linehill can we make this somehow automatic?

linehill:
Redundant except for a test case which is moved to separate PR (#300).

PR #268, opened 2022-11-30
Implement warpSize and cross-lane intrinsics

### LINKED ISSUE #226: Implement warpSize
Warp vs. subgroup is a messy affair (#219), so I suggest we define warpSize to 32 at least when building CUDA sources and pass the required subgroup annotation whenever a kernel refers to warpSize or any of the shuffles that rely on a fixed warp size (or thread-to-lane mapping). 

* CUDA defines the [warp size as a device-wide constant (32)](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#simt-architecture). 

* [HIP defines](https://github.com/ROCm-Developer-Tools/HIP/blob/develop/docs/markdown/hip_kernel_language.md#warpsize): "Note that all current Nvidia devices return 32 for this variable, and all current AMD devices return 64. Device code should use the warpSize built-in to develop portable wave-aware code." which sounds like it could be a per-kernel value, but since it's also [in the device properties](https://rocm-developer-tools.github.io/HIP/structhipDeviceProp__t.html#af3357d33c004608bf05bc21a352be81b), it has to be device specific.

This means that we might do with a simple `constexpr const unsigned int warpSize = 32` global variable implementation.
I believe most of the workloads originate from CUDA so it might be more robust to use and optimize for 32 here instead of 64? But let's make it a single global so it's easy to tweak/tune in the future. However, we have to fix it to something otherwise SPIR-Vs which rely on warpSize become non-portable (unless we introduce the warpSize as a SPIR-V built-in).

The warpSize parameter of the [HIP-versions of the shuffles](https://github.com/ROCm-Developer-Tools/HIP/blob/develop/docs/markdown/hip_kernel_language.md#warp-shuffle-functions) has to then check if the warpsize of the device matches the given one, if not, adjust the lane mapping function accordingly.

pjaaskel:
Fixes #226.


CMakeLists.txt
```diff
@@ -169,6 +176,7 @@ set(HIP_OFFLOAD_LINK_OPTIONS "" CACHE STRING "space separated list of compiler f
 
 # =============================================================================
 # CHIP-SPV CMAKE OPTIONS PARSING
+list(APPEND CHIP_SPV_DEFINITIONS DEFAULT_WARP_SIZE=${DEFAULT_WARP_SIZE})
```
linehill, line 179: Seems redundant as the warp size is put into the Config.hh too. I'd think the definition in the header is enough.

samples/shuffles/shuffles.cc
```diff
@@ -0,0 +1,104 @@
...
+  }                                                                            \
+  std::cout << std::endl
+
+#define LAUNCH_CASE(KERNEL, DELTA, WIDTH)                                      \
+  LAUNCH_CASE_T(int, KERNEL, DELTA, WIDTH);                                    \
+  LAUNCH_CASE_T(float, KERNEL, DELTA, WIDTH)
+
+int main(int argc, char *argv[]) {
+
+  uint Threads = DEFAULT_WARP_SIZE * 2;
```
linehill, line 81: Shouldn't the warp size be queried from the device?

llvm_passes/HipWarps.cpp
```diff
@@ -0,0 +1,97 @@
...
+      break;
+    }
+  }
+
+  if (!SensitiveFuncFound)
+    return PreservedAnalyses::all();
+
+  auto &Ctx = Mod.getContext();
+  for (auto &F : Mod) {
+    for (auto &BB : F) {
```
linehill, line 84: Unintended loop? I'm seeing that the loop repeatedly attaches the same metadata to the same function.

include/hip/devicelib/sync_and_util.hh
```diff
@@ -25,6 +25,17 @@
 
 #include <hip/devicelib/macros.hh>
 
+#ifndef DEFAULT_WARP_SIZE
+// Warpsize would optimally be a device-specific, queried and made effective
+// at runtime. However, we need to fix the warpsize since SPIR-Vs need to be
+// portable across multiple devices. It should be more portable to fix it to 32
+// instead of 64, with a trade-off of potential underutilization with GPUs
+// with 64 wide warps.
+#define DEFAULT_WARP_SIZE 32
```
linehill, line 34: Would `#error`be better here? There is a risk of having the warp size to be out of sync.

llvm_passes/HipWarps.cpp
```diff
@@ -0,0 +1,97 @@
...
+// synchronization for cases where warp lock-step semantics guarantees
+// a well-defined read-modify-write interleaving inside the warp. We should
+// add an annotation that guarantees subgroup lockstep semantics in that case.
+// There is not such an OpenCL extension yet to my knowledge.
+//===----------------------------------------------------------------------===//
+
+#include "HipWarps.h"
+
+#include <set>
+#include <iostream>
```
linehill, line 28: Unused headers?

llvm_passes/HipWarps.h
```diff
@@ -0,0 +1,30 @@
...
+// (c) 2022 Pekka Jääskeläinen / Intel
+//===----------------------------------------------------------------------===//
+
+#ifndef LLVM_PASSES_HIP_WARPS_H
+#define LLVM_PASSES_HIP_WARPS_H
+
+#include "llvm/IR/IRBuilder.h"
+#include "llvm/IR/PassManager.h"
+
+#include <map>
```
linehill, line 20: Unused header?

CMakeLists.txt
pjaaskel: It's not included everywhere, unfortunately.

samples/shuffles/shuffles.cc
pjaaskel: We cannot do that as the SPIR-Vs are warp-size specific, which are produced at compile time. To make it adapt to the device at runtime we either should introduce a builtin to query the warp size also at SPIR-V runtime or fix the warp size at CHIP-SPV build-time and make the devices adapt to whatever we have fixed it to. It's not very feasible approach as we rely on kernel annotation currently to make the warp-primitives work.

llvm_passes/HipWarps.cpp
pjaaskel: Yes, a copy-pasta eror.

include/hip/devicelib/sync_and_util.hh
pjaaskel: Yes, you are right, but the problem is hipcc, we should update it to also -DDEFAULT_WARP_SIZE=X and I wanted to avoid touching that in this PR for now.

CMakeLists.txt
pjaaskel: I made it installable (renamed to CHIPSPVConfig.hh to avoid file name clashes) and now included from the builtin header.

samples/shuffles/shuffles.cc
```diff
@@ -0,0 +1,105 @@
...
+  }                                                                            \
+  std::cout << std::endl
+
+#define LAUNCH_CASE(KERNEL, DELTA, WIDTH)                                      \
+  LAUNCH_CASE_T(int, KERNEL, DELTA, WIDTH);                                    \
+  LAUNCH_CASE_T(float, KERNEL, DELTA, WIDTH)
+
+int main(int argc, char *argv[]) {
+
+  uint Threads = CHIP_DEFAULT_WARP_SIZE * 2;
```
linehill, line 82: > We cannot do that as the SPIR-Vs are warp-size specific, which are produced at compile time. To make it adapt to the device at runtime we either should introduce a builtin to query the warp size also at SPIR-V runtime or fix the warp size at CHIP-SPV build-time and make the devices adapt to whatever we have fixed it to. It's not very feasible approach as we rely on kernel annotation currently to make the warp-primitives work.

I was meant to suggest that the warp size queried from `hipDeviceProp_t::warpSize` which is set to `CHIP_DEFAULT_WARP_SIZE` by the host runtime due to the reasons you mentioned.

linehill: This sample probably has an assumption for the warp size, so you probably want to check that the queried warp size is suitable and skip the test if it isn’t (with this CMake [feature](https://cmake.org/cmake/help/latest/prop_test/SKIP_RETURN_CODE.html#skip-return-code)).

CMakeLists.txt
```diff
@@ -577,6 +585,7 @@ install(TARGETS CHIP
   INCLUDES DESTINATION ${CHIP_CMAKE_INSTALL_INCLUDEDIR}
 )
 
+install(FILES ${CMAKE_BINARY_DIR}/include/CHIPSPVConfig.hh DESTINATION ${INCLUDE_INSTALL_DIR})
```
linehill, line 588: > It's not included everywhere, unfortunately.

It is, isn't it?
```
$ grep Config.hh -r src/
src/Filesystem.hh:#include "Config.hh"
src/common.hh:#include "Config.hh"
```

> I made it installable (renamed to CHIPSPVConfig.hh to avoid file name clashes) and now included from the builtin header.

I don't know why this file needs to be installable. What issue this solves?

pjaaskel: (This was clarified F2F)

samples/shuffles/shuffles.cc
pjaaskel: I added a check for the subgroup capabilities of the device and logWarn() if the CHIPSPV's set warpSize is not supported.

pjaaskel:
This is really to be pulled in (after main builds again).

linehill:
FYI, the patch does not build on my machine:
```
<...>/chip-spv/llvm_passes/HipGlobalVariables.cpp:40:
/mnt/md1/linehill/ws-pytorch-hip/chip-spv/llvm_passes/../src/common.hh:26:10: fatal error: 'CHIPSPVConfig.hh' file not found
#include "CHIPSPVConfig.hh"
         ^~~~~~~~~~~~~~~~~~
1 error generated.
```
This is probably caused by the config header (CHIPSPVConfig.hh) being placed in different location than before.

linehill (changes requested):
LGTM after the build error is fixed.

linehill:
FYI, shuffle samples fails too:

```
[ 34%] Building CXX object samples/4_shfl/CMakeFiles/broadcast2.dir/broadcast2.cpp.o
llvm-spirv: /mnt/md1/linehill/ws-chip-spv-2/llvm-project/llvm/projects/SPIRV-LLVM-Translator/lib/SPIRV/libSPIRV/SPIRVModule.cpp:172: virtual void SPIRV::SPIRVModuleImpl::setSPIRVVersion(SPIRV::SPIRVWord): Assertion `this->isAllowedToUseVersion(static_cast<VersionNumber>(Ver))' failed.
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace.
Stack dump:
0.	Program arguments: /mnt/md1/linehill/ws-chip-spv-2/install/bin/llvm-spirv --spirv-max-version=1.1 --spirv-ext=+all broadcast2-hip-spirv64-generic-lower.bc -o broadcast2-hip-spirv64-generic.out
...
```
To get this point, made the following change locally:

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 84f7cf69..512ddae5 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -367,7 +367,7 @@ target_include_directories(CHIP
   PRIVATE
   "${CMAKE_SOURCE_DIR}/src"
   "${CMAKE_SOURCE_DIR}/include"
-  "${CMAKE_BINARY_DIR}"
+  "${CMAKE_BINARY_DIR}/include"
 )
 
 # CHIP-SPV COMPILATION SETUP
diff --git a/llvm_passes/CMakeLists.txt b/llvm_passes/CMakeLists.txt
index ad5687d8..07daa0a7 100644
--- a/llvm_passes/CMakeLists.txt
+++ b/llvm_passes/CMakeLists.txt
@@ -48,7 +48,7 @@ endif()
 
 add_definitions(${LLVM_DEFINITIONS})
 
-include_directories(${LLVM_INCLUDE_DIRS} ${CMAKE_BINARY_DIR})
+include_directories(${LLVM_INCLUDE_DIRS} ${CMAKE_BINARY_DIR}/include)
 
 if(NOT LLVM_ENABLE_RTTI)
   add_compile_options("-fno-rtti")

```

pjaaskel:
Yep, the problem was hidden for me due to #274 and it finding my installed header.

pjaaskel:
Pushed the include path fix, and then I remembered that I had to up the max allowed SPIR-V number to enable the shuffles, lack of which causes your assertion. We need to upstream this:
```C++
diff --git a/clang/lib/Driver/ToolChains/HIPSPV.cpp b/clang/lib/Driver/ToolChains/HIPSPV.cpp
index d68c87e9b3e7..bf941ad272af 100644
--- a/clang/lib/Driver/ToolChains/HIPSPV.cpp
+++ b/clang/lib/Driver/ToolChains/HIPSPV.cpp
@@ -97,7 +97,7 @@ void HIPSPV::Linker::constructLinkAndEmitSpirvCommand(

   // Emit SPIR-V binary.

-  llvm::opt::ArgStringList TrArgs{"--spirv-max-version=1.1",
+  llvm::opt::ArgStringList TrArgs{"--spirv-max-version=1.2",
                                   "--spirv-ext=+all"};
   InputInfo TrInput = InputInfo(types::TY_LLVM_BC, TempFile, "");
   SPIRV::constructTranslateCommand(C, *this, JA, Output, TrInput, TrArgs);
```

pjaaskel:
https://github.com/intel/compute-runtime/issues/585

pjaaskel:
I'll create CHIP-SPV branches for LLVM and SPIRV-LLVM until they get upstreamed.

linehill:
> We need to upstream this: ...

There's a better way to set the desired SPIR-V version here: https://github.com/parmance/llvm-project/commit/a7b4ae2c5282d79145759db420eb56103dc23419.

pjaaskel:
@linehill: After getting my mind back to this, I recalled what the problem with your patch is. LLVM-SPIRV-Translator uses the version number from the LLVM module triple as a minimum. So while we'd pass the version (also) as a max via the cmdline arg, it grabs the minimum from the module header's triple which is propagated via the target string: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/commit/522d8f9134490b72fa11ccc687707e00408464f9

linehill:
A [link](https://github.com/llvm/llvm-project/issues/53596) of interest stating that the SPIR-V version in the triple should restrict the maximum version too.

pjaaskel:
I realized that since I now again use the intel_sub_group_shuffles, the patches to SPIR-V Translator or LLVM are not needed. Please test `ctest -R shuffle` in your platform. There should be no regressions in the test suite. Ballots might still fail due to the SPIR-V version issue, but they are not tested in the suite.

pjaaskel:
@linehill @pvelesko this should be ready, please test & push when you have a chance.

pvelesko:
lint & out of sync with main. 

> please test

we have CI now so it should test automatically once you rebase.

pvelesko:
I will review and merge over the weekend. @pjaaskel

pjaaskel:
Rebased, added __lane_id(). Great to hear there's now a CI somewhere!

pvelesko:
Sorry forgot about this - will try to review tomorrow.

PR #276, opened 2022-12-06
hipGetNativeHandle should call CHIPInitialize()

Sarbojit2019:
hipGetNativeHandle assumes that CHIP is already initialized but there can be cases where hipGetNativeHandle is the first call hence CHIPInitialize() need to be called.


linehill (approved):
LGTM.

Issue #48, opened 2022-04-27
OpenCL Regressions When Using Clang14


pvelesko:
A lot tests result in failures due to `CL_INVALID_KERNEL_ARGS` when using clang14 + OpenCL. This is not the case for Clang13 

```bash
CHIPKernelOpenCL::CHIPKernelOpenCL(const cl::Kernel &&, std::string, OCLFuncInfo *): Assertion `FuncInfo_->ArgTypeInfo.size() == NumArgs' failed.
```

pvelesko:
For some reason this does not affect Iris

linehill:
OpenCL test cases fail on some Intel Compute Runtime versions. The issue you described is probably related to this intel/compute-runtime#535.

For me, the OpenCL test cases pass on version 22.04.22286.

zjin-lcf:
Running a sample (e.g. matrix_multiply) shows the error with BE=opencl on Intel DevCloud.  Could you please update the document (install) so that the BE=level0 by default ?

PR #311, opened 2023-01-18
Preserve device symbols for better diagnostics

linehill:
... when possible.

When device code compilation fails at the runtime the build log is lacking the symbol names. For example:

```
CHIP error [TID 1984562] [1674030579.550524982] : ZE Build Log:
error: undefined reference to `'
in function: '' called by kernel: 'repro_kernel(double*, double)'

error: backend compiler failed build.
```

This is caused by dropping OpName instructions overzealously in the SPIR-V modules which is done for working around an issue in CHIP-SPV/chip-spv#48.

Change the workaround code to drop only the OpNames that are needed for the workaround so we see symbol names in build logs.

### DIFF
--- a/src/spirv.cc
+++ b/src/spirv.cc
@@ -235,15 +235,9 @@ public:
       Extra_ = Pp;
     }
 
-    if (Opcode_ == spv::Op::OpExtInstImport) {
-      const char *Pp = (const char *)(Stream + 2);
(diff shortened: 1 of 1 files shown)


pvelesko:
conflicts @linehill

linehill:
Rebased.

PR #312, opened 2023-01-18
CI

pvelesko:
Enable CI & Small Fixes
    * Level Zero Device Selection
    * Updated README
    * CMake - only LLVM_CONFIG is required
    * CMake - FindLLVM.cmake sets compiler and other binaries
    based on llvm-config
    * Cmake - LLVMCheck.cmake checks for LLVM version
    * Disabled tests for make check pass


pvelesko:
I'm gonna go ahead and merge this so we can get the unit tests running for all other outstanding PRs. If anyone has any comments please open an issue.

Issue #279, opened 2022-12-09
Size of struct counted wrongly


pjaaskel:
The SPIR-V parser seems to assume packed structs whereas it's unclear what SPIR-V expects for the struct alignment. The OpenCL SPIR-V spec [states only for the basic datatypes that they should be naturally aligned](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Env.html#_alignment_of_types). Typically structs are aligned according to their largest datatype (at least it was so in HSA IIRC -- need to check) and similar struct align constraint (and thus size computation) might be implied for SPIR-V although not clearly specified (to check more closely).

The bug shows itself via a test case where a struct with an int and long is passed as value as kernel arg. The arg size is computed as 12B whereas the drivers compute it 16B. CPU driver checks for the size and asserts. GPU driver likely doesn't check the size and proceeds. 

A reproducer is here: https://github.com/pjaaskel/chip-spv/commit/aaacf6f76ed4ca32803072402b2b46e8546507b3

pjaaskel:
There's a [CPacked decoration](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#_decoration) in SPIR-V for structs, but I cannot find the default padding rules.

franz:
> GPU driver likely doesn't check the size and proceeds

Have you checked (tried with other sizes) ? I vaguely remember that some driver (original AMD HSA ?) compiled by default all structs packed (possibly because it allows to cram more stuff into GPU registers), so i would not be surprised if GPU driver accepts 12 as correct value. I would be much more surprised if the GPU driver didn't check argument sizes.

pjaaskel:
Nope, didn't check more thoroughly since in any case we need to fix the size calculation, it likely now works by luck somehow when passing non-packed structs.

pjaaskel:
The test case produces the following with the CPU driver:
```
CHIP_DEVICE_TYPE=cpu ./hip-repro 
CHIP error [TID 197636] [1671211673.927720452] : hipErrorTbd (CL_INVALID_ARG_SIZE clSetKernelArg failed) in /home/pjaaskel/src/chip-spv/src/backend/OpenCL/CHIPBackendOpenCL.cc:1163:setupAllArgs

CHIP error [TID 197636] [1671211673.927842976] : Caught Error: hipErrorTbd
Result: 0
Device side sizeof: 94074260261712
Components: 94074259666016 94074260371200
```
[...]

PR #313, opened 2023-01-19
Fix kernel argument passing for aggregates

linehill:
Fix kernel argument passing for aggregates

Fix sizes of aggregate arguments which were not computed correctly causing kernel arguments to be copied partially or copy more bytes than needed to device. The issues found and fixed in the argument size computation logic:

* Array lengths were not extracted from the SPIR-V properly causing random argument sizes.
* The argument size computation logic for structures/arrays did not account alignment requirements of its members/elements.


pvelesko (approved):
Looks good but can you provide some background on how these issues were discovered? Were no unit tests affected by this? 

Will merge after after I merge the CI upgrades.

linehill:
> Looks good but can you provide some background on how these issues were discovered? 

It's already recognized issue (#279).

> Were no unit tests affected by this?

By a glance they don't seem to test kernel aggregate arguments.

PR #318, opened 2023-01-22
Coarse grain SVM support + Page Locked Memory + Resolved Unit Test Script

pvelesko:
* Implement page locking via `mlock()` 
* Check for fine-grain SVM support and if available, OpenCL will allocate memory using fine-grain and not perform any map/unmap operations
* If using coarse-grain SVM, perform map/unmap operations
* Add resolved unit test checker script


pvelesko:
From my understanding, `hipHostMalloc` should be doing a page lock on the allocated memory. I do it this way: 
```
  int PageLockSuccess = mlock(RetVal, Size);
```

It works when I run locally but it fails during CI. Any ideas? @pjaaskel

pvelesko:
Current mlock failure as reported by errno
```
ENOTTY 25 Inappropriate ioctl for device
```

pvelesko:
https://support.hashicorp.com/hc/en-us/articles/115012787688-Vault-and-mlock-

pvelesko:
This PR seems to crash the driver every time I run CI. In fact, the driver crashes during compilation. 

* Compilation affects the GPU because the test discovery happens post-build which invokes the driver. 
* By default, Level Zero is used. 
* Level Zero has proved to be quite unstable especially when multiple threads are used.
* Multiple threads are always used when building so `zeInit` is called from multiple threads.

PR #333, opened 2023-01-26
CI Upgrade 2

pvelesko:
* Load/unload device from the driver before each test


cmake/UnitTests.cmake
```diff
@@ -9,6 +9,7 @@ list(APPEND DGPU_OPENCL_FAILED_TESTS " ")
 list(APPEND IGPU_OPENCL_FAILED_TESTS " ") 
 list(APPEND IGPU_LEVEL0_FAILED_TESTS " ")
 list(APPEND DGPU_LEVEL0_FAILED_TESTS " ") 
+list(APPEND POCL_LEVEL0_FAILED_TESTS " ")
```
pjaaskel, line 12: I believe your are testing with the CPU driver still? PoCL-L0 likely fails a lot more tests currently. Best name this POCL_CPU_FAILED_TESTS as we will add also POCL_LEVEL0_FAILED_TESTS when most of the tests pass?

scripts/check_unit_tests.sh
```diff
@@ -0,0 +1,137 @@
...
+echo "end igpu_opencl_failed_tests"
+
+# Test OpenCL dGPU
+echo "begin dgpu_opencl_failed_tests"
+sudo /opt/ocl-icd/scripts/dgpu_bind &> /dev/null
+CHIP_BE=opencl ctest --timeout 180 -j 8 --output-on-failure -E "`cat ./test_lists/dgpu_opencl_failed_tests.txt`" | tee dgpu_opencl_make_check_result.txt
+sudo /opt/ocl-icd/scripts/dgpu_unbind &> /dev/null
+echo "end dgpu_opencl_failed_tests"
+
+# Test OpenCL CPU
```
pjaaskel, line 64: Intel OpenCL CPU

pjaaskel (changes requested):

pjaaskel (approved):
LGTM

pvelesko:
For some reason sycl test is failing when run through CI

pvelesko:
@pjaaskel I added a python script for running tests - let me know if it works for you. 

Still need to investigate why the sycl test segfaults upon exit but only when run through CI.

PR #347, opened 2023-02-03
minor fixes

franz:
mainly to improve testing

### DIFF
--- a/cmake/UnitTests.cmake
+++ b/cmake/UnitTests.cmake
@@ -391,7 +391,6 @@ list(APPEND CPU_OPENCL_FAILED_TESTS "Unit_hipMallocPitch_KernelLaunch - float") 
 list(APPEND CPU_OPENCL_FAILED_TESTS "Unit_hipMallocPitch_KernelLaunch - double") # Failed
 list(APPEND CPU_OPENCL_FAILED_TESTS "Unit_hipGraphAddHostNode_ClonedGraphwithHostNode") # Failed
 
-
 # iGPU OpenCL Unit Test Failures
 list(APPEND IGPU_OPENCL_FAILED_TESTS "cuda-simpleCallback") # SEGFAULT
 list(APPEND IGPU_OPENCL_FAILED_TESTS "cuda-qrng") # Subprocess aborted
@@ -2358,6 +2357,36 @@ list(APPEND CPU_POCL_FAILED_TESTS "Unit_deviceFunctions_CompileTest_y1f_float") 
 list(APPEND CPU_POCL_FAILED_TESTS "Unit_deviceFunctions_CompileTest_ynf_float") # Failed
 list(APPEND CPU_POCL_FAILED_TESTS "Unit_hipClassKernel_Friend") # SEGFAULT
 list(APPEND CPU_POCL_FAILED_TESTS "Unit_hipDeviceSynchronize_Functional") # Failed
+# broken tests, they all try to write outside allocated memory;
+# valgrind + pocl shows:
+#
+#==11492== Invalid write of size 1
+#==11492==    at 0x5605B83: pocl_fill_aligned_buf_with_pattern (pocl_util.c:2590)
+#==11492==    by 0x562B1F5: pocl_driver_svm_fill (common_driver.c:444)
+#==11492==    by 0x5626EBF: pocl_exec_command (common.c:693)
+#==11492==    by 0x562205F: pthread_scheduler_get_work (pthread_scheduler.c:529)
+#==11492==    by 0x56221B7: pocl_pthread_driver_thread (pthread_scheduler.c:588)
+#==11492==    by 0x5009B42: start_thread (pthread_create.c:442)
+#==11492==    by 0x509ABB3: clone (clone.S:100)
+#==11492==  Address 0x114baf00 is 0 bytes after a block of size 16,384 alloc'd
+#==11492==    by 0x561EE67: pocl_basic_svm_alloc (basic.c:841)
+#==11492==    by 0x561411A: POclSVMAlloc (clSVMAlloc.c:98)
+#==11492==    by 0x4966168: SVMemoryRegion::allocate(unsigned long) (source/chip-spv/src/backend>
+#
+# running with older PoCL: "double free or corruption"
+#
+# running with Intel CPU runtime or new PoCL: hipErrorRuntimeMemory (CL_INVALID_VALUE )
+# in CHIPBackendOpenCL.cc:1048:memFillAsyncImpl
+list(APPEND CPU_POCL_FAILED_TESTS "Unit_hipMemset2D_BasicFunctional") # Timeout
+list(APPEND CPU_POCL_FAILED_TESTS "Unit_hipMemset2DAsync_BasicFunctional") # Timeout
+list(APPEND CPU_POCL_FAILED_TESTS "Unit_hipMemset2DAsync_WithKernel") # Timeout
+list(APPEND CPU_POCL_FAILED_TESTS "Unit_hipMemset2DAsync_MultiThread") # Timeout
+list(APPEND CPU_POCL_FAILED_TESTS "Unit_hipMemsetFunctional_ZeroValue_2D") # Timeout
+list(APPEND CPU_POCL_FAILED_TESTS "Unit_hipMemsetASyncMulti") # Failed
+list(APPEND CPU_POCL_FAILED_TESTS "Unit_hipMemsetDASyncMulti - int8_t") # Failed
+list(APPEND CPU_POCL_FAILED_TESTS "Unit_hipMemsetDASyncMulti - int16_t") # Failed
+list(APPEND CPU_POCL_FAILED_TESTS "Unit_hipMemsetDASyncMulti - uint32_t") # Failed
+
 
 
 list(APPEND ALL_FAILED_TESTS ${DGPU_OPENCL_FAILED_TESTS})
--- a/samples/ccompat/CMakeLists.txt
+++ b/samples/ccompat/CMakeLists.txt
@@ -1,7 +1,7 @@
 add_custom_command(OUTPUT saxpy.o   DEPENDS ${CMAKE_SOURCE_DIR}/samples/ccompat/saxpy.hip COMMAND ${CMAKE_BINARY_DIR}/bin/hipcc.bin ${CMAKE_SOURCE_DIR}/samples/ccompat/saxpy.hip -c -g)
 add_custom_command(OUTPUT ccompat.o DEPENDS ${CMAKE_SOURCE_DIR}/samples/ccompat/ccompat.c COMMAND ${CMAKE_BINARY_DIR}/bin/hipcc.bin ${CMAKE_SOURCE_DIR}/samples/ccompat/ccompat.c -c -g)
 
-add_custom_target(ccompat DEPENDS saxpy.o ccompat.o COMMAND ${CMAKE_BINARY_DIR}/bin/hipcc.bin ./ccompat.o ./saxpy.o -o ccompat -g)
+add_custom_target(ccompat ALL DEPENDS saxpy.o ccompat.o COMMAND ${CMAKE_BINARY_DIR}/bin/hipcc.bin ./ccompat.o ./saxpy.o -o ccompat -g)
 
 add_dependencies(ccompat hipcc.bin CHIP)
 add_test(NAME ccompat COMMAND ${CMAKE_BINARY_DIR}/samples/ccompat/ccompat)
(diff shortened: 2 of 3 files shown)


cmake/UnitTests.cmake
```diff
@@ -391,6 +391,26 @@ list(APPEND CPU_OPENCL_FAILED_TESTS "Unit_hipMallocPitch_KernelLaunch - float")
 list(APPEND CPU_OPENCL_FAILED_TESTS "Unit_hipMallocPitch_KernelLaunch - double") # Failed
 list(APPEND CPU_OPENCL_FAILED_TESTS "Unit_hipGraphAddHostNode_ClonedGraphwithHostNode") # Failed
 
+# broken tests, they all try to write outside allocated memory;
+# valgrind + pocl shows:
```
pvelesko, line 395: Should be added to POCL exclusion tests. 
Also, these should be reproducible on the CI system.

cmake/UnitTests.cmake
```diff
@@ -2377,6 +2397,11 @@ string(REGEX REPLACE ";" "\$|" IGPU_LEVEL0_FAILED_TESTS_STR "${IGPU_LEVEL0_FAILE
 string(REGEX REPLACE ";" "\$|"    CPU_POCL_FAILED_TESTS_STR "${CPU_POCL_FAILED_TESTS}")
 string(REGEX REPLACE ";" "\$|" ALL_FAILED_TESTS_STR "${ALL_FAILED_TESTS}")
 
+if(NOT DEFINED PARALLEL_TESTS)
+  set(PARALLEL_TESTS 1)
+endif()
+
+set(TEST_OPTIONS -j ${PARALLEL_TESTS} --timeout 120 --output-on-failure)
 add_custom_target(check COMMAND ${CMAKE_CTEST_COMMAND} ${TEST_OPTIONS} -E ${ALL_FAILED_TESTS_STR} VERBATIM)
```
pvelesko, line 2405: make check is pretty much deprecated in favor for `scripts/check.py` because setting this option would force a reconfigure.

pvelesko (changes requested):
Seems that test failures are misplaced to CPU OpenCL instead of CPU PoCL. Also, should be be reproducible on the CI system.

cmake/UnitTests.cmake
pjaaskel: You mean they cannot be reproduced in the CI system? I think if it fails for someone we should still mark it (flaky?), otherwise we keep having uncertainty which tests are expected to work or not.

linehill: It does not seem like the CI pulls in any memory error detection analyzer/instrumentation so it's just luck that the out-of-bound accesses do not cause any troubles.

pvelesko: > I think if it fails for someone we should still mark it (flaky?), otherwise we keep having uncertainty which tests are expected to work or not.

I think we  should investigate why it fails for you and not for CI, especially for PoCL where we know that it's not related to the GPU driver.

pvelesko: > It does not seem like the CI pulls in any memory error detection analyzer/instrumentation so it's just luck that the out-of-bound accesses do not cause any troubles.

Any idea on what to do so that the CI system behavior matches yours?

cmake/UnitTests.cmake
franz: ok, removed the commit

franz:
Removed the `make check` commit, and moved the failing tests to `CPU_POCL_FAILED_TESTS`. 

Removed the commit with barriers, it seems the solution will need more work.

cmake/UnitTests.cmake
pjaaskel: I understood from the above that if you run it under valgrind, it should reveal writes that write over the buffers, but of course these type of errors might get unnoticed if they do not happen to overwrite anything essential. This is a major reason why we port to PoCL-CPU, to make it easier to debug issues like this.

franz:
`scripts/check.py` is broken. It prints:

```
 File "/home/michal/0/build/examples/chip-spv/src/chip-spv/scripts/check.py", line 55, in <module>
     run_cmd(cmd)
 NameError: name 'run_cmd' is not defined
```

pvelesko:
@franz pushed fix to main

cmake/UnitTests.cmake
```diff
@@ -2358,6 +2357,31 @@ list(APPEND CPU_POCL_FAILED_TESTS "Unit_deviceFunctions_CompileTest_y1f_float")
...
+#==11492==    by 0x4966168: SVMemoryRegion::allocate(unsigned long) (source/chip-spv/src/backend>
+list(APPEND CPU_POCL_FAILED_TESTS "Unit_hipMemset2D_BasicFunctional")
+list(APPEND CPU_POCL_FAILED_TESTS "Unit_hipMemset2DAsync_BasicFunctional")
+list(APPEND CPU_POCL_FAILED_TESTS "Unit_hipMemset2DAsync_WithKernel")
+list(APPEND CPU_POCL_FAILED_TESTS "Unit_hipMemset2DAsync_MultiThread")
+list(APPEND CPU_POCL_FAILED_TESTS "Unit_hipMemsetFunctional_ZeroValue_2D")
+list(APPEND CPU_POCL_FAILED_TESTS "Unit_hipMemsetASyncMulti")
+list(APPEND CPU_POCL_FAILED_TESTS "Unit_hipMemsetDASyncMulti - int8_t")
+list(APPEND CPU_POCL_FAILED_TESTS "Unit_hipMemsetDASyncMulti - int16_t")
+list(APPEND CPU_POCL_FAILED_TESTS "Unit_hipMemsetDASyncMulti - uint32_t")
```
pvelesko, line 2383: Can you add how these tests fail so that we're in line with the rest of the tests?

`# Failed/SEGFAULT/Subprocess Aborted`

franz: Added the failures.

I've added some [extra SVM pointer checks](https://github.com/pocl/pocl/pull/1171) in PoCL. With those, the first five tests (Memset2D) now "pass" - they get a `CL_INVALID_OPERATION` from PoCL and.... AFAICT they simply ignore it. So it's a "pass" but really a fail. BTW same thing happens with Intel OpenCL CPU runtime.

pvelesko: > So it's a "pass" but really a fail.

are you saying that the tests don't check the result of the HIP API call? Still don't understand why these tests are not failing on the CI system.

franz: > are you saying that the tests don't check the result of the HIP API call

That's my understanding. Here's a log from `ctest -V -R Unit_hipMemsetFunctional_ZeroValue_2D`:

```
test 860
    Start 860: Unit_hipMemsetFunctional_ZeroValue_2D

860: Test command: /home/michal/0/build/b_chip_15/catch/catch_tests/unit/memory/hipMemsetFunctional "Unit_hipMemsetFunctional_ZeroValue_2D"
860: Test timeout computed to be: 10000000
860: CHIP error [TID 19237] [1675784262.716106613] : hipErrorRuntimeMemory (CL_INVALID_VALUE ) in /home/michal/0/source/chip-spv/src/backend/OpenCL/CHIPBackendOpenCL.cc:1048:memFillAsyncImpl
860: 
860: CHIP error [TID 19237] [1675784262.716161386] : Caught Error: hipErrorRuntimeMemory
860: CHIP error [TID 19237] [1675784262.732102895] : hipErrorRuntimeMemory (CL_INVALID_VALUE ) in /home/michal/0/source/chip-spv/src/backend/OpenCL/CHIPBackendOpenCL.cc:1048:memFillAsyncImpl
860: 
860: CHIP error [TID 19237] [1675784262.732118556] : Caught Error: hipErrorRuntimeMemory
860: Filters: Unit_hipMemsetFunctional_ZeroValue_2D
860: ===============================================================================
860: All tests passed (32768 assertions in 1 test case)
860: 
1/1 Test #860: Unit_hipMemsetFunctional_ZeroValue_2D ...   Passed    0.62 sec
```

pvelesko: What is the issue with the log that you attached? @franz

cmake/UnitTests.cmake
```diff
@@ -2358,6 +2357,36 @@ list(APPEND CPU_POCL_FAILED_TESTS "Unit_deviceFunctions_CompileTest_y1f_float")
...
+# in CHIPBackendOpenCL.cc:1048:memFillAsyncImpl
+list(APPEND CPU_POCL_FAILED_TESTS "Unit_hipMemset2D_BasicFunctional") # Timeout
+list(APPEND CPU_POCL_FAILED_TESTS "Unit_hipMemset2DAsync_BasicFunctional") # Timeout
+list(APPEND CPU_POCL_FAILED_TESTS "Unit_hipMemset2DAsync_WithKernel") # Timeout
+list(APPEND CPU_POCL_FAILED_TESTS "Unit_hipMemset2DAsync_MultiThread") # Timeout
+list(APPEND CPU_POCL_FAILED_TESTS "Unit_hipMemsetFunctional_ZeroValue_2D") # Timeout
+list(APPEND CPU_POCL_FAILED_TESTS "Unit_hipMemsetASyncMulti") # Failed
+list(APPEND CPU_POCL_FAILED_TESTS "Unit_hipMemsetDASyncMulti - int8_t") # Failed
+list(APPEND CPU_POCL_FAILED_TESTS "Unit_hipMemsetDASyncMulti - int16_t") # Failed
+list(APPEND CPU_POCL_FAILED_TESTS "Unit_hipMemsetDASyncMulti - uint32_t") # Failed
```
pvelesko, line 2388: Do I need to update POCL on the CI machine to get these to fail?

franz: Not necessary. The PR with extra checks hasn't been merged to PoCL's master, so the first 5 tests should fail with current PoCL master branch. After it's merged, it'll be the last 4 tests that should fail. However, since any one of them can fail or pass, depending on where their out-of-bounds access falls, i'd like them all disabled for PoCL. Unreliable tests are a waste of time.

pvelesko: > so the first 5 tests should fail with current PoCL master branch

ok, so my question remains: what do I need to do to get them to fail on my machine?

> Unreliable tests are a waste of time.

right, that's why I keep pushing on this point. From what I understand about POCL these tests should fail everywhere not just on your machine. We should investigate why these fail for you and not on the CI machine and then make the necessary changes to the CI machine so that these fail universally. 

I want the CI machine to fail at least as much as any other machine the end user might have.

linehill: > what do I need to do to get them to fail on my machine?
> ...
> We should investigate why these fail for you and not on the CI machine...

Out-of-bounds accesses are undefined behavior in C/C++. Anything can happen and the outcome may change just every time the program is launched.

You could try to catch OoB accesses by compiling and running the CI with [AddressSanitizer](https://clang.llvm.org/docs/AddressSanitizer.html).

pvelesko: > Anything can happen and the outcome may change just every time the program is launched.

In this case I should be able to run it like 1000 times and get at least one of these tests to fail? 

I'm ok with excluding these tests - just trying to understand why these failures haven't popped up on the CI machine and if I need to do something to it to bring it more in-line with what end users might experience.

linehill: > In this case I should be able to run it like 1000 times and get at least one of these tests to fail?

There is no guarantee a failure would occur. Anything can happen - including a hypothetical case where the test failure is impossible in some builds/systems but fails sometimes/always in other builds/systems. If an undefined behavior is triggered in C/C++ application, all bets are off.

franz: FWIW, i have tried compiling PoCL with different Debug/Release configurations, and different -Ox options. Some failed with a timeout, and some failed with a segfault. The most reliable way to see the bug is using valgrind, however sanitizers Henry mentioned should work too.

cmake/UnitTests.cmake
franz: the issue with the log is that the test gets CL_INVALID_VALUE from one of the OpenCL calls, but the overall result (reported by ctest) is still "Passed". So either the test itself doesn't check the HIP call result, or there is a bug in CHIP-SPV and the result is ignored somewhere.

pvelesko: Maybe the tests are checking fot invalid parameters so an error must be thrown in order to pass the test.

franz: Possible but unlikely. With the OpenCL intel GPU as backend, the test doesn't print any HIP error and still passes.

franz:
@pjaaskel  yes it's ready.

cmake/UnitTests.cmake
pvelesko: I checked at least this one case and it's passing nullptr for device pointer so this is likely an argument check issue.

PR #353, opened 2023-02-07
fix unresolved fp16 symbols

### LINKED ISSUE #352: fp16 related build issues on JLSE
```
[100%] Built target RecursiveGaussian
/usr/lib64/gcc/x86_64-suse-linux/7/../../../../x86_64-suse-linux/bin/ld: CMakeFiles/fp16_math.dir/half_math.cpp.o: in function `__half::operator float<float, (void*)0>() const':
/home/pvelesko/t/chip-spv/include/hip/spirv_hip_fp16.h:296: undefined reference to `__extendhfsf2'
/usr/lib64/gcc/x86_64-suse-linux/7/../../../../x86_64-suse-linux/bin/ld: CMakeFiles/fp16_math.dir/half_math.cpp.o: in function `__half<float, nullptr>':
/home/pvelesko/t/chip-spv/include/hip/spirv_hip_fp16.h:167: undefined reference to `__truncsfhf2'
/usr/lib64/gcc/x86_64-suse-linux/7/../../../../x86_64-suse-linux/bin/ld: CMakeFiles/fp16_math.dir/half_math.cpp.o: in function `__half<short, nullptr>':
/home/pvelesko/t/chip-spv/include/hip/spirv_hip_fp16.h:181: undefined reference to `__truncsfhf2'
/usr/lib64/gcc/x86_64-suse-linux/7/../../../../x86_64-suse-linux/bin/ld: CMakeFiles/fp16_math.dir/half_math.cpp.o: in function `__half::operator unsigned short<unsigned short, (void*)0>() const':
/home/pvelesko/t/chip-spv/include/hip/spirv_hip_fp16.h:310: undefined reference to `__extendhfsf2'
```

pvelesko:
Fixes #352

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,7 +25,7 @@
 
 # temporary
 set(CMAKE_WARN_DEPRECATED OFF)
-add_compile_options(-Wno-format-extra-args)
+add_compile_options(-Wno-format-extra-args -mf16c)
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-duplicate-decl-specifier \
 -Wno-tautological-constant-compare  -Wno-c++20-extensions -Wno-unused-result \


CMakeLists.txt
```diff
@@ -25,7 +25,7 @@
 
 # temporary
 set(CMAKE_WARN_DEPRECATED OFF)
-add_compile_options(-Wno-format-extra-args)
+add_compile_options(-Wno-format-extra-args -mf16c)
```
pjaaskel, line 28: Is this safe to add unconditionally? Isn't this enabling the F16C ISA extension? What if the CPU doesn't support it?

pvelesko: > Isn't this enabling the F16C ISA extension?

if I understand correctly, yes. 

> What if the CPU doesn't support it?

based on this list https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html we'd be hard pressed to find a modern CPU that doesn't support it.

pjaaskel: > based on this list https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html we'd be hard pressed to find a modern CPU that doesn't support it.

I don't think we want to reduce the portability on "less-modern" CPUs or, say Arm hosts, just for this? Could we check for it?

pvelesko: I'll look into if we can check but let's say this option is not supported, then what?

pvelesko: @pjaaskel I'll merge this and open an issue to implement checks/whatever we decide. Not being able to compile main on JLSE is not good right now.

PR #358, opened 2023-02-09
hipComplex

pvelesko:


elwasif:
The copied source includes the file `hip/amd_detail/amd_hip_vector_types.h`  which does not exist in the SPIRV hip implementation

PR #351, opened 2023-02-06
Refactor kernel argument handling 

linehill:
* Replace complexity of the kernel argument iteration logic and its repetition in the code base with new visitor methods (SPVFuncInfo::visit*Arg).
* Remove argument handling for old launch API (`-fno-hip-new-launch-api`). It likely has bitrotted as we haven't been testing it.
* Remove redundancy in CHIPExecItem.
* As a by-product fixed an argument list iteration issue in RegisteredVarCopy(). It had the following change:
    ```
    -    if (ArgTyInfos[OutArgI].Type == OCLType::Sampler) {
    +    if (ArgTyInfos[OutArgI].Type != OCLType::Pointer) {
    ```
  Which caused argument index to become misaligned on kernels with hipTextureObject_t arguments.

* Moved OCLFuncInfo to its own file and changed it to use SPIR-V terminology.

This patch is a preparation for follow up patch(es) for fixing #350.

PR #359, opened 2023-02-13
Support 4KB kernel parameter lists

### LINKED ISSUE #350: error: Total size of kernel arguments exceeds limit
Some HIP and CUDA originated, hipified applications may fail to launch kernels due to their parameter list exceeding backend's limit (in my case 2 KB) while the sizes of the parameter list are below 4KB limit as stated [here](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html?highlight=nvrtc#function-parameters).

linehill:
Support 4KB kernel parameter lists on devices that might support lists less than that by passing some of the (large) argument in a device buffer.

Fixes #350.

TODOs before landing:
- [x] Merge #351.


pjaaskel:
@pvelesko: @linehill says there's flaky test that is sometimes failing here, can we put it to the flaky list to avoid silly reruns to push our chances? ;)

pvelesko:
Which test is that? Do we know why it became flaky?

linehill:
Unit_hipMemset3DAsync_ConcurrencyMthread is the one failing and it appeared after rebase - no changes in this patch were made.

pjaaskel (approved):

pvelesko (approved):

PR #360, opened 2023-02-13
Add a workaround for unsupported __float128 issue ...

linehill:
... which may appear with `-std=g++##` and some libstdc++ versions. Without it the device side compilation may fail with the following error message:

```
  /usr/lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/type_traits:436:39: error: __float128 is not supported on this target
    struct __is_floating_point_helper<__float128>
```

AFAIK, the clang incorrectly defines couple macros (`__FLOAT128__`, `__SIZEOF_FLOAT128`) for the device side compilation. It seems these macros leak from the host target. Clang bundles target macro defines for both the device and host target together and passes them to both the host and device compilation pipelines. The SPIR-V target does not define the macros as it does not support `__float128` but the host target may (in my case the x86_64 target defines them).

Also, as a byproduct, fixed `HOST_ARCH` variable from eating strings appearing after it by stripping whitespace in execute_process().

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -376,21 +376,30 @@ else()
   set(HIP_LINK_DEVICELIB_INSTALL --hip-path=${CMAKE_INSTALL_PREFIX})
 endif()
 
+# Include a header for applying fixups before any user or system includes.
+set(HIP_FIXUPS_HEADER_BUILD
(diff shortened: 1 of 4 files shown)


pjaaskel:
Is it ready?

linehill:
Yep, it's ready.

PR #361, opened 2023-02-15
Improve a shadow kernel for better native code-gen

linehill:
Improve a shadow kernel used for `__device__` variable initialization for achieving better native code in some drivers. This is known to fix the initialization code, for large `__device__` variables, compiling or running very slowly on an Intel iGPU.


pjaaskel (approved):
LGTM!

pvelesko (approved):
The test seems lacking pass/fail condition. Otherwise, LGTM

pjaaskel:
I assumed the "fail condition" is the timeout of ctest: If it (still) explodes the compile time it should fail due to the watchdog.

linehill:
Added a valgrind detected memory error.

PR #369, opened 2023-02-27
Updated samples with opencl interop

Sarbojit2019:
- [x] rebuild clang16 w/ shared libs
- [x] rebuild pocl against clang16

### DIFF
--- a/cmake/UnitTests.cmake
+++ b/cmake/UnitTests.cmake
@@ -2450,6 +2450,8 @@ list(APPEND CPU_POCL_FAILED_TESTS "Unit_deviceFunctions_CompileTest_ynf_float") 
 list(APPEND CPU_POCL_FAILED_TESTS "Unit_hipClassKernel_Friend") # SEGFAULT
 list(APPEND CPU_POCL_FAILED_TESTS "Unit_hipDeviceSynchronize_Functional") # Failed
 list(APPEND CPU_POCL_FAILED_TESTS "syncthreadsExitedThreads") # Timeout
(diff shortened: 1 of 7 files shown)


pvelesko:
@Sarbojit2019  unit tests

pvelesko (changes requested):
unit tests failed

pjaaskel:
Please just blacklist the sycl-interop test case for the PoCL-CPU until we get it fixed. It might be because of oneAPI SYCL blocking OpenCL implementations it doesn't know. It might work with the open source DPC++. We'll check.

pjaaskel:
@Sarbojit2019 PoCL is just an OpenCL implementation. We currently test the CPU driver of it, we also have Level0 driver, but it's not yet in the test suite to my knowledge. In this test environment it should work with the same CHIP-SPV parameters as the Intel's CPU OpenCL driver. But I suspect the oneAPI SYCL is more picky of the OpenCL platforms it wants to run on top (in contrary to the DPC++ open source branch).

pvelesko:
@pjaaskel I think Sarbojit was asking how to query the OpenCL runtime to check if it's PoCL or not.

pjaaskel:
Right. I don't know if HIP as an API for printing it out, but through the OpenCL binding at least. But likely clinfo is the best?

pvelesko:
> But likely clinfo is the best?

Wouldn't clinfo call OpenCL to find out? We can implement the necessary stuff inside CHIP-SPV. 

Or maybe just do the simple thing - wrapper call clinfo and either skip or continue with the test

PR #371, opened 2023-02-27
libCEED fixes

pvelesko:


tests/fromLibCeed/syncthreadsExitedThreads.cpp
```diff
@@ -0,0 +1,17 @@
+#include <hip/hip_runtime.h>
+
+__global__ void syncTest() { 
+ const int tid = threadIdx.x; 
+
+ if (tid > 10)
+ return;
+
+  __syncthreads();
```
pjaaskel, line 9: I don't think it's a good idea to add this test while we know we do not implement this by default due to the difference of semantics from OpenCL/SPIR-V.

pvelesko: It's in the exclusion list. I don't see any harm in having it.

pjaaskel: Ah ok, yep.

PR #382, opened 2023-03-16
Fix a shadow kernel initialized shared memory

linehill:
... which is useless and caused compilation error if total size of shared memories in the module exceeded device's limit.


pjaaskel (approved):
Look great. Should fix some of the rocPRIM cases too.

PR #392, opened 2023-04-05
Fix HipPrintf pass crashes on an invalid specifier

### LINKED ISSUE #391: printf transformation crashes if there's a lone '%'
`printf("%s %d get_prefix %\n", __func__,  __LINE__);`

Makes HipPrintf pass crash at

```C++
         while (FormatSpecCount--) {
            assert(OrigCallArgI != OrigCallArgs.end());
            Value *OrigArg = *OrigCallArgI++;
            Args.push_back(OrigArg);
          }
```

linehill:
The format string described in #391 has an invalid specifier which triggers undefined behavior. However, The compiler should not crash in such case.

Fixes #391.

### DIFF
--- a/llvm_passes/HipPrintf.cpp
+++ b/llvm_passes/HipPrintf.cpp
@@ -376,6 +376,21 @@ PreservedAnalyses HipPrintfToOpenCLPrintfPass::run(Module &Mod,
 
         IRBuilder<> B(&I);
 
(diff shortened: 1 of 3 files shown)


pjaaskel (approved):
LGTM

pvelesko:
`Unit_hipDeviceSynchronize_Positive_Nullstream` is failing for PoCL but it failed in another branch as well so merging this & will exclude in the next PR

PR #395, opened 2023-04-11
Fix #211 and others

### LINKED ISSUE #211: Synchronization of pinned memory passed in as data
This is to check/test: If we have memory that is allocated with cuHostMemAlloc() or made pinned with cuMemHostRegister() and those pointers are passed in the kernels as data (e.g. array or struct elements), do we synchronize the modifications correctly? I recall seeing code that only checks the immediate kernel arguments and synchronizes those, but doesn't check pointers in the passed data.

### LINKED ISSUE #386: Replace first touch workaround with kernel registration
https://github.com/intel/compute-runtime/issues/631

linehill:
* Fixed RegisterdVarCopy() didn’t synchronize pinned host allocations if the pointers were passed indirectly to kernels. Fixed by synchronizing all pinned host allocations known by the runtime.
* Fixed OpenCL and Level Zero backend drivers were not informed about indirect allocations.
* Fixed a test with false positive test outcome.

Fixes #211.
Closes #386.

### DIFF
--- a/src/backend/OpenCL/CHIPBackendOpenCL.cc
+++ b/src/backend/OpenCL/CHIPBackendOpenCL.cc
@@ -770,10 +770,12 @@ void CHIPModuleOpenCL::compile(CHIPDevice *ChipDev) {
       // OpenCLFunctionInfoMap",
       //                      hipErrorInitializationError);
     }
-    CHIPKernelOpenCL *ChipKernel = new CHIPKernelOpenCL(
-        std::move(Kernel), ChipDevOcl, HostFName, FuncInfo, this);
+    CHIPKernelOpenCL *ChipKernel =
+        new CHIPKernelOpenCL(Kernel, ChipDevOcl, HostFName, FuncInfo, this);
     addKernel(ChipKernel);
   }
+
+  Program_ = Program;
 }
 
 CHIPQueue *CHIPDeviceOpenCL::createQueue(CHIPQueueFlags Flags, int Priority) {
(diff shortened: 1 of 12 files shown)


pjaaskel (approved):
This looks good to me, but requires thorough testing on dGPUs too. The basic kernel analysis you mention for detecting indirect access could be worthwhile to implement since it should be straightforward and I have a gut feeling it'd handle 99% of current kernels.

linehill:
OpenCL backend needs fixing (multi-threading issues in the OpenCL BE).

linehill:
* OpenCL: Fixed invalid SVM pointers were annotated in multi-threaded applications.
* OpenCL: Fixed segfaults probably caused by cl_kernel objects being shared among threads.

linehill:
Nearly all tests on POCL-CPU fails with `CHIP critical [TID 694387] [1681808645.078377965] : OpenCL failed to initialize any devices` error. Is the installation broken in the CI?

pvelesko:
> Nearly all tests on POCL-CPU fails with `CHIP critical [TID 694387] [1681808645.078377965] : OpenCL failed to initialize any devices` error. Is the installation broken in the CI?

I was doing some maintenance earlier this morning and might have not shut off the runner service - I'll just rerun the test before investigating.

src/backend/OpenCL/CHIPBackendOpenCL.cc
```diff
@@ -938,22 +997,36 @@ CHIPEvent *CHIPQueueOpenCL::launchImpl(CHIPExecItem *ExecItem) {
...
+
+  if (SpillBuf || SvmAllocationsToKeepAlive) {
+    // Use an event call back to prolong the lifetimes of the
+    // following objects until the kernel terminates.
+    //
+    // * SpillBuffer holds an allocation referenced by the kernel
+    //   shared by exec item, which might get destroyed before the
+    //   kernel is launched/completed
+    //
+    // * Annotated SVM pointers may need to outlive the kernel
```
pjaaskel, line 1021: Isn't it down to the program logic to not free SVM allocations while they might be in use?

pjaaskel (approved):
Other than that, LGTM.

src/backend/OpenCL/CHIPBackendOpenCL.cc
linehill: All SVM allocations from all threads are annotated as we can’t tell which ones are needed for the kernel launch. One thread may free a SVM allocation while another thread still holds it.

pjaaskel: OK, I see -- annoying. Did you manage to crash an OpenCL implementation by freeing an SVM allocation while/before the kernel was running? Sounds like an undefined corner case indeed.

pjaaskel: We might want to update the OpenCL spec for this and make it defined behavior: If the SVM alloc is freed during or before, the implementation should ignore it when synchronizing the data.

pvelesko:
Cancelling CI, need to merge a fix for driver mgmt. Will restart once done

pvelesko:
Rebasing on main (CI upgrades) & running CI

Issue #418, opened 2023-05-03
hipStreamAddCallbacks() broken with OpenCL 


pjaaskel:
hipStreamSemantics test uncovers a difference-in-semantics issue with [OpenCL event callbacks](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clSetEventCallback) and [CUDA stream callbacks](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__STREAM.html#group__CUDART__STREAM_1g74aa9f4b1c2f12d994bf13876a5a2498). In OpenCL the event callbacks are called when event changes state and the event state can be changed before and during the callback execution. Furthermore, the timing of the callback execution is not specified. It can be executed before the next commands execute or much later (likely still before clFinish() returns).  CUDA's callbacks are more well-defined in their semantics, they execute only after the previous commands have finished in the stream and they block further commands executing. Thus it's more like calling clFinish() and then calling the function.

Kerilk:
If you are talking about [hipStreamAddCallback](https://rocm-developer-tools.github.io/HIP/group__Stream.html#ga3e098cd7478828b2104abb41a7bb00d3), if available I would use 
[clEnqueueNativeKernel](https://registry.khronos.org/OpenCL/sdk/2.2/docs/man/html/clEnqueueNativeKernel.html).

If not available, I would default back to what Paulius was working on, with a host side monitor on an event that spawn a task when the event is complete, and signals another one in the queue when the task is complete. This brings the implementation closer to what the level zero implementation does.

pjaaskel:
clEnqueueNativeKernel seems not be usually available unless it's a CPU device. For an implementation that should work always, I'm implementing it using a user event which the callback thread sets after the user CB is finished to ensure the ordering. Let's see if this works.

pvelesko:
> 
clEnqueueNativeKernel seems not be usually available unless it's a CPU device. For an implementation that should work always, I'm implementing it using a user event which the callback thread sets after the user CB is finished to ensure the ordering. Let's see if this works.

Sounds similar to what we're doing on the Level Zero side, no?

pjaaskel:
> Sounds similar to what we're doing on the Level Zero side, no?

OK? Didn't check. Sounds it _could_ work, although the callbacks are not supposed to run for a long time (in OpenCL), but what is long is not specified. 10s sounds long, but... ;)

PR #422, opened 2023-05-04
OpenCL: Implement CUDA callbacks using OCL CBs and user event hacks

### LINKED ISSUE #421: hipstreamsemantics occasional crash
```
malloc(): unaligned tcache chunk detected

Thread 1 "hipStreamSemant" received signal SIGABRT, Aborted.
0x00007ffff7514a7c in pthread_kill () from /lib/x86_64-linux-gnu/libc.so.6
(gdb) bt
#0  0x00007ffff7514a7c in pthread_kill () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x00007ffff74c0476 in raise () from /lib/x86_64-linux-gnu/libc.so.6
#2  0x00007ffff74a67f3 in abort () from /lib/x86_64-linux-gnu/libc.so.6
#3  0x00007ffff75076f6 in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#4  0x00007ffff751ed7c in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#5  0x00007ffff752345c in malloc () from /lib/x86_64-linux-gnu/libc.so.6
#6  0x00007ffff785d9cc in operator new(unsigned long) () from /lib/x86_64-linux-gnu/libstdc++.so.6
#7  0x00007ffff7df0580 in std::__new_allocator<CHIPQueue*>::allocate (this=0x7fffffffd370, __n=2) at /usr/lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/bits/new_allocator.h:137
#8  0x00007ffff7df0511 in std::allocator_traits<std::allocator<CHIPQueue*> >::allocate (__a=..., __n=2) at /usr/lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/bits/alloc_traits.h:464
#9  0x00007ffff7df04d0 in std::_Vector_base<CHIPQueue*, std::allocator<CHIPQueue*> >::_M_allocate (this=0x7fffffffd370, __n=2) at /usr/lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/bits/stl_vector.h:378
#10 0x00007ffff7dff8ae in std::vector<CHIPQueue*, std::allocator<CHIPQueue*> >::_M_realloc_insert<CHIPQueue* const&> (this=0x7fffffffd370, __position=..., __args=@0x5555555fabe0: 0x5555555fa960) at /usr/lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/bits/vector.tcc:453
#11 0x00007ffff7deb9ba in std::vector<CHIPQueue*, std::allocator<CHIPQueue*> >::push_back (this=0x7fffffffd370, __x=@0x5555555fabe0: 0x5555555fa960) at /usr/lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/bits/stl_vector.h:1287
#12 0x00007ffff7de60cf in CHIPBackend::findQueue (this=0x55555556f350, ChipQueue=0x5555555fa960) at /home/pvelesko/CHIP-SPV/threadJoin/src/CHIPBackend.cc:1494
#13 0x00007ffff7e4624f in hipStreamDestroy (Stream=0x5555555fa960) at /home/pvelesko/CHIP-SPV/threadJoin/src/CHIPBindings.cc:1957
#14 0x00005555555560fc in TestStreamSemantics_3 () at /home/pvelesko/CHIP-SPV/threadJoin/samples/hipStreamSemantics/hipStreamSemantics.cc:190
#15 0x000055555555688e in main () at /home/pvelesko/CHIP-SPV/threadJoin/samples/hipStreamSemantics/hipStreamSemantics.cc:267
```

pjaaskel:
hipStreamSemantics tests behavior that CUDA callbacks are asynchronously executed in the background whereas OpenCL might or might not execute them in the background. It could execute the callback in the user thread in case the command it is calling back from was finished before the CB was set.

Hack around this by creating a barrier which listens to a user event which is triggered only after setting the CB, forcing parallel execution of the callback. That is, in case the OpenCL platform supports parallel execution of callbacks the first place. I don't believe that's required to be the case, so this is down to implementation still.

Might Fix #421 and #418.

With this my local test envs now pass the expected tests once again.

### DIFF
--- a/samples/CMakeLists.txt
+++ b/samples/CMakeLists.txt
@@ -60,7 +60,6 @@ function(add_chip_test EXEC_NAME TEST_NAME TEST_PASS SOURCE)
              PASS_REGULAR_EXPRESSION "${TEST_PASS}"
              LABELS internal)
 
-
 endfunction()
 
 
--- a/samples/hipStreamSemantics/CMakeLists.txt
+++ b/samples/hipStreamSemantics/CMakeLists.txt
@@ -1,2 +1,6 @@
 
 add_chip_test(hipStreamSemantics hipStreamSemantics PASSED hipStreamSemantics.cc)
+set_tests_properties(hipStreamSemantics PROPERTIES
+  FAIL_REGULAR_EXPRESSION "Failed"
+  LABELS internal)
+
--- a/samples/hipStreamSemantics/hipStreamSemantics.cc
+++ b/samples/hipStreamSemantics/hipStreamSemantics.cc
@@ -117,7 +117,6 @@ bool TestStreamSemantics_2() {
   hipLaunchKernelGGL(addOne, 1, 1, 0, 0, dev_ptr);
   CHECK(hipGetLastError());
   CHECK(hipMemcpyAsync(host_ptr, dev_ptr, size, hipMemcpyDefault));
-  CHECK(hipStreamSynchronize(0));
   // printf("End of null stream task\n");fflush(stdout);
 
   bool testStatus = true;
--- a/src/backend/OpenCL/CHIPBackendOpenCL.cc
+++ b/src/backend/OpenCL/CHIPBackendOpenCL.cc
@@ -633,10 +633,12 @@ bool CHIPEventOpenCL::updateFinishStatus(bool ThrowErrorIfNotReady) {
     CHIPERR_LOG_AND_THROW("Event not yet ready", hipErrorNotReady);
   }
 
-  if (UpdatedStatus <= CL_COMPLETE)
+  if (UpdatedStatus <= CL_COMPLETE) {
     EventStatus_ = EVENT_STATUS_RECORDED;
-
-  return true;
+    return false;
+  } else {
+    return true;
+  }
 }
 
 float CHIPEventOpenCL::getElapsedTime(CHIPEvent *OtherIn) {
@@ -907,6 +909,7 @@ struct HipStreamCallbackData {
   hipError_t Status;
   void *UserData;
   hipStreamCallback_t Callback;
+  CHIPEventOpenCL *CallbackFinishEvent;
 };
 
 void CL_CALLBACK pfn_notify(cl_event Event, cl_int CommandExecStatus,
@@ -917,6 +920,10 @@ void CL_CALLBACK pfn_notify(cl_event Event, cl_int CommandExecStatus,
   if (Cbo->Callback == nullptr)
     return;
   Cbo->Callback(Cbo->Stream, Cbo->Status, Cbo->UserData);
+  if (Cbo->CallbackFinishEvent != nullptr) {
+    clSetUserEventStatus(Cbo->CallbackFinishEvent->ClEvent, CL_COMPLETE);
+    Cbo->CallbackFinishEvent->decreaseRefCount("Notified finished.");
+  }
   delete Cbo;
 }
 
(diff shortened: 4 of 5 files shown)


pvelesko:
What about taking this a step further and lifting the L0 callback implementation to CHIP level and then just using that generic implementation for both L0 and OCL? 

That way we don't need the OpenCL implementation to support async callback execution and we have less backend divergence.

pvelesko:
Failing for all platforms: 
```
dgpu_opencl_make_check_result.txt: FAIL
	708 - Unit_hipStreamAddCallback_WithDefaultStream (Failed)
	709 - Unit_hipStreamAddCallback_WithCreatedStream (Failed)
```

Also, if there are bugs that this PR might fix please remove them from exclusion list so that they get run through CI.

pvelesko:
rebased

pjaaskel:
> What about taking this a step further and lifting the L0 callback implementation to CHIP level and then just using that generic implementation for both L0 and OCL?
> 
> That way we don't need the OpenCL implementation to support async callback execution and we have less backend divergence.
> 
> Not as a part of this PR but as a future follow-up

Yes, that crossed my mind, but then I started to think that it might be most efficient to push execution (decisions) down to driver level as much as possible so they can optimize the parallelism and synchronizations as they see best. Here we might need to resort to our own host thread anyhow because the OpenCL callbacks are so weakly defined. Let's see if this works out or not.

> 	708 - Unit_hipStreamAddCallback_WithDefaultStream (Failed)
>	709 - Unit_hipStreamAddCallback_WithCreatedStream (Failed)

These pass for my OpenCL envs (Intel iGPU, Intel CPU, PoCL-CPU and PoCL-Level0-iGPU) for me. What is the error output?

> Also, if there are bugs that this PR might fix please remove them from exclusion list so that they get run through CI.

I do not know any immediately, but we should anyhow start going through the exclusion lists and fix things systematically when stabilizing the 1.0. Do you have hunch for some tests which could be fixed by this?

pvelesko:
> Yes, that crossed my mind, but then I started to think that it might be most efficient to push execution (decisions) down to driver level as much as possible so they can optimize the parallelism and synchronizations as they see best. Here we might need to resort to our own host thread anyhow because the OpenCL callbacks are so weakly defined. Let's see if this works out or not.

Well if OpenCL doesn't guarantee async execution we kind of are bound by the HIP API to use a callback executor thread, no? 

> I do not know any immediately, but we should anyhow start going through the exclusion lists and fix things systematically when stabilizing the 1.0. Do you have hunch for some tests which could be fixed by this?

I have a workflow that will run for days testing each failure in the exclusion list for resolution.

pjaaskel:
> Well if OpenCL doesn't guarantee async execution we kind of are bound by the HIP API to use a callback executor thread, no?

Not sure of this, since I didn't find a mention in the CUDA PM that guarantees it either. There are other similar spots in terms of concurrency "assumptions" we discussed some 6 months ago that might not be true when running on various OpenCL implementations. I'm not sure if e.g. OpenCL mandates command queue async execution at all. We've created an implementation that doesn't need multithreading (or even fibers) for very simple microcontroller deployments (PoCL-basic), and it should be workable (the current implementation has a couple of issues though).

pjaaskel:
hipStreamAddCallback seems fishy, it lacks volatile in a global it uses to signal completion.

Pls try if this helps (it still works here):

```C++
diff --git a/tests/catch/unit/stream/hipStreamAddCallback.cc b/tests/catch/unit/stream/hipStreamAddCallback.cc
index 279f372d..b1b97099 100644
--- a/tests/catch/unit/stream/hipStreamAddCallback.cc
+++ b/tests/catch/unit/stream/hipStreamAddCallback.cc
@@ -39,7 +39,7 @@ Testcase Scenarios :
 namespace hipStreaAddCallbackTest {
 size_t NSize = 4 * 1024 * 1024;
 float *A_h, *C_h;
-bool gcbDone = false;
+volatile bool gcbDone = false;
 bool gPassed = true;
 void *ptr0xff = reinterpret_cast<void *>(0xffffffff);
 void *gusrptr;
```

pvelesko:
@pjaaskel  I ran the failed tests manually and they passed. On the CI pass they failed via a correctness issue.

pvelesko:
Failed again in the CI env:

```
 78/806 Test #708: Unit_hipStreamAddCallback_WithDefaultStream ...............................***Failed    0.70 sec
Filters: Unit_hipStreamAddCallback_WithDefaultStream
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
hipStreamAddCallback is a Catch v2.13.4 host application.
Run with -? for options
-------------------------------------------------------------------------------
Unit_hipStreamAddCallback_WithDefaultStream
-------------------------------------------------------------------------------
/home/pvelesko/actions-runner/_work/chip-spv/chip-spv/HIP/tests/catch/unit/stream/hipStreamAddCallback.cc:230
...............................................................................
/home/pvelesko/actions-runner/_work/chip-spv/chip-spv/HIP/tests/catch/unit/stream/hipStreamAddCallback.cc:233: FAILED:
  REQUIRE( TestPassed )
with expansion:
  false
with message:
  Data mismatch :58364
===============================================================================
test cases: 1 | 1 failed
assertions: 4 | 3 passed | 1 failed
        Start 650: Unit_hipMemsetFunctional_ZeroValue_3D
 79/806 Test #709: Unit_hipStreamAddCallback_WithCreatedStream ...............................***Failed    0.75 sec
Filters: Unit_hipStreamAddCallback_WithCreatedStream
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
hipStreamAddCallback is a Catch v2.13.4 host application.
Run with -? for options
-------------------------------------------------------------------------------
Unit_hipStreamAddCallback_WithCreatedStream
-------------------------------------------------------------------------------
/home/pvelesko/actions-runner/_work/chip-spv/chip-spv/HIP/tests/catch/unit/stream/hipStreamAddCallback.cc:239
...............................................................................
/home/pvelesko/actions-runner/_work/chip-spv/chip-spv/HIP/tests/catch/unit/stream/hipStreamAddCallback.cc:242: FAILED:
  REQUIRE( TestPassed )
with expansion:
  false
with message:
  Data mismatch :0
```

I'll try your patch but it's very strange that I can't reproduce it outside of CI

pjaaskel:
Yes, strange. I'd expect the same binary to fail in this case the same way as the volatile should be optimized by the compiler if something.

BTW, this is the description related to callbacks in CUDA in the concurrency Chapter:  https://docs.nvidia.com/cuda/archive/10.0/cuda-c-programming-guide/index.html#stream-callbacks

It doesn't mention the CBs should run in parallel. In the current implementation they usually should do, or they'll run in a blocking fashion when the holding user event is set CL_COMPLETE. I'd guess all implementations than PoCL-basic will execute it in parallel, as expected by the hipStreamSemantics tests (which might tests semantics that is not guaranteed to be there, if I read it right).

pjaaskel:
Is it happening only with dGPU?

pvelesko:
> Is it happening only with dGPU?

yes dGPU OpenCL. You can checkout the logs yourself.

pvelesko:
> The log is not very helpful. But I'll test with a dGPU when I get to one.

Would it be helpful to get access to the CI machine to test this?

pjaaskel:
OK, I reproduced it successfully in your box and fixed the problem (I think).

While testing, I noticed hipStreamSemantics is falsely detected as passing as it likely scans for PASSED but this test has multiple tests, and some of them currently fail (without this patch):

```
ctest -R hipStreamSemantics  -V
UpdateCTestConfiguration  from :/home/pjaaskel/src/chip-spv/build/DartConfiguration.tcl
UpdateCTestConfiguration  from :/home/pjaaskel/src/chip-spv/build/DartConfiguration.tcl
Test project /home/pjaaskel/src/chip-spv/build
Constructing a list of tests
Done constructing a list of tests
Updating test list for fixtures
Added 0 tests to meet fixture requirements
Checking test dependency graph...
Checking test dependency graph end
test 1309
    Start 1309: hipStreamSemantics

1309: Test command: /home/pjaaskel/src/chip-spv/build/samples/hipStreamSemantics/hipStreamSemantics
1309: Test timeout computed to be: 10000000
1309: ------------------------------------------------------------------
1309: TestStreamSemantics_1(stream query) : PASSED
1309: callback_sleep10: Going to sleep for 10sec
1309: ------------------------------------------------------------------
1309: TestStreamSemantics_2 (non-blocking stream): PASSED
1309: ------------------------------------------------------------------
1309: Going to sync stream1
1309: Going to call query
1309: TestStreamSemantics_3(independent execution) :  Failed 
1309: ------------------------------------------------------------------
1309: Going to destroy a stream which has pending work
1309: Waiting for device to finish the task
1309: TestStreamSemantics_4 (stream destroy): PASSED
1309: ------------------------------------------------------------------
1309: TestStreamSemantics_5 (sync streams):  Failed 
1/1 Test #1309: hipStreamSemantics ...............   Passed    0.57 sec

The following tests passed:
        hipStreamSemantics

100% tests passed, 0 tests failed out of 1
```

pjaaskel:
> While testing, I noticed hipStreamSemantics is falsely detected as passing as it likely scans for PASSED but this test has multiple tests, and some of them currently fail (without this patch):

Added a FAIL regexp for this.

pvelesko:
> OK, I reproduced it successfully in your box and fixed the problem (I think).

you mentioned hipStreamSemantics but the only two tests that were failing were 

```
Unit_hipStreamAddCallback_WithDefaultStream
Unit_hipStreamAddCallback_WithCreatedStream
```

pvelesko:
ok seems that they're passing now

pvelesko:
only `cuda-convolutionSeparable` test failed for Level Zero. Merging this & will investigate that failure separately.

pjaaskel:
I think I didn't touch anything which should affect the L0 BE, I hope.

pjaaskel:
Ah, the test fix of course could cause LZ to fail if it handles it wrongly.

PR #432, opened 2023-05-09
MemMap Fixes

### LINKED ISSUE #430: [opencl] hipStreamPerThread_basic failing for OpenCL dGPU
The following change in the test makes the test pass:

```
//   HIP_CHECK(hipHostMalloc(&hostMem, size));
  HIP_CHECK(hipMalloc(&devMem, size));
```

pvelesko:
* ~~Fix assert that event use after delete (recycled events would not get their `Deleted_` reset~~
* Registered memory map/unmap was not getting applied to async memcpy
* Registered memcopy mapped to `HOST_READ` in all cases instead of `HOST_WRITE` seems that `HOST_WRITE` doesn't guarantee a proper read whereas using `HOST_READ` allows to both write and read host pointers. 

fixes #430

### DIFF
--- a/src/CHIPBackend.hh
+++ b/src/CHIPBackend.hh
@@ -2020,6 +2020,7 @@ public:
   enum MEM_MAP_TYPE {
     HOST_READ,
     HOST_WRITE,
+    HOST_READ_WRITE
   };
   virtual void MemMap(const AllocationInfo *AllocInfo, MEM_MAP_TYPE MapType) {}
   virtual void MemUnmap(const AllocationInfo *AllocInfo) {}
--- a/src/CHIPBackend.hh
+++ b/src/CHIPBackend.hh
@@ -607,10 +607,7 @@ protected:
   CHIPEventFlags Flags_;
   std::vector<CHIPEvent *> DependsOnList;
 
-#ifndef NDEBUG
-  // A debug flag for cathing use-after-delete.
-  bool Deleted_ = false;
-#endif
+
 
   // reference count
   size_t *Refc_;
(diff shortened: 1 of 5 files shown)


src/backend/Level0/CHIPBackendLevel0.hh
```diff
@@ -305,8 +305,10 @@ public:
     // go through all pools and try to get an allocated event
     for (size_t i = 0; i < EventPools_.size(); i++) {
       CHIPEventLevel0 *Event = EventPools_[i]->getEvent();
-      if (Event)
+      if (Event) {
+        Event->Deleted_ = false;
```
linehill, line 309: This is redundant. The `Deleted_` flag is reset by the `getEvent()`call in the line 307 (LZEventPool::getEvent() -> CHIPEventLevel0::reset() -> CHIPEventLevel0::markDeleted(false)).

pvelesko: wasn't working for Release, I believe. I was getting "event use after delete" errors. I need to clean up this branch, though.

src/CHIPBackend.hh
```diff
@@ -607,10 +607,7 @@ protected:
   CHIPEventFlags Flags_;
   std::vector<CHIPEvent *> DependsOnList;
 
-#ifndef NDEBUG
-  // A debug flag for cathing use-after-delete.
-  bool Deleted_ = false;
-#endif
+
```
linehill, line 610: This is not needed if (1) is reverted.

src/CHIPBackend.hh
```diff
@@ -629,6 +626,10 @@ protected:
   CHIPEvent() = default;
 
 public:
+// #ifndef NDEBUG
+  // A debug flag for cathing use-after-delete.
+  bool Deleted_ = false;
+// #endif
```
linehill, line 632: This is not needed if (1) is reverted.

PR #434, opened 2023-05-10
Graph fixes + other stuff

franz:
* fixes in the Graph API (unfinished)
* "native" Graph support (OpenCL command-buffers or LevelZero command-lists)
* code cleanup in the OpenCL backend
* support for ``cl_intel_unified_shared_memory``
* updated some benchmarks

approx 50-60 additional ``Unit_hipGraph`` tests pass


src/CHIPBackend.cc
```diff
@@ -495,7 +463,9 @@ void *CHIPArgSpillBuffer::allocate(const SPVFuncInfo::Arg &Arg) {
 // CHIPExecItem
 //*************************************************************************************
 void CHIPExecItem::copyArgs(void **Args) {
-  for (int i = 0; i < getNumArgs(); i++) {
+  // args need to be set up again
+  ArgsSetup = false;
+  for (size_t i = 0; i < getNumArgs(); i++) {
     Args_.push_back(Args[i]);
   }
```
linehill, line 470: Style nit:
```suggestion
  for (size_t i = 0; i < getNumArgs(); i++)
    Args_.push_back(Args[i]);
```

src/CHIPBindings.cc
```diff
@@ -334,21 +350,42 @@ hipError_t hipGraphGetEdges(hipGraph_t graph, hipGraphNode_t *from,
 hipError_t hipGraphGetNodes(hipGraph_t graph, hipGraphNode_t *nodes,
                             size_t *numNodes) {
   CHIP_TRY
+  if (!graph || !numNodes)
+    RETURN(hipErrorInvalidValue);
+  //  if (!nodes && !numNodes)
+  //    RETURN(hipErrorInvalidValue);
```
linehill, line 356: Missed clean-up?

cmake/UnitTests.cmake
```diff
@@ -150,6 +150,74 @@ list(APPEND FAILING_FOR_ALL "Unit_deviceFunctions_CompileTest_atomicXor_system_u
 list(APPEND FAILING_FOR_ALL "hipStreamSemantics") # SEGFAULT - likely due to main thread exiting without calling join
 list(APPEND FAILING_FOR_ALL "Unit_hipMultiStream_multimeDevice") # SEGFAULT - likely due to multiple GPU support
 
+list(APPEND FAILING_FOR_ALL "Unit_hipGraphAddDependencies_NegTest") # (Failed)
+list(APPEND FAILING_FOR_ALL "Unit_hipGraphAddEventRecordNode_MultipleRun") # subprocess aborted
+list(APPEND FAILING_FOR_ALL "Unit_hipGraphAddEventWaitNode_MultipleRun") # subprocess aborted
+list(APPEND FAILING_FOR_ALL "Unit_hipGraph_BasicFunctional") # subprocess aborted
+list(APPEND FAILING_FOR_ALL "Unit_hipGraph_SimpleGraphWithKernel") # subprocess aborted
```
pvelesko, line 157: lots of failing tests for graphs still? Even the most basic tests such as `Unit_hipGraph_BasicFunctional`?

src/CHIPBindings.cc
```diff
@@ -334,21 +350,42 @@ hipError_t hipGraphGetEdges(hipGraph_t graph, hipGraphNode_t *from,
...
   CHIPInitialize();
   auto Nodes = GRAPH(graph)->getNodes();
-  *nodes = *(Nodes.data());
-  *numNodes = GRAPH(graph)->getNodes().size();
+  if (nodes) {
+    if (numNodes && (*numNodes > Nodes.size()))
```
linehill, line 360: Redundant `numNodes` nullptr check?

src/CHIPBindings.cc
```diff
@@ -285,6 +289,10 @@ hipError_t hipGraphAddDependencies(hipGraph_t graph, const hipGraphNode_t *from,
                                    const hipGraphNode_t *to,
                                    size_t numDependencies) {
   CHIP_TRY
+  if (!graph)
+    RETURN(hipErrorInvalidValue);
+  if (!from || !to)
+    RETURN(hipErrorInvalidValue);
```
linehill, line 295: Unimportant nit:
```suggestion
  if (!graph || !from || !to)
    RETURN(hipErrorInvalidValue);
```

src/backend/OpenCL/CHIPBackendOpenCL.hh
```diff
@@ -86,30 +84,46 @@ public:
 };
 
 class CHIPEventOpenCL : public CHIPEvent {
-public:
-  cl_event ClEvent;
-  friend class CHIPEventOpenCL;
+private:
+  cl::Event ClEvent;
```
pvelesko, line 88: style: non-public member missing `_` postfix.

src/CHIPGraph.cc
```diff
@@ -53,6 +53,18 @@ void CHIPGraphNode::DFS(std::vector<CHIPGraphNode *> CurrPath,
...
+void CHIPGraphNode::checkDependencies(size_t numDependencies,
+                                      CHIPGraphNode **pDependencies) {
+  if (numDependencies > 0 && pDependencies == nullptr) {
+    CHIPERR_LOG_AND_THROW("numDependencies > 0 && pDependencies == nullptr",
+                          hipErrorInvalidValue);
+  }
+  if (numDependencies == 0 && pDependencies != nullptr) {
+    CHIPERR_LOG_AND_THROW("numDependencies == 0 && pDependencies != nullptr",
+                          hipErrorInvalidValue);
+  }
```
linehill, line 65: Style nit:
```suggestion
  if (numDependencies > 0 && pDependencies == nullptr)
    CHIPERR_LOG_AND_THROW("numDependencies > 0 && pDependencies == nullptr",
                          hipErrorInvalidValue);
  if (numDependencies == 0 && pDependencies != nullptr)
    CHIPERR_LOG_AND_THROW("numDependencies == 0 && pDependencies != nullptr",
                          hipErrorInvalidValue);
```

src/backend/OpenCL/CHIPBackendOpenCL.hh
```diff
@@ -86,30 +84,46 @@ public:
...
+  void reset(cl::Event &&Ev) {
+    ClEvent = Ev;
+    logTrace("Event {} Moved into {} || NOW: {}", (void *)Ev.get(),
+             (void *)this, (void *)ClEvent.get());
+  }
+  void reset(cl_event Ev) {
+    ClEvent = Ev;
+    logTrace("Event {} Moved into {} ||| NOW: {}", (void *)Ev, (void *)this,
+             (void *)ClEvent.get());
+  }
```
pvelesko, line 123: Why does event reset take any arguments at all, especially a new event? Should be `init()` or something similar.

src/backend/OpenCL/CHIPBackendOpenCL.hh
```diff
@@ -86,30 +84,46 @@ public:
...
+
+  // for elapsedTime
+  uint64_t getFinishTime();
```
pvelesko, line 126: Seems like should be part of `CHIPEvent` and overriden here

src/CHIPGraph.cc
```diff
@@ -184,24 +204,69 @@ void CHIPGraph::removeNode(CHIPGraphNode *Node) {
...
-    auto Nodes = ExecQueueCopy.front();
-    std::string NodesInThisLevel = "";
-    for (auto Node : Nodes) {
-      NodesInThisLevel += Node->Msg + " ";
+bool CHIPGraphExec::tryLaunchNative(CHIPQueue *Queue) {
+  bool UsedNativeGraph = false;
+  if (NativeGraph) {
+    if (NativeGraph->isFinalized()) {
+      logDebug("NativeGraph: launching existing graph");
+    } else {
```
linehill, line 212: Nit:
```suggestion
    if (NativeGraph->isFinalized())
      logDebug("NativeGraph: launching existing graph");
    else {
```

src/backend/OpenCL/CHIPBackendOpenCL.hh
```diff
@@ -154,38 +177,76 @@ public:
...
+
+public:
+  bool allDevicesSupportFineGrainSVMorUSM();
+  CHIPContextOpenCL(cl::Context ClContext, cl::Device Dev, cl::Platform Plat);
   virtual ~CHIPContextOpenCL() {}
   void *allocateImpl(size_t Size, size_t Alignment, hipMemoryType MemType,
                      CHIPHostAllocFlags Flags = CHIPHostAllocFlags()) override;
 
   bool isAllocatedPtrMappedToVM(void *Ptr) override { return false; } // TODO
+  const SVMemoryRegion &getRegion() const { return SvmMemory; }
```
pvelesko, line 228: Bit confusing

```suggestion
  const SVMemoryRegion &getMemRegion() const { return SvmMemory; }
```

src/CHIPGraph.cc
```diff
@@ -184,24 +204,69 @@ void CHIPGraph::removeNode(CHIPGraphNode *Node) {
...
+    if (!NativeGraph)
+      return false;
+
+    for (auto &Node : OriginalGraph_->getNodes()) {
+      if (!NativeGraph->addNode(Node)) {
+        logError("NativeGraph: failed to add node of type: {}",
+                 Node->getType());
+        return false;
+      }
     }
```
linehill, line 228: Style nit:
```suggestion
    for (auto &Node : OriginalGraph_->getNodes())
      if (!NativeGraph->addNode(Node))
        logError("NativeGraph: failed to add node of type: {}",
                 Node->getType());
        return false;
```

src/CHIPGraph.cc
```diff
@@ -184,24 +204,69 @@ void CHIPGraph::removeNode(CHIPGraphNode *Node) {
...
+  }
+
+  assert(NativeGraph->isFinalized());
+  if (Queue->enqueueNativeGraph(NativeGraph.get())) {
+    logDebug("NativeGraph: launched");
+    Queue->finish();
+    return true;
+  } else {
+    return false;
+  }
```
linehill, line 243: Style nit:
```suggestion
  } else
    return false;
```

src/backend/OpenCL/SVMemoryRegion.cc
```diff
@@ -24,33 +24,75 @@
...
+    case hipMemoryTypeHost:
+      Ptr = USM.clHostMemAllocINTEL(Context_(), NULL, Size, Alignment, &Err);
+      break;
+    case hipMemoryTypeDevice:
+      Ptr = USM.clDeviceMemAllocINTEL(Context_(), Device_(), NULL, Size,
+                                      Alignment, &Err);
+      break;
+    case hipMemoryTypeManaged:
+    case hipMemoryTypeUnified:
+    */
```
pvelesko, line 67: please open an issue to track this

src/CHIPGraph.cc
linehill, line 243: Style nit: [no else after return](https://llvm.org/docs/CodingStandards.html#don-t-use-else-after-a-return):
```suggestion
  if (Queue->enqueueNativeGraph(NativeGraph.get())) {
    logDebug("NativeGraph: launched");
    Queue->finish();
    return true;
  }
  return false;
```

src/backend/OpenCL/CHIPBackendOpenCL.cc
```diff
@@ -544,36 +530,21 @@ CHIPEventOpenCL::CHIPEventOpenCL(CHIPContextOpenCL *ChipContext,
...
-size_t CHIPEventOpenCL::getRefCount() {
-  cl_uint RefCount;
-  if (ClEvent == nullptr)
-    return 0;
-  int Status = clGetEventInfo(getNativeRef(), CL_EVENT_REFERENCE_COUNT, 4,
-                              &RefCount, NULL);
-  CHIPERR_CHECK_LOG_AND_THROW(Status, CL_SUCCESS, hipErrorTbd);
-  return RefCount;
+  return Ret;
 }
```
pvelesko, line 547: losing refcount on OpenCL

src/CHIPGraph.cc
```diff
@@ -184,24 +204,69 @@ void CHIPGraph::removeNode(CHIPGraphNode *Node) {
...
+    logDebug("NativeGraph: failed to construct/finalize/launch, using the "
+             "original code path");
+    compile();
+    auto ExecQueueCopy = ExecQueues_;
+    while (ExecQueueCopy.size()) {
+      auto Nodes = ExecQueueCopy.front();
+      std::string NodesInThisLevel = "";
+      for (auto Node : Nodes) {
+        NodesInThisLevel += Node->Msg + " ";
+      }
```
linehill, line 260: Style nit:
```suggestion
      for (auto Node : Nodes)
        NodesInThisLevel += Node->Msg + " ";
```

src/CHIPGraph.hh
```diff
@@ -119,6 +121,10 @@ public:
    * @param TheNode
    */
   void addDependency(CHIPGraphNode *TheNode) {
+    if (TheNode == nullptr) {
+      CHIPERR_LOG_AND_THROW("addDependency called with nullptr",
+                            hipErrorInvalidValue);
+    }
```
linehill, line 127: Style nit:
```suggestion
    if (TheNode == nullptr)
      CHIPERR_LOG_AND_THROW("addDependency called with nullptr",
                            hipErrorInvalidValue);
```

src/CHIPBackend.cc
```diff
@@ -220,36 +218,6 @@ void CHIPEvent::releaseDependencies() {
   DependsOnList.clear();
 }
```
pvelesko, line 220: see previous comments about moving refcounting

src/CHIPBindings.cc
```diff
@@ -630,7 +694,12 @@ hipError_t hipGraphKernelNodeGetParams(hipGraphNode_t node,
                                        hipKernelNodeParams *pNodeParams) {
   CHIP_TRY
   CHIPInitialize();
-  *pNodeParams = ((CHIPGraphNodeKernel *)node)->getParams();
+  if (!node || !pNodeParams)
+    RETURN(hipErrorInvalidValue);
+  auto CastNode = static_cast<CHIPGraphNodeKernel *>(node);
```
pvelesko, line 699: we should probably check the result of the cast?

src/CHIPBindings.cc
```diff
@@ -705,9 +776,14 @@ hipError_t hipGraphMemcpyNodeGetParams(hipGraphNode_t node,
...
   CHIP_TRY
   CHIPInitialize();
-  hipMemcpy3DParms Params =
-      static_cast<CHIPGraphNodeMemcpy *>(node)->getParams();
-  pNodeParams = &Params;
+  if (!node || !pNodeParams)
+    RETURN(hipErrorInvalidValue);
+  auto CastNode = static_cast<CHIPGraphNodeMemcpy *>(node);
+  if (CastNode->getType() != hipGraphNodeTypeMemcpy)
+    CHIPERR_LOG_AND_THROW("Node is not Memcpy", hipErrorInvalidValue);
```
pvelesko, line 783: if the cast failed, woudln't CastNode be null?

src/backend/OpenCL/CHIPBackendOpenCL.cc
```diff
@@ -584,25 +555,37 @@ CHIPEventOpenCL *CHIPBackendOpenCL::createCHIPEvent(CHIPContext *ChipCtx,
...
-    this->Msg = Other->Msg;
+size_t CHIPEventOpenCL::getCHIPRefc() {
+  int Err = CL_SUCCESS;
+  size_t RefC = ClEvent.getInfo<CL_EVENT_REFERENCE_COUNT>(&Err);
+  if (Err != CL_SUCCESS) {
+    logError("failed to get Reference count from OpenCL event");
+    return 0;
+  } else {
+    return RefC;
   }
```
linehill, line 588: Style nit (no else after return):
```suggestion
  if (Err != CL_SUCCESS) {
    logError("failed to get Reference count from OpenCL event");
    return 0;
  }
  return RefC;
```

src/CHIPGraph.cc
```diff
@@ -150,21 +173,18 @@ CHIPGraphNodeKernel::CHIPGraphNodeKernel(const void *HostFunction, dim3 GridDim,
...
-
-  ExecItem_->copyArgs(Args);
-  ExecItem_->setupAllArgs();
+  ExecItem_ = Backend->createCHIPExecItem(Params_.gridDim, Params_.blockDim,
+                                          Params_.sharedMemBytes, nullptr);
+  ExecItem_->setKernel(Kernel_);
 }
 
-int NodeCounter = 1;
+static unsigned NodeCounter = 1;
```
pvelesko, line 186: bit of a random placement and no explanation

src/CHIPGraph.cc
```diff
@@ -184,24 +204,69 @@ void CHIPGraph::removeNode(CHIPGraphNode *Node) {
...
+}
+
+void CHIPGraphExec::launch(CHIPQueue *Queue) {
+  logDebug("{} CHIPGraphExec::launch({})", (void *)this, (void *)Queue);
+  bool UsedNativeGraph = tryLaunchNative(Queue);
+
+  if (!UsedNativeGraph) {
+    logDebug("NativeGraph: failed to construct/finalize/launch, using the "
+             "original code path");
+    compile();
```
pvelesko, line 253: Maybe add comment on why the new code path can fail

pvelesko (changes requested):
comments

src/backend/OpenCL/CHIPBackendOpenCL.cc
```diff
@@ -871,25 +816,130 @@ CHIPKernelOpenCL::CHIPKernelOpenCL(cl::Kernel ClKernel, CHIPDeviceOpenCL *Dev,
...
+            Plat(), "clDeviceMemAllocINTEL");
+    Exts.USM.clHostMemAllocINTEL =
+        (clHostMemAllocINTEL_fn)::clGetExtensionFunctionAddressForPlatform(
+            Plat(), "clHostMemAllocINTEL");
+    Exts.USM.clMemFreeINTEL =
+        (clMemFreeINTEL_fn)::clGetExtensionFunctionAddressForPlatform(
+            Plat(), "clMemFreeINTEL");
+  } else {
+    logDebug("Device does not support Intel USM");
+  }
```
linehill, line 928: Style nit:
```suggestion
  } else
    logDebug("Device does not support Intel USM");
```

src/backend/OpenCL/CHIPBackendOpenCL.cc
```diff
@@ -871,25 +816,130 @@ CHIPKernelOpenCL::CHIPKernelOpenCL(cl::Kernel ClKernel, CHIPDeviceOpenCL *Dev,
...
+  cl_device_svm_capabilities DeviceSVMCapabilities;
+  int Err = Dev.getInfo(CL_DEVICE_SVM_CAPABILITIES, &DeviceSVMCapabilities);
+  CHIPERR_CHECK_LOG_AND_THROW(Err, CL_SUCCESS, hipErrorTbd);
+  SupportsFineGrainSVM =
+      DeviceSVMCapabilities & CL_DEVICE_SVM_FINE_GRAIN_BUFFER;
+  if (SupportsFineGrainSVM) {
+    logTrace("Device supports fine grain SVM");
+  } else {
+    logTrace("Device does not support fine grain SVM");
+  }
```
linehill, line 939: Style nit:
```suggestion
  if (SupportsFineGrainSVM)
    logTrace("Device supports fine grain SVM");
  else
    logTrace("Device does not support fine grain SVM");
```

src/backend/OpenCL/CHIPBackendOpenCL.cc
```diff
@@ -1260,46 +1320,430 @@ CHIPQueueOpenCL::enqueueBarrierImpl(std::vector<CHIPEvent *> *EventsToWaitFor) {
...
+bool CHIPGraphNativeOpenCL::finalize() {
+  int Status = Exts->clFinalizeCommandBufferKHR(Handle);
+  if (Status == CL_SUCCESS) {
+    Finalized = true;
+    return true;
+  } else {
+    logError("clFinalizeCommandBufferKHR FAILED with status {}",
+             resultToString(Status));
+    return false;
+  }
```
linehill, line 1469: Style nit (no else after return):
```suggestion
  if (Status == CL_SUCCESS) {
    Finalized = true;
    return true;
  }
  logError("clFinalizeCommandBufferKHR FAILED with status {}",
           resultToString(Status));
  return false;
```

src/backend/OpenCL/CHIPBackendOpenCL.cc
```diff
@@ -1260,46 +1320,430 @@ CHIPQueueOpenCL::enqueueBarrierImpl(std::vector<CHIPEvent *> *EventsToWaitFor) {
...
+    if (Dst == Src) {
+      Status = Exts->clCommandBarrierWithWaitListKHR(
+          Handle, nullptr, SyncPointDeps.size(), SyncPointDeps.data(),
+          SyncPoint, nullptr);
+      CHIPERR_CHECK_LOG_AND_THROW(Status, CL_SUCCESS, hipErrorTbd);
+    } else {
+      Status = Exts->clCommandSVMMemcpyPOCL(
+          Handle, nullptr, Dst, Src, Size, SyncPointDeps.size(),
+          SyncPointDeps.data(), SyncPoint, nullptr);
+    }
```
linehill, line 1600: Style nit:
```suggestion
    } else
      Status = Exts->clCommandSVMMemcpyPOCL(
          Handle, nullptr, Dst, Src, Size, SyncPointDeps.size(),
          SyncPointDeps.data(), SyncPoint, nullptr);
```

linehill, line 1600: Style nit:
```suggestion
    } else
      Status = Exts->clCommandSVMMemcpyPOCL(
          Handle, nullptr, Dst, Src, Size, SyncPointDeps.size(),
          SyncPointDeps.data(), SyncPoint, nullptr);
```

src/backend/OpenCL/CHIPBackendOpenCL.cc
```diff
@@ -1260,46 +1320,430 @@ CHIPQueueOpenCL::enqueueBarrierImpl(std::vector<CHIPEvent *> *EventsToWaitFor) {
...
+    std::vector<cl_sync_point_khr> &SyncPointDeps,
+    cl_sync_point_khr *SyncPoint) {
+  if (!Exts->clCommandSignalEventPOCL)
+    return false;
+
+  CHIPEvent *E = Node->getEvent();
+  CHIPEventOpenCL *CLE = static_cast<CHIPEventOpenCL *>(E);
+
+  int Status;
+  // TODO BROKEN
```
linehill, line 1713: Perhaps clarify? (Is the driver broken or is the design broken?)

src/backend/OpenCL/CHIPBackendOpenCL.cc
```diff
@@ -1260,46 +1320,430 @@ CHIPQueueOpenCL::enqueueBarrierImpl(std::vector<CHIPEvent *> *EventsToWaitFor) {
...
+  Status = Exts->clCommandSignalEventPOCL(Handle, nullptr, nullptr, SyncPoint,
+                                          nullptr);
+  return Status == CL_SUCCESS;
+}
+
+// DONE
```
linehill, line 1719: Unneeded comment?

src/backend/OpenCL/CHIPBackendOpenCL.hh
```diff
@@ -154,38 +177,76 @@ public:
...
   virtual void freeImpl(void *Ptr) override;
-  cl::Context *get();
+  cl::Context &get() { return ClContext; }
+  bool supportsCommandBuffers() { return SupportsCommandBuffers; }
+  bool supportsCommandBuffersSVM() { return SupportsCommandBuffersSVM; }
+  bool supportsCommandBuffersHost() { return SupportsCommandBuffersHost; }
+  const CHIPContextClExts *exts() { return &Exts; }
```
linehill, line 234: Would recommend:
```suggestion
  bool supportsCommandBuffers() const { return SupportsCommandBuffers; }
  bool supportsCommandBuffersSVM() const { return SupportsCommandBuffersSVM; }
  bool supportsCommandBuffersHost() const { return SupportsCommandBuffersHost; }
  const CHIPContextClExts *exts() const { return &Exts; }
```

src/backend/OpenCL/CHIPBackendOpenCL.hh
```diff
@@ -285,9 +350,10 @@ public:
                    CHIPModuleOpenCL *Parent);
 
   virtual ~CHIPKernelOpenCL() {}
-  SPVFuncInfo *getFuncInfo() const;
-  std::string getName();
-  cl::Kernel *get();
+
+  SPVFuncInfo *getFuncInfo() const { return FuncInfo_; }
+  std::string getName() { return Name_; }
```
linehill, line 355: Avoid copy too?
```suggestion
  const std::string &getName() const { return Name_; }
```

src/backend/OpenCL/SVMemoryRegion.cc
```diff
@@ -24,33 +24,75 @@
...
+      Ptr = USM.clSharedMemAllocINTEL(Context_(), Device_(), NULL, Size,
+                                      Alignment, &Err);
+      break;
+    }
+  } else if (SupportsFineGrain) {
     Ptr = ::clSVMAlloc(
         Context_(), CL_MEM_READ_WRITE | CL_MEM_SVM_FINE_GRAIN_BUFFER, Size, 0);
+  } else {
+    Ptr = ::clSVMAlloc(Context_(), CL_MEM_READ_WRITE, Size, 0);
   }
```
linehill, line 78: Style nit:
```suggestion
  else if (SupportsFineGrain)
    Ptr = ::clSVMAlloc(
        Context_(), CL_MEM_READ_WRITE | CL_MEM_SVM_FINE_GRAIN_BUFFER, Size, 0);
  else
    Ptr = ::clSVMAlloc(Context_(), CL_MEM_READ_WRITE, Size, 0);
```

src/CHIPBackend.hh
```diff
@@ -639,9 +636,9 @@ public:
   CHIPEventFlags getFlags() { return Flags_; }
   std::mutex EventMtx;
   std::string Msg;
-  size_t getCHIPRefc();
-  virtual void decreaseRefCount(std::string Reason);
-  virtual void increaseRefCount(std::string Reason);
+  virtual size_t getCHIPRefc() = 0;
```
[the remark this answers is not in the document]
linehill: To me, moving these down to Level0 BE seems sensible as OpenCL BE does not need to implement these.

src/backend/OpenCL/CHIPBackendOpenCL.cc
```diff
@@ -1260,46 +1320,430 @@ CHIPQueueOpenCL::enqueueBarrierImpl(std::vector<CHIPEvent *> *EventsToWaitFor) {
 #endif
   CHIPEventOpenCL *Event =
       (CHIPEventOpenCL *)Backend->createCHIPEvent(this->ChipContext_);
-  cl_int RefCount;
+
   int Status;
-  Status = clGetEventInfo(Event->getNativeRef(), CL_EVENT_REFERENCE_COUNT, 4,
-                          &RefCount, NULL);
+  cl::Event RetEv;
+
```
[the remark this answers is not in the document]
linehill: cl::Event (and other cl:: objects) does do reference counting and automatically release the resources when not needed anymore.

src/CHIPBackend.hh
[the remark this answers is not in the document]
pvelesko: but we can use native OpenCL refcount tracking to verify that ours is working correctly, plus we want to have as much code as possible in the shared code paths - the backend implementations should be as thin as possible and use CHIP functions when possible.

src/CHIPBindings.cc
linehill: Static cast is not a runtime check and the check results in undefined behavior (UB) if the `node` is not a subclass of `hipGraphNode_t`. Such case to occur probably means that the user does something which results in UB before the call.

src/CHIPBindings.cc
franz: As Henry mentioned above, `static_cast` either fails at compile time, or doesn't fail at all. Null from failed cast is only returned by `dynamic_cast`, but using that requires compiling all code with RTTI enabled. There is a graph test that tries to call some API with node object of wrong typed, and `static_cast` didn't return null. For this reason i changed these checks to call `getType()` and compare.

src/CHIPGraph.cc
franz: fixed

src/backend/OpenCL/CHIPBackendOpenCL.hh
franz: fixed

src/CHIPBindings.cc
franz: yep fixed

src/CHIPBindings.cc
franz: fixed

src/backend/OpenCL/CHIPBackendOpenCL.cc
franz: fixed

src/backend/OpenCL/CHIPBackendOpenCL.cc
franz: fixed

src/backend/OpenCL/CHIPBackendOpenCL.cc
```diff
@@ -1135,10 +1187,10 @@ CHIPQueueOpenCL::CHIPQueueOpenCL(CHIPDevice *ChipDevice, int Priority,
     logWarn("CHIPQueueOpenCL is ignoring Priority value");
 
   if (Queue)
-    ClQueue_ = new cl::CommandQueue(Queue);
+    ClQueue = cl::CommandQueue(Queue);
```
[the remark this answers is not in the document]
franz: fixed

src/CHIPBackend.hh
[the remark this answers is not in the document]
franz: > but we can use native OpenCL refcount tracking to verify that ours is working correctly

If by that you mean: comparing the numbers coming from `clGetEventInfo(CL_EVENT_REFERENCE_COUNT)` with a separate refcount kept in `CHIPEvent` - that does not work in general. See e.g. [official OpenCL documentation](https://github.com/KhronosGroup/OpenCL-Docs/issues/620) on object lifetime, in particular:

> Implementations may also modify the reference count, e.g. to track attached objects or to ensure correct operation of in-progress or scheduled activities

PoCL does this in multiple places, e.g. with every `clEnqueueNDRangeKernel` we retain the `cl_kernel` plus every `cl_mem` object used to set kernel's arguments. This is necessary because the user is permitted to call `clReleaseKernel` right after enqueuing the kernel, and then use `clFinish` on the queue. I'd be surprised if other implementations don't do this.

src/backend/Level0/CHIPBackendLevel0.cc
```diff
@@ -547,6 +547,37 @@ void CHIPEventLevel0::hostSignal() {
...
+  assert(Refc_ > 0 && "Increasing refcount from zero!");
+  Refc_++;
+}
+
+size_t CHIPEventLevel0::getCHIPRefc() {
+  LOCK(this->EventMtx);
+  assert(!Deleted_ && "Event use after delete!");
+  return Refc_;
+}
+
```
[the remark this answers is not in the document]
franz: Ok, lets continue the chat in a single place

src/backend/OpenCL/CHIPBackendOpenCL.cc
```diff
@@ -234,12 +234,12 @@ annotateSvmPointers(const CHIPContextOpenCL &Ctx, cl_kernel KernelAPIHandle) {
   std::vector<void *> SvmAnnotationList;
   std::unique_ptr<std::vector<std::shared_ptr<void>>> SvmKeepAlives;
   LOCK(Ctx.ContextMtx); // CHIPContextOpenCL::SvmMemory
-  auto NumSvmAllocations = Ctx.SvmMemory.getNumAllocations();
+  auto NumSvmAllocations = Ctx.getRegion().getNumAllocations();
```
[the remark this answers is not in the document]
franz: fixed

src/backend/OpenCL/CHIPBackendOpenCL.hh
franz: This is meant to mimic `std::shared_ptr<T>::reset()` - it replaces the object (ClEvent) stored currently, and releases the old one. `init()` is usually for stuff that needs to be done once, but for some reason can't go into the constructor, so that's quite different.

src/backend/OpenCL/CHIPBackendOpenCL.cc
```diff
@@ -1014,26 +1065,25 @@ void CHIPQueueOpenCL::addCallback(hipStreamCallback_t Callback,
...
 
 CHIPEvent *CHIPQueueOpenCL::enqueueMarkerImpl() {
   CHIPEventOpenCL *MarkerEvent =
       (CHIPEventOpenCL *)Backend->createCHIPEvent(ChipContext_);
-  auto Status =
-      clEnqueueMarker(this->get()->get(), MarkerEvent->getNativePtr());
+  cl::Event RetEv;
+  auto Status = ClQueue.enqueueMarkerWithWaitList(nullptr, &RetEv);
   CHIPERR_CHECK_LOG_AND_THROW(Status, CL_SUCCESS, hipErrorTbd);
+  MarkerEvent->reset(std::move(RetEv));
```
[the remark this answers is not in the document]
franz: commented below

src/backend/OpenCL/CHIPBackendOpenCL.cc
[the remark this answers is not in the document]
franz: as Henry says, `cl::Event` acts similarly to `std::shared_ptr` - it wraps the cl_event and automatically calls release/retain as required. The only thing we need to ensure, is that we have an instance of cl::Event object alive as long as we need it.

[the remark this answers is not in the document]
pvelesko: Right, which is useful for checking CHIP-SPV reference count for consistency. I don't see how losing this feature is useful.

src/backend/OpenCL/CHIPBackendOpenCL.hh
franz: fixed

src/backend/OpenCL/SVMemoryRegion.cc
franz: Opened issue #446

src/CHIPBackend.hh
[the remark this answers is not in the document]
pvelesko: Ah I see, the OpenCL refcount can change in unpredictable ways depending on the OpenCL implementation. You can mark all comments about losing OpenCL refcounting as resolved, but I think that `increaseRefCount()` and related functionality should still be kept at the CHIP-SPV level if not for any other reason than to keep backend implementations as thin as possible.

src/CHIPBackend.cc
franz: continue @ top

src/CHIPGraph.cc
franz: fixed

cmake/UnitTests.cmake
franz: yeah about 50% of graph tests still fail, the work is not finished

src/CHIPBackend.hh
```diff
@@ -1295,13 +1296,6 @@ public:
   CHIPQueue *createQueueAndRegister(const uintptr_t *NativeHandles,
                                     const size_t NumHandles);
 
-  void removeContext(CHIPContext *Ctx);
```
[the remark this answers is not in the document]
franz: ok, i can return it and add a comment

src/CHIPBackend.hh
```diff
@@ -2060,6 +2054,12 @@ public:
 
   CHIPDevice *PerThreadQueueForDevice = nullptr;
 
+  virtual CHIPEvent *enqueueNativeGraph(CHIPGraphNative *NativeGraph) {
+    return nullptr;
+  }
+  virtual CHIPGraphNative *createNativeGraph() { return nullptr; }
```
linehill, line 2060: Do these methods have preconditions? Is it an error to call these if the backends do not override them?

src/CHIPGraph.cc
linehill: I see it makes the variable not exported which is probably the intent. That said, I think the variable should be atomic to avoid race condition in CHIPGraph::addNode().

pvelesko:
@franz conflicts

PR #436, opened 2023-05-10
Implement `__device__ __assert_fail()`

### LINKED ISSUE #387: [Kokkos] Missing __asert_fail
```
// The two keywords below are not contradictory. `noinline` is a
// directive to the optimizer.
[[noreturn]] __device__ __attribute__((noinline)) inline void hip_abort(
    char const *msg) {
  const char empty[] = "";
   __assert_fail(msg, empty, 0, empty);
  // This loop is never executed. It's intended to suppress warnings that the
  // function returns, even though it does not. This is necessary because
  // abort() is not marked as [[noreturn]], even though it does not return.
  while (true)
    ;
}
```

linehill:
... which is needed for assert() being called from device code. Also, some HIP applications calls it directly.

Along the way, fixed HipAbort pass crashing on indirect calls.

Resolves #387.

### DIFF
--- a/tests/runtime/inputs/AssertFail.hip
+++ b/tests/runtime/inputs/AssertFail.hip
@@ -1,0 +1,10 @@
+// Regression test for https://github.com/CHIP-SPV/hipstar/issues/387.
+#include <hip/hip_runtime.h>
+
+__global__ void k() { __assert_fail("expression", "file", 123, "function"); }
+
+int main() {
+  k<<<1, 1>>>();
+  (void)hipDeviceSynchronize();
+  return 0;
+}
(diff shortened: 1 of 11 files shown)


tests/runtime/inputs/AssertFail.hip
```diff
@@ -0,0 +1,10 @@
+// Regression test for https://github.com/CHIP-SPV/hipstar/issues/387.
+#include <hip/hip_runtime.h>
+
+__global__ void k() { __assert_fail("expression", "file", 123, "function"); }
```
pjaaskel, line 4: Can you add a test for the regular assert() macro that delegates to this then?

pjaaskel (changes requested):

tests/runtime/inputs/AssertFail.hip
linehill: Done.

pjaaskel (approved):

linehill:
Put into draft mode: HipAbort pass needs update for addressing regressions.

PR #439, opened 2023-05-15
Find Pekka's clang

pvelesko:


pvelesko:
@pjaaskel let me know if this fixes it for you

cmake/FindLLVM.cmake
```diff
@@ -118,3 +118,4 @@ endif()
 enable_language(C CXX)
 # required by ROCm-Device-Libs, must be after project() call
 find_package(LLVM REQUIRED CONFIG NO_DEFAULT_PATH PATHS ${CLANG_ROOT_PATH}/lib/cmake/llvm)
+find_package(Clang REQUIRED CONFIG NO_DEFAULT_PATH PATHS ${CLANG_ROOT_PATH}/lib/cmake/clang)
```
pjaaskel, line 121: Why NO_DEFAULT_PATH is there? It doesn't find it here if I have it set.

pjaaskel (changes requested):

cmake/FindLLVM.cmake
pvelesko: We want to make sure we use the same Clang that's associated with the `llvm-config` that the user provided. many systems have a pre-installed Clang and we want to make sure we use matching versions.

This again would indicate to me that your Clang was built in some different way if LLVM cmake config and Clang cmake config reside in different directories.

pjaaskel: OK. I'll investigate more.

PR #441, opened 2023-05-16
Fix for CP2K crash observed

Sarbojit2019:
While running CP2K I could see app crash. It is indicating system is out of memory (OOM) hence the crash. Exact error message is "**CHIP error [TID 431805] [1684153005.314770770] : hipErrorTbd (ZE_RESULT_ERROR_OUT_OF_DEVICE_MEMORY Level Zero Event_ pool creation fail! ) in /nfs/site/home/sarbojit/src/chip-spv/src/backend/Level0/CHIPBackendLevel0.cc:312:CHIPEventLevel0**"

I could notice multiple issues in CHIP-SPV event management
- Even though hipEventDestroy() was called reference count was always more than 1 hence runtime can't destroy the event. I could reproduce the issue with simple hipEventCreate and hipEventDestroy sample attached here. [hipEvent_test.cc.txt](https://github.com/CHIP-SPV/chip-spv/files/11486904/hipEvent_test.cc.txt)
- RT was never destroying event pool handle created as part of new event creation. Since CP2K is creating ~60k events system was getting OOM.

With the fix proposed in the PR I could run CP2K multiple times successfully without any crash.


src/backend/Level0/CHIPBackendLevel0.cc
```diff
@@ -246,6 +246,10 @@ CHIPEventLevel0::~CHIPEventLevel0() {
     // '~CHIPEventLevel0' has a non-throwing exception specification
     assert(Status == ZE_RESULT_SUCCESS);
   }
+  if (EventPoolHandle_) {
+    auto Status = zeEventPoolDestroy(EventPoolHandle_);
+    assert(Status == ZE_RESULT_SUCCESS);
+  }
```
pvelesko, line 252: What if there are other events still running that are part of this pool?

Sarbojit2019: I did not see user requested events are kept under same pool. Every time RT creates new event pool see the below code.
image

Sarbojit2019: Either we should not create pool every time while creating new events or we need to delete it while deleting the event.

pvelesko:
Also, please add the test you ran.

pvelesko:
closed in favor of #449

PR #449, opened 2023-05-18
User Event Destruction + EventPool release

### LINKED ISSUE #451: HIPCC unknown escape seq
Thought this is already fixed - maybe in a some outstanding PR

```
In file included from /home/pvelesko/CHIP-SPV/main/HIPCC/src/hipBin.cpp:26:
/home/pvelesko/CHIP-SPV/main/HIPCC/src/hipBin_spirv.h:161:18: warning: unknown escape sequence '\.' [-Wunknown-escape-sequence]
      "(?:\\s|^)[\.a-zA-Z0-9_\\/-]+\\.(?:c)(?:\\s|$)",
                 ^~
```

pvelesko:
While running CP2K I could see app crash. It is indicating system is out of memory (OOM) hence the crash. Exact error message is "CHIP error [TID 431805] [1684153005.314770770] : hipErrorTbd (ZE_RESULT_ERROR_OUT_OF_DEVICE_MEMORY Level Zero Event_ pool creation fail! ) in /nfs/site/home/sarbojit/src/chip-spv/src/backend/Level0/CHIPBackendLevel0.cc:312:CHIPEventLevel0"

I could notice multiple issues in CHIP-SPV event management

Even though hipEventDestroy() was called reference count was always more than 1 hence runtime can't destroy the event. I could reproduce the issue with simple hipEventCreate and hipEventDestroy sample attached here. [hipEvent_test.cc.txt](https://github.com/CHIP-SPV/chip-spv/files/11486904/hipEvent_test.cc.txt)
RT was never destroying event pool handle created as part of new event creation. Since CP2K is creating ~60k events system was getting OOM.
With the fix proposed in the PR I could run CP2K multiple times successfully without any crash.

Fixes #451

### DIFF
--- a/src/CHIPBindings.cc
+++ b/src/CHIPBindings.cc
@@ -2170,9 +2170,7 @@ hipError_t hipEventDestroy(hipEvent_t Event) {
   CHIPEvent *ChipEvent = static_cast<CHIPEvent *>(Event);
 
   ChipEvent->decreaseRefCount("hipEventDestroy");
-  if (ChipEvent->getCHIPRefc() != 0) {
-    logError("hipEventDestroy was called but remaining refcount is not 0");
-  }
+
   RETURN(hipSuccess);
 
   CHIP_CATCH
--- a/src/CHIPBindings.cc
+++ b/src/CHIPBindings.cc
@@ -2168,7 +2168,10 @@ hipError_t hipEventDestroy(hipEvent_t Event) {
   NULLCHECK(Event);
   CHIPEvent *ChipEvent = static_cast<CHIPEvent *>(Event);
 
-  ChipEvent->decreaseRefCount("hipEventDestroy");
+  size_t Refc = ChipEvent->decreaseRefCount("hipEventDestroy");
+  if(Refc == 0) {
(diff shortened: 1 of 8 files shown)


src/CHIPBindings.cc
Sarbojit2019, line 1: line no. 2140  "ChipEvent->increaseRefCount("hipEventCreateWithFlags");" has to be removed else ref count will be two as ref count initialized with 1.

Sarbojit2019 (changes requested):
Ideally there is not need of creating multiple event pools for each create event. RT should have one event pool per device. We can defer it for future optimization.

pvelesko:
Curious case: 

```
 82/883 Test  #82: Unit_deviceFunctions_CompileTest___popcll_int .............................Subprocess aborted***Exception:   0.14 sec
__popcll: /home/pvelesko/actions-runner/_work/chip-spv/chip-spv/src/backend/Level0/CHIPBackendLevel0.cc:240: ze_event_handle_t CHIPEventLevel0::peek(): Assertion `!Deleted_ && "Event use after delete!"' failed.
```

PR #440, opened 2023-05-15
CI Upgrades

pvelesko:
- [ ] Add a seperate step for checking for working devices

PR #460, opened 2023-05-25
SmallFixesLibCeed

pvelesko:
* Re-enable debug printing for refcounts
* launch Event append kernel name
* Implement proper generation of hip_version.h


scripts/unit_tests.sh
```diff
@@ -28,17 +30,30 @@ else
...
-# icpx --version
-# ulimit -a
 sudo /opt/ocl-icd/scripts/igpu_unbind &> /dev/null
 sudo /opt/ocl-icd/scripts/dgpu_unbind &> /dev/null
 
+# Use OpenCL for building/test discovery to prevent Level Zero from being used in multi-thread/multi-process environment
+module load $CLANG
+module load opencl/pocl-cpu-$LLVM
+output=$(clinfo -l 2>&1 | grep "Platform #0")
+if [ $? -ne 0 ]; then
```
linehill, line 46: FYI, the `$?` captures the exit code of grep command.

scripts/unit_tests.sh
```diff
@@ -28,17 +30,30 @@ else
...
+    echo "clinfo failed to execute."
+    exit 1
+fi
+
+# check if the output is empty
+if [ -z "$output" ]; then
```
linehill, line 52: This is not needed if the grep command succeeds I believe.

scripts/unit_tests.sh
pvelesko: As opposed to the result? I want to fail if "Platform #0" is not found in the output of clinfo

linehill: Ok, perhaps make the error message more descriptive. E.g. "clinfo failed to execute." -> "Error: no OpenCL platform available."

pvelesko:
I'll take the comments into consideration in #440

Issue #455, opened 2023-05-23
[Intel OpenCL CPU] Unit_deviceFunctions_CompileTest_j0f_float fails with global is external, but doesn't have external or weak linkage!


pjaaskel:
Unit_deviceFunctions_CompileTest_j0f_float fails with `Global is external, but doesn't have external or weak linkage!

This might be yet another SPIRV-Translator builtin that is generated wrongly without the external attribute.

Also these tests seem to fail similarly:
```
	129 - Unit_deviceFunctions_CompileTest_j0f_float (Subprocess aborted)
	130 - Unit_deviceFunctions_CompileTest_j1f_float (Subprocess aborted)
	131 - Unit_deviceFunctions_CompileTest_jnf_float (Subprocess aborted)
	178 - Unit_deviceFunctions_CompileTest_y0f_float (Subprocess aborted)
	179 - Unit_deviceFunctions_CompileTest_y1f_float (Subprocess aborted)
	180 - Unit_deviceFunctions_CompileTest_ynf_float (Subprocess aborted)
```

pvelesko:
These are passing for me. Are you using 2023.1 oneapi?

pjaaskel:
I'm using the latest drivers from the compute-runtime repo.

pvelesko:
Then I'm not sure. I remember bringing these issues to you and then 2023.1 + SPIR-V Patches seems to have resolved these issues

pjaaskel:
Seems to be due to illegal debug info generated to SPIR-V:

```
spirv-val j0f-hip-spirv64-generic.out
error: line 543: OpenCL.DebugInfo.100 DebugSource: expected operand Text must be a result id of OpString
  %578 = OpExtInst %void %2 DebugSource %576 %577
...
        %577 = OpExtInst %void %2 DebugInfoNone
        %578 = OpExtInst %void %2 DebugSource %576 %577
...

```
That is, it presents the missing optional parameter with DebugInfoNone although it should be omitted completely (since it's an optional one) or perhaps described as OpString None or such.

pjaaskel:
...or not. Seems valid usage:

```
DebugInfoNone

Other instructions can refer to this one in case the debugging information is unknown, not available, or not applicable
```
https://registry.khronos.org/SPIR-V/specs/unified1/OpenCL.DebugInfo.100.html#DebugInfoNone

pjaaskel:
[...]

PR #464, opened 2023-05-29
[Intel CPU] These fail at least with LLVM 16 and Debug build of CHIP-SPV

pjaaskel:
Might need upstreaming fixes to SPIR-V translator so the Intel CPU driver receives them. To investigate.

See https://github.com/CHIP-SPV/chip-spv/issues/455


pvelesko (changes requested):
These are not failing on the CI machine - please reproduce there

pvelesko:
Passes on JLSE as well

pvelesko:
Sunspot too

pvelesko:
Cancelling the last unit test check since all tests pass without adding these to the exclusion list + I need to use the CI machine

pjaaskel:
#466

PR #466, opened 2023-05-29
Fix dropping external linkage attributes ...

### LINKED ISSUE #455: [Intel OpenCL CPU] Unit_deviceFunctions_CompileTest_j0f_float fails with global is external, but doesn't have external or weak linkage!

linehill:
... in SPIR-V binary too eagerly makes the reverse SPIR-V translation crash which is known to affect PoCL.

May fix #455 (couldn't reproduce the issue for Intel OpenCL CPU).

### DIFF
--- a/src/spirv.cc
+++ b/src/spirv.cc
@@ -743,18 +743,16 @@ bool filterSPIRV(const char *Bytes, size_t NumBytes, std::string &Dst) {
       continue;
     if (Insn.isDecoration(spv::DecorationLinkageAttributes)) {
       auto LinkName = parseLinkageAttributeName(Insn);
-      if (parseLinkageAttributeType(Insn) == spv::LinkageTypeImport) {
+      if (EntryPoints.count(LinkName))
+        continue;
+      if (parseLinkageAttributeType(Insn) == spv::LinkageTypeImport)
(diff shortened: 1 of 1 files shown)


pjaaskel:
Confirmed that it fixes the tests both for PoCL-CPU and Intel-CPU here. No clue why the latter doesn't fail for you even though we have the same driver version.

PR #474, opened 2023-06-05
Uninitialize runtime despite unloaded user modules

### LINKED ISSUE #373: [libCEED] Error in uninit
```
// Uninitializes the backend when the application exits.
void __attribute__((destructor)) uninitializeBackend() {
  // Generally, __hipUnregisterFatBinary would uninitialize the
  // backend when it finishes unregistration of all modules. However,
  // there won't be hip(Un)registerFatBinary() calls if the HIP
  // program does not have embedded kernels. This makes sure we
  // uninitialize the backend at exit.
  if (Backend && getSPVRegister().getNumSources() == 0) {
    CHIPUninitialize();
    delete Backend;
    Backend = nullptr;
  }
}
```

When this function is called, `getNumSources()` returns 1 which prevents the backend from properly uninitializing resulting in segfaults.

linehill:
A warning is given at program exit if there are unloaded user modules.

Fixes #373.


pvelesko:
Some failures from event use after delete - investigating that separately.

PR #482, opened 2023-06-08
Work around llvm-spirv crash on switch instruction

linehill:
... with "non-standard" integer bitwidth used as switch value (e.g. i4).

Related issue: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/1685.


pjaaskel:
Thanks for this. In the future we might want to do more smaller pre-processing fixups in a single pass, but this looks good now.

pjaaskel (approved):

pjaaskel:
I originally noted this with some rocPRIM case, IIRC. But doesn't this fix some tests in the CHIP-SPV unit test suites which we could now include?

pvelesko:
> But doesn't this fix some tests in the CHIP-SPV unit test suites which we could now include?

we can just run the resolved failures script prior to 1.0 release

pjaaskel:
> we can just run the resolved failures script prior to 1.0 release

It would be actually nice to do that soon so we can organize the bug fixing efforts better. I mean it would be nice to identify issues that fix most cases and focus on those first (and open tickets to them to avoid overlapping works).

pvelesko:
Weekend after this one if that's alright with you.

PR #489, opened 2023-06-12
Fixes for Unit_hipFree{Negative,Double}{Dev,Host,Array}

linehill:


src/CHIPBackend.cc
```diff
@@ -1150,7 +1150,9 @@ hipError_t CHIPContext::free(void *Ptr) {
   CHIPDevice *ChipDev = Backend->getActiveDevice();
   AllocationInfo *AllocInfo = ChipDev->AllocationTracker->getAllocInfo(Ptr);
   if (!AllocInfo)
-    return hipErrorInvalidDevicePointer;
+    // HIP API doc says we should return hipErrorInvalidDevicePointer but HIP
+    // test suite excepts hipErrorInvalidValue. Go with the latter.
+    return hipErrorInvalidValue;
```
pvelesko, line 1155: We should follow the API and fix the test. Many issues have been found in the tests before.

pvelesko (changes requested):
Let's fix the test instead of returning the wrong error

src/CHIPBackend.cc
linehill: Acknowledged, chose to follow hip-tests expectation until it reflects the API.

PR #490, opened 2023-06-12
Fixes for Unit_hipHost{Register,Unregister}_*

linehill:


pjaaskel (approved):

src/CHIPBindings.cc
```diff
@@ -2439,25 +2438,31 @@ hipError_t hipHostRegister(void *HostPtr, size_t SizeBytes,
...
+                             hipErrorInvalidValue);
+  }
 
   void *DevPtr;
-  auto Err = hipMalloc(&DevPtr, SizeBytes);
-  ERROR_IF(Err != hipSuccess, Err);
+  if (hipMalloc(&DevPtr, SizeBytes) != hipSuccess)
+    // Translate hipOutOfMemory to hipErrorInvalidValue. The latter is
+    // the one hip-tests suite expects in case of OoM.
+    RETURN(hipErrorInvalidValue);
```
pvelesko, line 2465: What does the spec say about this? We shouldn't rely on tests as ground-truth

src/CHIPBindings.cc
```diff
@@ -2472,10 +2477,14 @@ hipError_t hipHostRegister(void *HostPtr, size_t SizeBytes,
...
   CHIP_TRY
   CHIPInitialize();
-  NULLCHECK(HostPtr);
+  if (!HostPtr)
+    CHIPERR_LOG_AND_THROW("Host pointer is nullptr!", hipErrorInvalidValue);
 
-  auto Device = Backend->getActiveDevice();
-  auto AllocInfo = Device->AllocationTracker->getAllocInfo(HostPtr);
+  auto *Device = Backend->getActiveDevice();
+  auto *AllocInfo = Device->AllocationTracker->getAllocInfo(HostPtr);
```
pvelesko, line 2484: What's the point of `auto *`? Doesn't auto automatically deduce the full type no matter if it's a pointer or not?

linehill: For avoiding accidental copies ([link](https://llvm.org/docs/CodingStandards.html#beware-unnecessary-copies-with-auto)).

src/CHIPBindings.cc
linehill: Acknowledged, chose to follow hip-tests expectation until it reflects the API.

PR #557, opened 2023-07-18
fix one test + bump version

franz:


pvelesko:
Why not default to debug?

1 test failed which is known to be flaky - have the fix for it in rc1fixes but struggling with PoCL right now.

pjaaskel:
> Why not default to debug?

I'd default to Debug in the development branch and Release in release branches /  tar balls to try to match the guess of what the default user wants in these cases.

PR #560, opened 2023-07-24
hipGraph API fixes & "native" Graph implementation

franz:
This is the same as PR #434, except:

* removed the code not related to Graphs (i'll make a separate PR for USM)
* merged the style fixes
* updated the code to use chipstar namespace


src/CHIPGraph.cc
```diff
@@ -184,29 +206,71 @@ void CHIPGraph::removeNode(CHIPGraphNode *Node) {
...
-      logDebug("Executing {}", Node->Msg);
-      Node->execute(Queue);
-      Queue->finish();
+  } else {
+    logDebug("NativeGraph: trying to construct");
+    NativeGraph.reset(Queue->createNativeGraph());
+    if (!NativeGraph)
+      return false;
+
+    for (auto &Node : OriginalGraph_->getNodes())
```
pjaaskel, line 224: Perhaps a comment here also to explain what's being done.

src/CHIPGraph.hh
```diff
@@ -645,20 +700,34 @@ public:
   }
 };
 
-class CHIPGraphExec : public hipGraphExec {
+class GraphNative {
```
pjaaskel, line 703: This could have a class comment which states what is meant with Native here. A brief discussion on how the native graph is built with fallback to host-side execution could be elsewhere, maybe in the file comment?

src/backend/OpenCL/CHIPBackendOpenCL.cc
```diff
@@ -1291,6 +1366,401 @@ std::shared_ptr<chipstar::Event> CHIPQueueOpenCL::enqueueBarrierImpl(
...
+chipstar::GraphNative *CHIPQueueOpenCL::createNativeGraph() {
+  // should not raise an error if we fail to create a graph,
+  // because there is a fallback solution
+  CHIPContextOpenCL *Ctx = (CHIPContextOpenCL *)ChipContext_;
+  if (!Ctx->supportsCommandBuffers())
+    return nullptr;
+
+  cl_command_queue CQ = ClQueue_->get();
+  int err = CL_SUCCESS;
+  cl_command_buffer_khr Res =
```
pjaaskel, line 1380: Add a check somewhere that we are targeting only one device, if using the single-device CB.

src/backend/OpenCL/CHIPBackendOpenCL.cc
```diff
@@ -1291,6 +1366,401 @@ std::shared_ptr<chipstar::Event> CHIPQueueOpenCL::enqueueBarrierImpl(
...
+                             SyncPointDeps, &NewSyncPoint);
+    break;
+  case hipGraphNodeTypeHost:
+    Res = addHostNode((chipstar::GraphNodeHost *)NewNode, SyncPointDeps,
+                      &NewSyncPoint);
+    break;
+#endif
+
+  default:
+    Res = false;
```
pjaaskel, line 1479: Perhaps log printout the failing command?

src/backend/OpenCL/CHIPBackendOpenCL.cc
```diff
@@ -1291,6 +1366,401 @@ std::shared_ptr<chipstar::Event> CHIPQueueOpenCL::enqueueBarrierImpl(
   return Event;
 }
 
+/********************************************************************************/
+
+chipstar::GraphNative *CHIPQueueOpenCL::createNativeGraph() {
```
pjaaskel, line 1371: Why GraphNative, why not NativeGraph? Or BackendGraph?

src/backend/OpenCL/CHIPBackendOpenCL.cc
```diff
@@ -1291,6 +1366,401 @@ std::shared_ptr<chipstar::Event> CHIPQueueOpenCL::enqueueBarrierImpl(
...
+}
+
+void CHIPQueueOpenCL::destroyNativeGraph(chipstar::GraphNative *NativeGraph) {
+  if (NativeGraph == nullptr)
+    return;
+  CHIPGraphNativeOpenCL *G = (CHIPGraphNativeOpenCL *)NativeGraph;
+  delete G;
+}
+
+bool CHIPGraphNativeOpenCL::addNode(chipstar::GraphNode *NewNode) {
```
pjaaskel, line 1418: Should these be OpenCLQueue and OpenCLNativeGraph? With CHIP dropped and changed to the chipstar namespace?

pjaaskel (changes requested):
Great work. Just a few comments. Will test the branch on my platforms.

pjaaskel:
One test fails here with iGPU (both LZ and CL) and Intel CPU:
```sh
hipSimpleGraphWithKernel: /home/pjaaskel/src/chip-spv/src/SPIRVFuncInfo.cc:183: void SPVFuncInfo::visitKernelArgsImpl(const std::vector<void *> &, KernelArgVisitor) const: Assertion `ArgData && "nullptr in the argument list"' failed.
Filters: Unit_hipGraph_SimpleGraphWithKernel
```

PoCL-CPU (latest main) fails a lot of tests, but I suppose it's expected before the SVM/CB updates land to main, right?
```
70% tests passed, 419 tests failed out of 1384
```

src/backend/OpenCL/CHIPBackendOpenCL.cc
franz: Maybe if it's decided one day to rename all backend classes, but currently they are all called CHIPxyzOpenCL and have no namespace, so this is consistent with current status.

cmake/UnitTests.cmake
```diff
@@ -167,64 +226,6 @@ list(APPEND CPU_OPENCL_FAILED_TESTS "unroll") # Failed
 list(APPEND CPU_OPENCL_FAILED_TESTS "hipConstantTestDeviceSymbol") # Subprocess aborted
 list(APPEND CPU_OPENCL_FAILED_TESTS "cuda-simpleCallback") # SEGFAULT
 list(APPEND CPU_OPENCL_FAILED_TESTS "cuda-qrng") # Subprocess aborted
-list(APPEND CPU_OPENCL_FAILED_TESTS "Unit_hipGraphAddEmptyNode_NegTest") # SEGFAULT
```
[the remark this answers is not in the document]
franz: None of the tests that i removed from failing list should require anything additional in PoCL (or any other OpenCL or in Level0 runtimes).

franz:
> One test fails here with iGPU (both LZ and CL) and Intel CPU:

It works for me on my iGPU (with both LZ and CL), so i'm going to need more information to debug (llvm/translator exact commits, cmake build flags etc).

> PoCL-CPU (latest main) fails a lot of tests, but I suppose it's expected before the SVM/CB updates land to main, right?

No, this PR is not dependent on PoCL updates. The native graphs are entirely optional, and the PR also passes tests with Level0 backend (which does not yet have the "native" graph implemented).

> 70% tests passed, 419 tests failed out of 1384

Is the 1384 from running ctest directly with no arguments ? The official runner is `scripts/check.py`; with `igpu` option it should run & pass 984/986 (LZ/CL) tests with this PR, and with `cpu pocl` there should be 957 tests. 1384 - 419 = 965 so that is actually a bit more than should be passing with PoCL.

src/CHIPBindings.cc
```diff
@@ -280,15 +280,19 @@ static void handleAbortRequest(chipstar::Queue &Q, chipstar::Module &M) {
...
-  CHIPGraph *Graph = new CHIPGraph();
+  chipstar::Graph *Graph = new chipstar::Graph();
   *pGraph = Graph;
   RETURN(hipSuccess);
   CHIP_CATCH
 }
 
 hipError_t hipGraphDestroy(hipGraph_t graph) {
   CHIP_TRY
+  if (!graph)
```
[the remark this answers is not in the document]
franz: This follows the original AMD's HIP API naming & style, see e.g. [here](https://rocm.docs.amd.com/projects/HIP/en/latest/.doxygen/docBin/html/group___graph.html). Not sure if it's a good idea to change it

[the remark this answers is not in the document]
pjaaskel: We already did when we decided to use the LLVM C++ style in chipStar codes.

src/backend/OpenCL/CHIPBackendOpenCL.cc
pjaaskel: I think we decided to indeed do that. If it's only partially done, we should convert the nonconverted ones, not add new wrongly named ones to away from that decision, right?

pjaaskel:
Please also add some text to ReleaseNotes of this feature.

pvelesko:
@franz conflicts

PR #570, opened 2023-07-31
Address performance issue in #550

### LINKED ISSUE #550: volatile array in device-side memset (performance issue)
With @jjennychen, we noticed that the device-side memset for CHIP-SPV was much slower than writing our own loop to do a memset. With @pengtu we narrowed this down to the fact that the array in memset (and memcpy) is declared as volatile (https://github.com/CHIP-SPV/chipStar/blob/96d21716110212530f663342c9ab214b9b866061/bitcode/devicelib.cl#L446).

Is there a reason this is volatile? I can't think of why we would need that for the memset, but I could be missing something. When we tested without the volatile, we saw much better performance.

Thanks!

linehill:
... by removing volatile keywords in __chip_mem{set,cpy}.

Closes #550.

### DIFF
--- a/llvm_passes/HipLowerMemset.cpp
+++ b/llvm_passes/HipLowerMemset.cpp
@@ -1,0 +1,74 @@
+//===- HipLowerMemset.cpp -------------------------------------------------===//
+//
+// Part of the chipStar Project, under the Apache License v2.0 with LLVM
+// Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+// Expand llvm.memset intrinsics to loops before handing the bitcode to
+// llvm-spirv translator. The translator expands memset intrinsics to an
+// emulation function with a name encoding the destination pointer type
+// (e.g. "spirv.spirv.llvm_memset_p0_i32..."). During reverse translation,
+// llvm-spirv attempts to recreate the original memset intrinsic which can be
+// incorrect if we are jumping from opaque pointer world to typed pointer world
+// or vice versa. LLVM module verification may fail with the following message:
+//
+//   Fails to verify module: Intrinsic name not mangled correctly for
+//   type arguments!
+//
+// This issue has been filed here:
+// https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/2128
+//
+// llvm.memcpy does not have this issue as it gets translated to actual SPIR-V
+// instruction (OpCopyMemorySized).
+//
+// (c) 2023 Henry Linjamäki / Intel
+//===----------------------------------------------------------------------===//
+#include "HipLowerMemset.h"
+
+#include <llvm/ADT/SmallPtrSet.h>
+#include <llvm/IR/Function.h>
+#include <llvm/IR/Instructions.h>
+#include <llvm/IR/IRBuilder.h>
+#include <llvm/Passes/PassBuilder.h>
+#include <llvm/Passes/PassPlugin.h>
+#include <llvm/Transforms/Utils/LowerMemIntrinsics.h>
+
+using namespace llvm;
+
+static bool lowerMemsets(Function &F) {
+  SmallPtrSet<MemSetInst *, 8> WorkList;
+  for (auto &BB : F)
+    for (auto &I : BB)
+      if (auto *MemSet = dyn_cast<MemSetInst>(&I))
+        WorkList.insert(MemSet);
+
+  for (auto *MemSet : WorkList) {
+    expandMemSetAsLoop(MemSet);
+    MemSet->eraseFromParent();
+  }
+
+  return !WorkList.empty();
+}
+
+PreservedAnalyses HipLowerMemsetPass::run(Function &F,
+                                          FunctionAnalysisManager &AM) {
+  return lowerMemsets(F) ? PreservedAnalyses::none() : PreservedAnalyses::all();
+}
+
+extern "C" ::llvm::PassPluginLibraryInfo LLVM_ATTRIBUTE_WEAK
+llvmGetPassPluginInfo() {
+  return {LLVM_PLUGIN_API_VERSION, "hip-lower-memset", LLVM_VERSION_STRING,
+          [](PassBuilder &PB) {
+            PB.registerPipelineParsingCallback(
+                [](StringRef Name, FunctionPassManager &FPM,
+                   ArrayRef<PassBuilder::PipelineElement>) {
+                  if (Name == "hip-lower-memset") {
+                    FPM.addPass(HipLowerMemsetPass());
+                    return true;
+                  }
+                  return false;
+                });
+          }};
+}
--- a/bitcode/devicelib.cl
+++ b/bitcode/devicelib.cl
@@ -458,7 +458,7 @@ EXPORT void __chip_threadfence_system() {
 // sets size bytes of the memory pointed to by ptr to value
 // interpret ptr as a unsigned char so that it writes as bytes
 EXPORT void *__chip_memset(DEFAULT_AS void *ptr, int value, size_t size) {
-  volatile unsigned char *temporary = ptr;
+  unsigned char *temporary = ptr;
 
   for (int i = 0; i < size; i++)
     temporary[i] = value;
@@ -468,8 +468,8 @@ EXPORT void *__chip_memset(DEFAULT_AS void *ptr, int value, size_t size) {
 
 EXPORT void *__chip_memcpy(DEFAULT_AS void *dest, DEFAULT_AS const void *src,
                            size_t n) {
-  volatile unsigned char *temporary_dest = dest;
-  volatile const unsigned char *temporary_src = src;
+  unsigned char *temporary_dest = dest;
+  const unsigned char *temporary_src = src;
 
   for (int i = 0; i < n; i++)
     temporary_dest[i] = temporary_src[i];
--- a/llvm_passes/CMakeLists.txt
+++ b/llvm_passes/CMakeLists.txt
@@ -99,7 +99,7 @@ add_library(LLVMHipPasses MODULE HipPasses.cpp
     HipPrintf.cpp HipGlobalVariables.cpp HipTextureLowering.cpp HipAbort.cpp
     HipEmitLoweredNames.cpp HipWarps.cpp HipKernelArgSpiller.cpp
     HipLowerZeroLengthArrays.cpp HipSanityChecks.cpp HipLowerSwitch.cpp
-    ${EXTRA_OBJS})
+    HipLowerMemset.cpp ${EXTRA_OBJS})
 
 if("${LLVM_VERSION}" VERSION_GREATER_EQUAL 14.0)
   set_target_properties(LLVMHipPasses PROPERTIES
--- a/llvm_passes/CMakeLists.txt
+++ b/llvm_passes/CMakeLists.txt
@@ -60,15 +60,20 @@ endif()
 
 set(EXTRA_OBJS)
 
+# Determine if the opt tool is statically linked and import missing
+# function definitions, needed by our pass plugin, because statically
+# linked opt tool only includes symbols which referenced by it. A similar
+# bug has been filed here: https://github.com/llvm/llvm-project/issues/148.
 if(LLVM_LINK_LLVM_DYLIB OR LLVM_ENABLE_SHARED_LIBS)
   message(STATUS "LLVM built with shared libraries -> pass plugin should just work")
 else()
   message(STATUS "LLVM built with static libraries -> adding workaround for issue #102")
 
-  get_target_property(LLVMCORE_CONF LLVMCore IMPORTED_CONFIGURATIONS)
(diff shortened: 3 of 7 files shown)


pjaaskel (approved):

llvm_passes/HipLowerMemset.cpp
```diff
@@ -0,0 +1,71 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+// Expand llvm.memset intrinsics to loops before handing the bitcode to
+// llvm-spirv translator. The translator expands memset intrinsics to emulation
+// function with a name encoding the destination pointer type
+// (e.g. "spirv.spirv.llvm_memset_p0_i32..."). During reverse translation,
+// llvm-spirv attempts to recreate the original memset intrinsic which can be
+// incorrect if we are jumping from opaque pointer world to typed pointer world
+// or vice versa. LLVM module verification may fail with the following message:
```
pjaaskel, line 15: Is this a SPIR-V translator bug? Can you add a link to the issue here?

llvm_passes/HipLowerMemset.cpp
```diff
@@ -0,0 +1,71 @@
...
+//   type arguments!
+//
+// llvm.memcopy does not have this issue as it gets translated to actual SPIR-V
+// instruction (OpCopyMemorySized).
+//
+// (c) 2023 Henry Linjamäki / Intel
+//===----------------------------------------------------------------------===//
+#include "HipLowerMemset.h"
+
+#include "llvm/ADT/SmallPtrSet.h"
```
pjaaskel, line 27: Use "system include" (#include <>) if not referring to headers in our own tree.

llvm_passes/HipLowerMemset.cpp
```diff
@@ -0,0 +1,71 @@
...
+#include "llvm/IR/Function.h"
+#include "llvm/IR/Instructions.h"
+#include "llvm/IR/IRBuilder.h"
+#include "llvm/Passes/PassBuilder.h"
+#include "llvm/Passes/PassPlugin.h"
+#include "llvm/Transforms/Utils/LowerMemIntrinsics.h"
+
+using namespace llvm;
+
+static bool lowerMemsets(Function &F) {
```
pjaaskel, line 37: Are you sure there is not such a utility function in upstream LLVM? Would make sense.

pjaaskel (approved):
If there is no SPIR-V translator issue, pls open one. Otherwise LGTM.

llvm_passes/HipLowerMemset.cpp
linehill: I'll file the issue later.

llvm_passes/HipLowerMemset.cpp
linehill: expandMemSetAsLoop() is an utility function provided by the upstream LLVM.

linehill:
> If there is no SPIR-V translator issue, pls open one.

Opened one and added a link to it in the HipLowerMemset.cpp.

pvelesko:
@pjaaskel 
```
1299: Test command: /home/pvelesko/chipStar/addLibCeed/build/tests/runtime/TestMemFunctions
1299: Working Directory: /home/pvelesko/chipStar/addLibCeed/build/tests/runtime
1299: Test timeout computed to be: 10000000
1299: Fails to verify module: Intrinsic name not mangled correctly for type arguments! Should be: llvm.memset.p1i8.i64
1299: void (i8 addrspace(1)*, i8, i64, i1)* @llvm.memset.p1.i64
1299: CHIP error [TID 517151] [1692337746.876520025] : Program BUILD LOG for device #0:cpu-12th Gen Intel(R) Core(TM) i3-12100:
1299: Device cpu-12th Gen Intel(R) Core(TM) i3-12100 failed to build the program
1299: 
1299: 
1299: CHIP error [TID 517151] [1692337746.876553120] : hipErrorNotInitialized (CL_BUILD_PROGRAM_FAILURE ) in /home/pvelesko/chipStar/addLibCeed/src/backend/OpenCL/CHIPBackendOpenCL.cc:729:compile
1299: 
1299: CHIP error [TID 517151] [1692337746.876616195] : Caught Error: hipErrorNotInitialized
1/1 Test #1299: TestMemFunctions .................***Failed    0.55 sec
```

pvelesko:
```
/opt/install/llvm/15.0/bin/opt: symbol lookup error: /opt/actions-runner/_work/chipStar/chipStar/build/lib/libLLVMHipSpvPasses.so: undefined symbol: _ZN4llvm18expandMemSetAsLoopEPNS_10MemSetInstE
```

pvelesko (changes requested):
build errors

pvelesko:
This is specific to this PR

pvelesko:
Fails for both GitHub runner via Brice's CI and on cupcake.

pjaaskel:
```
nm -DC ~/local/lib/libLLVM-16.0.6.so | grep expandMemSetAsLoop
00000000017b9c50 T llvm::expandMemSetAsLoop(llvm::MemSetInst*)@@LLVM_16
```
I have that symbol. What about your LLVM so, @pvelesko ?

linehill:
Also, what does `ldd /opt/install/llvm/15.0/bin/opt` say? Does the `opt` link to the LLVM it is supposed to?

pvelesko:
> Also, what does `ldd /opt/install/llvm/15.0/bin/opt` say? Does the `opt` link to the LLVM it is supposed to?

```
╭─pvelesko@cupcake ~
╰─$ ldd /opt/install/llvm/15.0/bin/opt                                                                                                                                                                                                         130 ↵
	linux-vdso.so.1 (0x00007ffe8ade2000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f52a14b9000)
	libzstd.so.1 => /lib/x86_64-linux-gnu/libzstd.so.1 (0x00007f52a13ea000)
	libtinfo.so.6 => /lib/x86_64-linux-gnu/libtinfo.so.6 (0x00007f52a13b8000)
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f52a118c000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f52a10a5000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f52a1083000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f52a0e5b000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f52a470d000)
```

pvelesko:
I don't have this library in the install dir? I'm using `scripts/configure_llvm.sh` to build and install llvm into `/opt/install/llvm/16.0` 

I have a couple of these older .so in my system paths but these are old and not used.
```
╭─pvelesko@cupcake /opt/install/llvm
╰─$ find / -name "libLLVM-*" 2> /dev/null
/usr/lib/llvm-15/lib/python3/dist-packages/lldb/libLLVM-15.so.1
/usr/lib/llvm-15/lib/python3/dist-packages/lldb/libLLVM-15.0.7.so.1
/usr/lib/llvm-15/lib/libLLVM-15.so.1
/usr/lib/llvm-15/lib/libLLVM-15.0.7.so.1
/usr/lib/llvm-15/lib/libLLVM-15.so
/usr/lib/llvm-15/lib/libLLVM-15.0.7.so
/usr/lib/llvm-14/lib/libLLVM-14.so.1
/usr/lib/llvm-14/lib/libLLVM-14.0.0.so
/usr/lib/llvm-14/lib/libLLVM-14.so
/usr/lib/llvm-14/lib/libLLVM-14.0.0.so.1
/usr/lib/x86_64-linux-gnu/libLLVM-15.so.1
/usr/lib/x86_64-linux-gnu/libLLVM-15.0.7.so.1
/usr/lib/x86_64-linux-gnu/libLLVM-15.so
/usr/lib/x86_64-linux-gnu/libLLVM-14.so.1
/usr/lib/x86_64-linux-gnu/libLLVM-13.so.1
/usr/lib/x86_64-linux-gnu/libLLVM-14.so
/usr/lib/x86_64-linux-gnu/libLLVM-13.so
/usr/lib/x86_64-linux-gnu/libLLVM-14.0.0.so.1
```

pjaaskel:
I think this is a problem with the static linkage, which is an annoying problem with LLVM and the dynamically loaded passes: opt hasn't needed that symbol, thus hasn't linked it in and since it doesn't have a dynlib to load it from, it fails. If we link in that symbol statically to our pass then we might get multiple defs of that symbol if loading to a dynamically linked opt which also gets that symbol from the LLVM dynlib.

linehill:
It could also be a fault in libLLVMHipSpvPasses.so as we currently don't link it to LLVM (dynamically or statically):

```
/mnt/md1/linehill/ws-chipstar-2/install/bin/clang++ -fPIC  -Wno-duplicate-decl-specifier -Wno-tautological-constant-compare  -Wno-c++20-extensions -Wno-unused-result -Wno-delete-abstract-non-virtual-dtor -Wno-deprecated-declarations -Wunused-command-line-argument -gdwarf-4 -g -shared  -o ../lib/libLLVMHipSpvPasses.so CMakeFiles/LLVMHipPasses.dir/HipPasses.cpp.o CMakeFiles/LLVMHipPasses.dir/HipDynMem.cpp.o CMakeFiles/LLVMHipPasses.dir/HipStripUsedIntrinsics.cpp.o CMakeFiles/LLVMHipPasses.dir/HipDefrost.cpp.o CMakeFiles/LLVMHipPasses.dir/HipPrintf.cpp.o CMakeFiles/LLVMHipPasses.dir/HipGlobalVariables.cpp.o CMakeFiles/LLVMHipPasses.dir/HipTextureLowering.cpp.o CMakeFiles/LLVMHipPasses.dir/HipAbort.cpp.o CMakeFiles/LLVMHipPasses.dir/HipEmitLoweredNames.cpp.o CMakeFiles/LLVMHipPasses.dir/HipWarps.cpp.o CMakeFiles/LLVMHipPasses.dir/HipKernelArgSpiller.cpp.o CMakeFiles/LLVMHipPasses.dir/HipLowerZeroLengthArrays.cpp.o CMakeFiles/LLVMHipPasses.dir/HipSanityChecks.cpp.o CMakeFiles/LLVMHipPasses.dir/HipLowerSwitch.cpp.o 
```
^ No linking to LLVM libraries here.

```
$ readelf -d lib/libLLVMHipSpvPasses.so | grep NEEDED
 0x0000000000000001 (NEEDED)             Shared library: [libstdc++.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libm.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libgcc_s.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
```

pjaaskel:
> What is being statically linked right now?

Your 'opt', if I read your ldd output right.

@linehill yep, we need to link, but the static linkage might have the problem I described. It should have a solution though as it should be normal to link in out-of-tree LLVM passes as plugins (I recall struggling with that long time ago).

pvelesko:
PoCL failrues when trying to use dynamic LLVM? @pjaaskel 
Maybe PoCL didn't get rebuilt once I switched LLVM? 
```
/home/runner/opt/llvm/16/bin/opt: symbol lookup error: /home/runner/work/chipStar/chipStar/build/lib/libLLVMHipSpvPasses.so: undefined symbol: _ZN4llvm18expandMemSetAsLoopEPNS_10MemSetInstE
```

pvelesko:
@pjaaskel I forced the rebuild of PoCL and still:
```
[ 64%] Building CXX object samples/abort/CMakeFiles/abort.dir/abort.cc.o
/home/runner/opt/llvm/16/bin/opt: symbol lookup error: /home/runner/work/chipStar/chipStar/build/lib/libLLVMHipSpvPasses.so: undefined symbol: _ZN4llvm18expandMemSetAsLoopEPNS_10MemSetInstE

```

linehill:
The cause of the issue is probably still the same as before. The plugin is loaded by statically built opt tool which does not provide symbols needed by the plugin. Made a patch (#645) for the plugin  so it links to the LLVM dynamically. Try if it fixes the problem (make sure you build the chipStar from scratch).

pvelesko:
Does building llvm as dynamic not link opt dynamic as well?

linehill:
> Does building llvm as dynamic not link opt dynamic as well?

I don't know LLVM build system well enough to answer the question. Like, is there a configuration that let one build shared LLVM library but still have binary tools statically linked? 

The recent lookup error suggests that the opt tool is statically linked but that should not be a matter anymore. I pushed a change that makes HipMemsetLowering work with statically linked opt tool.

franz:
> Like, is there a configuration that let one build shared LLVM library but still have binary tools statically linked?

I think so. Using `-DLLVM_BUILD_LLVM_DYLIB=ON` together with `-DLLVM_LINK_LLVM_DYLIB=OFF`. The manual mentions: `Tools will only be linked to the libLLVM shared library if LLVM_LINK_LLVM_DYLIB is also ON.`

PR #571, opened 2023-08-01
remove the hip-config.cmake requirement for clang

### LINKED ISSUE #340: CMake generated hip-config enforces the use of clang for downstream objects
- [ ] Add a test

```
CMake Error at /home/pvelesko/install/HIP/chip-spv/lib/cmake/hip/hip-config.cmake:143 (message):
  this project must be compiled with clang.  CMAKE_CXX_COMPILER_ID = GNU
Call Stack (most recent call first):
  CMakeLists.txt:126 (find_package)
```

pvelesko:
Fixes #340 

This requirement is being removed because for example, building the shim layer for MKL requires us to incldue hip-config.cmake but also requires us to use Intel compilers thus failing this check.

### DIFF
--- a/cmake/hip-config-in.cmake
+++ b/cmake/hip-config-in.cmake
@@ -103,22 +103,6 @@ else()
   set_and_check(hip_HIPCONFIG_EXECUTABLE "${hip_BIN_INSTALL_DIR}/hipconfig")
 endif()
 
-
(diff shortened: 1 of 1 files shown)


pjaaskel:
Perhaps add in the commit log a bit of explanation why this was removed (for future reference). Otherwise should be fine - I at least cannot come up with a better solution here.

PR #572, opened 2023-08-01
Fix couln't find hipcc, symlink for hipcc.bin

### LINKED ISSUE #554: TestHiprtcCPPKernels and TestHiprtcOptions assume 'hipcc' binary found in PATH
If I don't install chipStar, but just test the build, the test fails. It gets fixed if I ensure 'hipcc' is found in PATH. This indicates that it tests installed chipStar not the build dir one, which is of course a problem. build.py should put 'build' as the first one in the PATH to prevent testing installed chipStar accidentally. It also needs to add symblink hipcc.bin -> hipcc under the bin directory in the build.

Occurs in #553 and likely #552.

linehill:
* Fix hiprtc couldn't find runnable hipcc. hiprtc ignored predefined hipcc candidates in getHIPCCPath() due to too strict file permission condition check. The hipcc binaries in the build and install locations were rejected if they didn't have executable bit set in the 'other' field (o+x).

* Add <build-dir>/bin/hipcc -> <build-dir>/bin/hipcc.bin symlink and replace most of the hipcc.bin references to use the link instead. Enables use of the hipcc from the build directory.

Resolves #554.


src/Utils.cc
```diff
@@ -73,17 +73,14 @@ void dumpSpirv(std::string_view Spirv) {
...
+  if (!fs::exists(HipccPath))
     return false;
 
-  const auto &Perms = fs::status(Path).permissions();
-
-  // Only considering file to be executable if it has others_exec bit
-  // set on. For owner_exec and group_exec bits we need to check
-  // user's file access privileges.
-  return (Perms & fs::perms::others_exec) != fs::perms::none;
+  auto Cmd = std::string("sh -c '\"") + HipccPath.string() +
```
pjaaskel, line 81: Is this called often? If so, perhaps cache the return value in a static variable?

src/Utils.cc
```diff
@@ -135,9 +132,8 @@ std::optional<fs::path> getHIPCCPath() {
   //       installed CHIP library. Mixing the installed and the built
   //       resources could lead to obscure issues.
   for (const auto &ExeCand : {fs::path(CHIP_INSTALL_DIR) / "bin/hipcc",
-                              fs::path(CHIP_BUILD_DIR) / "bin/hipcc",
-                              fs::path(CHIP_BUILD_DIR) / "bin/hipcc.bin"})
-    if (canExecute(ExeCand))
+                              fs::path(CHIP_BUILD_DIR) / "bin/hipcc"})
+    if (canExecuteHipcc(ExeCand))
```
pjaaskel, line 136: ...or just cache this result?

pjaaskel (approved):
If the caching is unnecessary (not called often), looks good to me.

src/Utils.cc
linehill: Every time a HIP compilation occurs. I think the overhead is not significant but the caching is a small and easy thing to add so I added it.

src/Utils.cc
```diff
@@ -128,15 +128,22 @@ std::optional<std::string> readFromFile(const fs::path Path) {
...
-                              fs::path(CHIP_BUILD_DIR) / "bin/hipcc"})
-    if (canExecuteHipcc(ExeCand))
-      return ExeCand;
-
-  return std::nullopt;
+  static std::once_flag Flag;
+  static std::optional<fs::path> HIPCCPath;
+
+  std::call_once(Flag, []() {
+    // TODO: Probably should detect if we are using a built or an
```
pjaaskel, line 135: For now, pls add CHIP_DEVEL_MODE env to control this, like in PoCL. This is indeed annoying.

pjaaskel (changes requested):

src/Utils.cc
linehill: Would it be fine if it were controlled by CMAKE_BUILD_TYPE? Debug -> Pick hipcc from build directory, Release* -> prefer installed hipcc over one in the build directory.

pjaaskel: Makes sense!

pjaaskel (approved):

linehill:
Rebased on main, fixed an issue that prevented the chipStar being built from scratch.

PR #577, opened 2023-08-07
Improve atomics implementations

pjaaskel:
* Call the generic address space ones instead of checking which AS it is and branching
* Implement system atomics. They require CC >= 6.0 to actually function. Added a check to not return CC major larger than 5 if the matching SVM capabilities are not available.


src/backend/OpenCL/CHIPBackendOpenCL.cc
```diff
@@ -1457,7 +1473,7 @@ chipstar::EventMonitor *CHIPBackendOpenCL::createStaleEventMonitor_() {
 }
 
 std::string CHIPBackendOpenCL::getDefaultJitFlags() {
-  return std::string("-x spir -cl-kernel-arg-info");
+  return std::string("-x spir -cl-kernel-arg-info -cl-std=CL3.0");
```
linehill, line 1476: Should update README.md to state the OpenCL 3.0 is the minimum driver version required? Currently, it states OpenCL 2.0 is the minimum required version.

bitcode/devicelib.cl
```diff
@@ -39,36 +39,46 @@
...
+#ifndef __opencl_c_generic_address_space
+#error __opencl_c_generic_address_space needed!
+#endif
+
+EXPORT unsigned /* long */ long int
+__chip_umul64hi(unsigned /* long */ long int x,
+                unsigned /* long */ long int y) {
+  unsigned /* long */ long int mul =
+      (unsigned /* long */ long int)x * (unsigned /* long */ long int)y;
+  return (unsigned /* long */ long int)(mul >> 64);
```
linehill, line 51: It would be better if the changes unrelated to this patch were not included here. It's hard spot issues through the mass style-formatting.

src/backend/OpenCL/CHIPBackendOpenCL.cc
pjaaskel: Actually, I'll try if it works with CL2.0 still.

pjaaskel: No, actually it's best use CL3.0 since it's likely more wide spread (in the future) than CL2.0.

bitcode/devicelib.cl
pjaaskel: Yep, I often do a separate clang-format commit, but in this case it was so small diff, I incldued it here.

pjaaskel:
Rebased. I still cannot reproduce the failing test with PoCL-CPU. I'm using main branch revision pocl/pocl@0041923e0e3

pjaaskel:
(Rebased)

pvelesko:
PoCL tests pass on the new CI for PoCL without disabling kernel cache and all previous tests already passed. Merging - let's keep a lookout for random segfaults in devicelib tests on PoCL

PR #583, opened 2023-08-14
add libCEED to unit testing + exclude textures from checks

pvelesko:


scripts/unit_tests.sh
```diff
@@ -82,6 +87,11 @@ echo "begin igpu_level0_failed_tests"
 module load levelzero/igpu
 module list
 ctest --timeout 180 -j 1 --output-on-failure -E "`cat ./test_lists/igpu_level0_failed_tests.txt`" | tee igpu_level0_make_check_result.txt
+
+pushd ${LIBCEED_DIR}
+make FC= CC=clang CXX=clang++ BACKENDS="/gpu/hip/ref /gpu/hip/shared /gpu/hip/gen" prove -j PROVE_OPS="-j" | tee igpu_level0_make_check_result.txt
```
linehill, line 92: Please, don't use `-j` option without a job count for Makefile. That will spawn inifinite amount of jobs which may, for example, lead to OoM errors.
```
-j [N], --jobs[=N]          Allow N jobs at once; infinite jobs with no arg.
```
Rather, use something like, for example,  `make -j$(nproc) ...` instead.

PR #590, opened 2023-08-21
switch to using system level zero headers

pvelesko:
Switch to using level-zero headers found on the system (along with ze_loader)


linehill:
Could you add a check to ensure the required headers are found (ze_api.h) at configure time. Users might have the ze_loader library installed but not the headers. Otherwise, LGTM.

linehill (approved):

linehill:
I get an configuration error because the ze_api.h header is not in the directory the chipStar expects it to be.
```
-- Setting LevelZeroLib_DIR to /usr/lib/x86_64-linux-gnu
CMake Error at CMakeLists.txt:91 (message):
  ze_api.h was not found in /usr/lib/x86_64-linux-gnu/../include/level_zero
```
On my system the header is located at `/usr/include/level_zero/ze_api.h`.

pvelesko:
@linehill hmm I wonder which other locations should be checked and how to check if this header is associated with the correct loader?

linehill:
I installed the header from [.deb](https://github.com/oneapi-src/level-zero/releases/tag/v1.13.5) packages. They also provides pkg-config file which can tell their include directories:
```
$ pkg-config --cflags-only-I level-zero
-I/usr/include/level_zero
```

pvelesko:
@linehill check if this works for you now.

linehill:
There is still a problem with the configuration:
```
-- Setting LevelZeroInclude_DIR to /usr/include
-- Setting LevelZeroLib_DIR to /usr/lib/x86_64-linux-gnu
CMake Error at CMakeLists.txt:97 (message):
  ze_api.h was not found in /usr/include
```

pvelesko:
oh my bad - I misread.

linehill:
The configuration passes now.

PR #609, opened 2023-08-29
Level Zero - env var CHIP_L0_IMM_CMD_LISTS

pvelesko:
* Introduce a new env var CHIP_L0_IMM_CMD_LISTS
* This will simplify the use of immediate command lists such that the user will not have to recompile the code
* Easily enable testing without having to recompile which takes a while

TODO:

- [x] Merge #598 
- [ ] Merge hipEventRecord blocking test
- [ ] Add unit testing
- [ ] Run unit tests on Cupcake + Sunspot
- [ ] Adjust test exclusions as necessary

### DIFF
--- a/src/backend/Level0/CHIPBackendLevel0.cc
+++ b/src/backend/Level0/CHIPBackendLevel0.cc
@@ -25,8 +25,9 @@
 
 /**
  *  CHIPQueueLevel0::getCmdList() will return an immediate command list handle
- * if L0_IMM_QUEUES is used. There is only one such handle for a queue and a
- * queue can be shared between multiple threads thus this lock is necessary.
+ * if level zero immediate command lists is used. There is only one such handle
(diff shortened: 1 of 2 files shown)


pjaaskel:
Is there a device query in LZ which one could use to ask if ICL is supported or not and then we'd always use it if it's advertised? Or is this entirely for (suspected) buggy drivers?

pvelesko:
All level zero devices support ICL. After installing 2023.2 it seems that ICL issue which used to fail no longer does so - good news! @pjaaskel

pjaaskel:
So, if you believe ICL is always the best choice for LZ, we should just use it by default and perhaps retain the non-ICL as a build option for some time?

pvelesko:
> we should just use it by default

that's the idea. Working on it right now.

PR #611, opened 2023-08-29
CI using pocl-cpu.

Kerilk:
This add a working CI that leverages GitHub actions and pocl to test the OpenCL code path.
- Run is 1h
- Building llvm is a couple of hours, but it is cached
- SPIRV-Tools is apparently unneeded, so we may want to remove it to streamline the workflow.

There are many ways we could improve on it I assume, but at least it demonstrates the feasibility of the approach.

### DIFF
--- a/.github/actions/get-shas/action.yml
+++ b/.github/actions/get-shas/action.yml
@@ -1,0 +1,25 @@
+name: 'Get SHAs'
+description: 'Get SHAs of tips of branches we use.'
+inputs:
+  version:
+    description: 'LLVM version'
+    required: true
+outputs:
+  llvm-sha:
+    description: 'SHA of llvm-projetc to checkout'
+    value: ${{ steps.llvm-sha.outputs.sha }}
+  spirv-llvm-sha:
+    description: 'SHA of SIPRV LLVM Translator to checkout'
+    value: ${{ steps.spirv-llvm-sha.outputs.sha }}
+  cache-key:
+    description: 'Key to use when caching LLVM builds'
+    value: ${{ runner.os }}-build-cache-llvm-${{ inputs.version }}-${{ steps.llvm-sha.outputs.sha }}-${{ steps.spirv-llvm-sha.outputs.sha }}
+runs:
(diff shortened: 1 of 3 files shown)


pjaaskel:
Good! We should also add the chipStar tests to the PoCL CI.

Kerilk:
Here is how to integrate the commit hash to the cache, for llvm builds mainly since they may change in the future, but we could generalize for the others so that several versions could be cached:
https://github.com/Kerilk/OpenCL-ICD-Loader/blob/5eb5a5b9d2ac973a6fd2971b9a1ebcef9e1beb51/.github/workflows/linux.yml#L25-L37
```yaml
    - name: Get sha of Headers
      id: get-sha
      run: |
        echo ::set-output name=sha::$( curl -u "u:${{github.token}}" https://api.github.com/repos/KhronosGroup/OpenCL-Headers/git/ref/heads/main | jq .object.sha | tr -d '"' )


    - uses: actions/cache@v2
      name: Cache Headers
      id: headers
      env:
        cache-name: cache-headers
      with:
        path: external/OpenCL-Headers/install
        key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ steps.get-sha.outputs.sha }}
```

pvelesko:
LGTM - One thing to look out for are some random failures in devicelib tests which occured on cupcake due to what we found out to be the PoCL kernel cache.

pjaaskel:
Could be PoCL kernel cache, or just a timing issue that usually doesn't appear when it has to build the kernel. I suspect the latter as the kcache is old and has been very robust as such. My hunch is more towards the program scope variable support, which is new (and writes something to the cache).

pvelesko:
If I recall correctly all of these segfaults happened inside PoCL when chipStar was executing global variable init ghost funcs

pjaaskel:
Yep, so I suspect it's something timing sensitive with the gvariable init inside PoCL. Your system is the only one I've seen it happen though, rest of us have been luckier so far.

PR #614, opened 2023-09-01
Add CI for Intel CPU OpenCL

Kerilk:
This does a couple of things:
- Rebuild pocl on llvm changes
- Remove `tee` that was preventing tests from being reported as failed
- Add test of the Intel OpenCL CPU runtime

### DIFF
--- a/.github/actions/get-shas/action.yml
+++ b/.github/actions/get-shas/action.yml
@@ -11,15 +11,33 @@ outputs:
   spirv-llvm-sha:
     description: 'SHA of SIPRV LLVM Translator to checkout'
     value: ${{ steps.spirv-llvm-sha.outputs.sha }}
-  cache-key:
+  pocl-sha:
+    description: 'SHA of pocl to checkout'
+    value: ${{ steps.pocl-sha.outputs.sha }}
+  llvm-cache-key:
     description: 'Key to use when caching LLVM builds'
     value: ${{ runner.os }}-build-cache-llvm-${{ inputs.version }}-${{ steps.llvm-sha.outputs.sha }}-${{ steps.spirv-llvm-sha.outputs.sha }}
+  pocl-cache-key:
+    description: 'Key to use when caching pocl builds'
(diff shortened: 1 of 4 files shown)


pvelesko:
disabled intel cpu tests on cupcake

Kerilk:
Tests started failing since I rebased, see here for a clean run:
 - https://github.com/Kerilk/chipStar/actions/runs/6043269819
 and Here for a bad one:
- https://github.com/Kerilk/chipStar/actions/runs/6044595910
But maybe it is unrelated since it passed for you here:
https://github.com/CHIP-SPV/chipStar/actions/runs/6044596233

Kerilk:
Looking at the failures, they appear random:
See here
https://github.com/Kerilk/chipStar/actions/runs/6044595910/job/16405450911
vs there
https://github.com/CHIP-SPV/chipStar/actions/runs/6043827942/job/16401611205

but always on pocl, maybe load dependent. Could be a race condition. Or a bad interaction from whatever Intel installed on the system. We could schedule pocl first and then install the Intel toolkit. Or use different runners, but I wanted to avoid rebuilding chipStar for no good reason... Let me know what you'll rather do, or better, just go ahead and do it :)

Kerilk:
I'll rerun the job on branch 2 just to make sure I cannot reproduce before the merge. It may help pinpoint what is going on.

Kerilk:
I split the CI between Intel and pocl, but I still see sporadic pocl failures. I also upload artifacts after the run, so we should be able to investigate these failures.

pjaaskel:
https://github.com/intel/compute-runtime/issues/564 pls remember that issue

Kerilk:
@pjaaskel what do you mean? I know I am running around the issue inside this PR, but that's hardly been a problem for the CI

pjaaskel:
Is that ILLEGAL an illegal instruction exception? Could that be a case of wrongly detected ISA in PoCL so it generates too advanced instructions. We've had those cases with VMs as the LLVM detection was not coping with it very well. https://github.com/pocl/pocl/issues/793 just shooting in the dark here.

PR #629, opened 2023-09-22
FirstTouch workaround  via zeContextMakeMemoryResident

pvelesko:
The zeContextMakeMemoryResident function is used to make memory resident for a context, which means to allocate it in a way that it remains “pinned” in physical memory and is not paged out to disk.

FirstTouch test on Sunspot passses.

Added CMake option to enable/disable this - off by default since using the latest runtime should make it unnecessary. 

Performance impact: makes allocations slower so in the case of a lot of small allocations/deallocations (not common?) there will be a performance penalty.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -216,6 +216,7 @@ option(CHIP_ENABLE_NON_COMPLIANT_DEVICELIB_CODE "Enable non-compliant devicelib 
 option(CHIP_FAST_MATH "Use native_ OpenCL functions which are fast but their precision is implementation defined" OFF)
 option(CHIP_USE_INTEL_USM "enable support for cl_intel_unified_shared_memory in the OpenCL backend" OFF)
 option(CHIP_EXT_FLOAT_ATOMICS "Use cl_ext_float_atomics." ON)
+option(CHIP_L0_FIRST_TOUCH "First-touch workaroudn for Level Zero." ON)
 
 # Warpsize would optimally be a device-specific, queried and made
 # effective at runtime. However, we need to fix the warpsize since SPIR-Vs need
@@ -248,6 +249,10 @@ endif()
 
 if(CHIP_DUBIOUS_LOCKS)
   list(APPEND CHIP_SPV_DEFINITIONS CHIP_DUBIOUS_LOCKS)
+endif()
+
+if(CHIP_L0_FIRST_TOUCH)
+  list(APPEND CHIP_SPV_DEFINITIONS CHIP_L0_FIRST_TOUCH)
 endif()
 
 if(CHIP_ENABLE_NON_COMPLIANT_DEVICELIB_CODE)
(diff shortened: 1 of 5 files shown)


src/backend/Level0/CHIPBackendLevel0.cc
```diff
@@ -1899,17 +1897,25 @@ void *CHIPContextLevel0::allocateImpl(size_t Size, size_t Alignment,
...
     CHIPERR_CHECK_LOG_AND_THROW(Status, ZE_RESULT_SUCCESS,
                                 hipErrorMemoryAllocation);
+  } else
+    CHIPERR_LOG_AND_THROW("Failed to allocate memory",
+                          hipErrorMemoryAllocation);
 
-    return Ptr;
+#ifdef CHIP_L0_FIRST_TOUCH
+  CHIPDeviceLevel0 *ChipDev = static_cast<CHIPDeviceLevel0 *>(getDevice());
+  if (ChipDev) {
```
linehill, line 1911:
```suggestion
  if (auto *ChipDev = static_cast<CHIPDeviceLevel0 *>(getDevice())) {
```

linehill (approved):
A style nit. Otherwise, LGTM.

CMakeLists.txt
```diff
@@ -216,6 +216,7 @@ option(CHIP_ENABLE_NON_COMPLIANT_DEVICELIB_CODE "Enable non-compliant devicelib
 option(CHIP_FAST_MATH "Use native_ OpenCL functions which are fast but their precision is implementation defined" OFF)
 option(CHIP_USE_INTEL_USM "enable support for cl_intel_unified_shared_memory in the OpenCL backend" OFF)
 option(CHIP_EXT_FLOAT_ATOMICS "Use cl_ext_float_atomics." ON)
+option(CHIP_L0_FIRST_TOUCH "First-touch workaroudn for Level Zero." ON)
```
pjaaskel, line 219:
```suggestion
option(CHIP_L0_FIRST_TOUCH "First-touch workaround for Level Zero." ON)
```

pjaaskel: Can you link to the issue in comments of this flag? "First-touch workaround" might not say much to outside users nor to us in a few years time.

src/backend/Level0/CHIPBackendLevel0.cc
```diff
@@ -1899,17 +1897,24 @@ void *CHIPContextLevel0::allocateImpl(size_t Size, size_t Alignment,
...
     // TODO Check if devices support cross-device sharing?
     ze_result_t Status = zeMemAllocHost(ZeCtx, &HmaDesc, Size, Alignment, &Ptr);
     CHIPERR_CHECK_LOG_AND_THROW(Status, ZE_RESULT_SUCCESS,
                                 hipErrorMemoryAllocation);
+  } else
+    CHIPERR_LOG_AND_THROW("Failed to allocate memory",
+                          hipErrorMemoryAllocation);
 
-    return Ptr;
+#ifdef CHIP_L0_FIRST_TOUCH
```
pjaaskel, line 1909: Comment/documentation might be useful in this spot.

pjaaskel (changes requested):

PR #630, opened 2023-09-26
QueueSync without barriers

pvelesko:
Implement HIP/CUDA stream semantics using dependency chains via `LastEvent` as opposed to barriers. `Queue::getSyncQueuesLastEvens()` gathers `LastEvent` for all the queues it's supposed to sync. Backends can then extract handles and add dependencies to all enqueue commands.

### DIFF
--- a/src/backend/Level0/CHIPBackendLevel0.cc
+++ b/src/backend/Level0/CHIPBackendLevel0.cc
@@ -1486,16 +1486,16 @@ void CHIPQueueLevel0::finish() {
   LOCK(Backend->DubiousLockLevel0)
 #endif
 
-  if (static_cast<CHIPBackendLevel0 *>(Backend)->getUseImmCmdLists()) {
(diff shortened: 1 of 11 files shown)


src/backend/Level0/CHIPBackendLevel0.cc
```diff
@@ -821,6 +821,28 @@ CHIPQueueLevel0::~CHIPQueueLevel0() {
   }
 }
 
+std::vector<ze_event_handle_t> CHIPQueueLevel0::addDependenciesQueueSync(
+    std::shared_ptr<chipstar::Event> TargetEvent) {
+  auto EventsToWaitOn = getSyncQueuesLastEvents();
+  // Every event in EventsToWaitOn should have a dependency on MemCopyEvent so
```
pjaaskel, line 827: Cannot we get "last events" updated after the call and before we enter the next loop to add the next dep?

PR #633, opened 2023-09-27
use relaxed memory order for atomics

franz:
this fixes the devicelib atomics to use the relaxed order (same as CUDA)


README.md
```diff
@@ -35,6 +35,7 @@ While chipStar 1.0 can already be used to run various large HPC applications suc
 
   * An OpenCL 2.0 or 3.0 driver with at least the following features supported:
     * Coarse-grained buffer Shared Virtual Memory
+    * OpenCL 2.0 atomics (CL\_DEVICE\_SVM\_ATOMICS)
```
pjaaskel, line 38: Please note that this section is for the _required_ extensions (for any input CUDA/HIP application). If the application doesn't use atomics, they are not required, right? We shall list them in optional extensions (similar to ballots, shuffles,...)

pjaaskel (changes requested):

franz:
```
CMake Error at cmake/FindLLVM.cmake:33 (message):
  Provided LLVM_CONFIG_BIN (/home/runner/opt/llvm/16/bin/llvm-config) does
  not exist
```

@pvelesko something up with the CI ?

pvelesko:
@franz nope, CI is all good. Where do you see that error?

README.md
franz: I removed it, upon further reading `CL_DEVICE_SVM_ATOMICS` is not required.

PR #637, opened 2023-10-02
CI Update new runtime modules + parallel ctest + default to ICL

pvelesko:
* Use new runtime modules
* Build & run tests in parallel to take advantage of the new 24 core CPU on the cupcake
* Running ctest in parallel causes a couple of cases to fail - known previously but still needs to be investigated. Excluding for now.
* sycl-HIP interop segfault after using new modules - Excluding for now and will open a new issue once this is merged. 
* When tests are run in parallel, a lot more tests fail for regular command lists now - defaulting to immediate command lists from now on
* Re-enable LLVM-15 Testing


pjaaskel:
Let me test this (ICL by default) with my local envs.

pvelesko:
@pjaaskel ok thanks, please make sure that the runtime is up-to-date. Recent versions seem to have resolved some ICL issues.

PR #638, opened 2023-10-02
Implement multiplexing strategy

Kerilk:
This implements a multiplexing strategy for HIP similar to Vulkan, where a loader can load installable client driver. It does so by adding a void pointer as the first field of handles, that the loader can use to store it's dispatching information.

Given our virtual classes, objects start by their virtual table, so I use a placement new strategy inside an allocation obtained by malloc. AMD uses a similar strategy in their Vulkan driver.

For more details on the multiplexing strategy please refer to ANL corresponding milestones.

A wrapper library (if it can be cleanly done) would add way more complexity and overhead. We can discuss this in our biweekly as well.

### DIFF
--- a/src/CHIPBackend.cc
+++ b/src/CHIPBackend.cc
@@ -801,7 +801,8 @@ void chipstar::Device::eraseModule(chipstar::Module *Module) {
   LOCK(DeviceMtx); // SrcModToCompiledMod_
   for (auto &Kv : SrcModToCompiledMod_)
     if (Kv.second == Module) {
-      delete Module;
+      Module->~Module();
+      free(CHIP_OBJ_TO_HANDLE(Module, ihipModule_t));
       SrcModToCompiledMod_.erase(Kv.first);
       break;
     }
--- a/src/CHIPBackend.cc
+++ b/src/CHIPBackend.cc
@@ -542,10 +542,8 @@ chipstar::Queue *chipstar::Device::getPerThreadDefaultQueue() {
 
 void chipstar::Device::QueueDeleter::operator()(
     chipstar::Queue *q) const noexcept {
-  if (q) {
-    q->~Queue();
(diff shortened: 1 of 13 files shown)


src/CHIPBackend.cc
```diff
@@ -540,11 +542,20 @@ chipstar::Queue *chipstar::Device::getPerThreadDefaultQueue() {
   return getPerThreadDefaultQueueNoLock();
 }
 
+void chipstar::Device::QueueDeleter::operator()(
+    chipstar::Queue *q) const noexcept {
+  if (q) {
+    q->~Queue();
```
pjaaskel, line 548: What's going on here? Calling constructor directly and freeing the handle. Why not put the freeing to the Queue's destructor and delete it?

pjaaskel (commented):
I do not fully follow what is being done in this PR. Can you describe what you exactly plan to do here? Did I understand it right that we want to make L0 and CL pluggable (runtime loadable) so they can be optionally enabled and packaged without needing to have deps for either unless calling them. PoCL has a similar (optional) mechanism in its driver layer. Could we implement it via a plugin layer? That is, add a plugin function which creates the C++ instance of the correct concrete backend class.

Kerilk:
I opened this yesterday to discuss what was done with @pvelesko . I should have it labeled draft. I'll elaborate above. The idea is to refactor and incorporate later this year.

linehill:
I pushed a commit for proposing an alternate approach that does not need placement new and manual destructor invocation. This is demonstrated with hipStream_t/chipstar::Queue in the commit ([a44162c](https://github.com/CHIP-SPV/chipStar/pull/638/commits/a44162c5d844d07983ae990d0f3721c37fac9d64)).

pvelesko:
bitrotten so will close for now

PR #645, opened 2023-10-06
Link chipStar's LLVM pass plugin to LLVM

linehill:
Additionally, ensure there is no undefined references while linking the plugin to LLVM (`--no-undefined`).

This attempts to fix issue seen in https://github.com/CHIP-SPV/chipStar/pull/570 where the plugin loaded by statically built opt tool does not provide the symbols needed by the plugin.


linehill:
Abandoned: this patch does not work well with static linked opt tool and it is not worthwhile getting it work currently.

PR #650, opened 2023-10-14
Cmake cleanup

pvelesko:
* Update README
* Move some cmake options to config.hh.in
* Rename CHIPSPVConfig.hh.in to chipStarConfig.hh.in
* Remove old `queueSync` code

- [x] Remove old check.py
- [x] Runtime check for disabling ICL if iGPU is detected
- [x] checkv2 error code


README.md
```diff
@@ -117,6 +117,39 @@ python3 $SOURCE_DIR/scripts/check.py $BUILD_DIR $DEVICE $BACKEND $PARALLEL 1
...
+
+```bash
+CHIP_BE=<opencl/level0>                         # Selects the backend to use
+CHIP_PLATFORM=<N>                               # If there are multiple platforms present on the system, selects which one to use
+CHIP_DEVICE=<N>                                 # If there are multiple devices present on the system, selects which one to use
+CHIP_LOGLEVEL=<trace/debug/info/warn/err/crit>  # Sets the log level. If compiled in RELEASE, only err/crit are available
+CHIP_DUMP_SPIRV=<ON/OFF>                        # Dumps the generated SPIR-V code to a file
+CHIP_JIT_FLAGS=<flags>                          # String to override the default JIT flags (-x spir -cl-kernel-arg-info -cl-std=CL3.0)
+CHIP_L0_COLLECT_EVENTS_TIMEOUT=<N(30s default)> # Timeout in milliseconds for collecting Level Zero events
+CHIP_L0_IMM_CMD_LISTS=<ON(default)/OFF>         # Use immediate command lists in Level Zero
```
linehill, line 130: Please document the default values to the variables.

src/backend/Level0/CHIPBackendLevel0.hh
```diff
@@ -566,14 +566,27 @@ public:
...
-  void setUseImmCmdLists() {
+  void setUseImmCmdLists(std::string_view DeviceName) {
+    // Immediate command lists seem to not work on some Intel iGPUs
+    // https://en.wikipedia.org/wiki/List_of_Intel_graphics_processing_units
+    std::vector<std::string> IgpuDevices = {"UHD", "HD", "Iris"};
+    // check if the device name contains any of the strings in IgpuDevices
+    bool IsIgpu = std::any_of(IgpuDevices.begin(), IgpuDevices.end(),
+                              [&](const std::string &S) {
+                                return DeviceName.find(S) != std::string::npos;
+                              });
```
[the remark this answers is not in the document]
pvelesko: Level Zero is Intel exclusive. @linehill

scripts/check.py
```diff
@@ -1,73 +1,66 @@
...
-    if (len(sys.argv) != 6):
-        print("not enough args")
-        print(sys.argv)
-        print(usage_line)
-        sys.exit(1)
+parser.add_argument('work_dir', type=str, help='Path to build directory')
+parser.add_argument('device_type', type=str, choices=['cpu', 'igpu', 'dgpu'], help='Device type')
+parser.add_argument('backend', type=str, choices=['opencl', 'level0-reg', 'level0-imm', 'pocl'], help='Backend to use')
+parser.add_argument('num_threads', type=int, help='Number of threads to use')
+parser.add_argument('num_tries', type=int, help='Number of tries')
```
linehill, line 24: Some of these options could be optional for convenience - like:
* num_threads which defaults to one or number of processor cores in the system.
* num_tries which defaults to one.

src/backend/Level0/CHIPBackendLevel0.hh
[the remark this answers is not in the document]
linehill: True (for now).

src/backend/Level0/CHIPBackendLevel0.hh
```diff
@@ -566,14 +566,27 @@ public:
...
     // assert that str is either 0, 1, on, off or empty
     assert((str.empty() || str == "0" || str == "1" || str == "on" ||
             str == "off") &&
            "CHIP_L0_IMM_CMD_LISTS must be 0, 1, on, off or "
-           "empty (default: off)");
-    useImmCmdLists_ = (str == "1" || str == "on");
+           "empty (default: on)");
+    useImmCmdLists_ = (str == "1" || str == "on" || str.empty());
+    if (IsIgpu && useImmCmdLists_) {
+      logWarn("Immediate command lists are not supported on this device. "
```
pjaaskel, line 586: Maybe we want to still allow the user to force set ICL using the env var even if it's listed as a broken device just to test/debug? It works for most test cases after all.

pjaaskel (approved):
Some of the commit messages are useless and should be improved (if you rebase), otherwise looks quite good.

src/backend/Level0/CHIPBackendLevel0.hh
pvelesko: I guess I can leave it as a warning and and not force off

README.md
```diff
@@ -117,6 +117,39 @@ python3 $SOURCE_DIR/scripts/check.py $BUILD_DIR $DEVICE $BACKEND $PARALLEL 1
 
 Please refer to the [user documentation](docs/Using.md) for instructions on how to use the installed chipStar to build CUDA/HIP programs.
 
+## Environment Variables
```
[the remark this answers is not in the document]
pvelesko: I prefer first page since it's the most basic of use

pvelesko:
> Some of the commit messages are useless and should be improved (if you rebase), otherwise looks quite good.

Since we do squash merge, there will be custom, single, commit going into main.

pjaaskel:
The new check.py doesn't seem to output the test progress. Does it buffer it? It would be useful to see the tests running.

Also, in check.py the "pocl" env/backend is redundant since PoCL is just an OpenCL implementation. It doesn't seem to do anything special when "pocl" is used instead of "opencl".

cmake/UnitTests.cmake
```diff
@@ -16,9 +16,10 @@ list(APPEND GPU_POCL_FAILED_TESTS " ")  # TODO
 # It fails with "error: cannot find ROCm device library;
 #  provide its path via '--rocm-path' or '--rocm-device-lib-path', or pass
 #  '-nogpulib' to build without ROCm device library"
-list(APPEND FAILING_FOR_ALL "Unit_hipMalloc_LoopRegressionAllocFreeCycles") # 
-list(APPEND FAILING_FOR_ALL "Unit_hipGraphAddHostNode_ClonedGraphwithHostNode") #
-list(APPEND FAILING_FOR_ALL "Unit_hipGraphDestroyNode_DestroyDependencyNode") #
+list(APPEND FAILING_FOR_ALL "abort") #
```
pjaaskel: Don't these work with OpenCL in your problematic platform? So, excluding in the L0 also should work? Can you also add some comments here so we remember what you've found of the failure?

PR #665, opened 2023-10-31
Recycle L0 Command Lists

pvelesko:
- [x] Merge #664
- [x] Move command list storage to `Context`


src/backend/Level0/CHIPBackendLevel0.cc
```diff
@@ -850,18 +886,55 @@ void CHIPQueueLevel0::addCallback(hipStreamCallback_t Callback,
...
+    // via CHIPEventLevel0::dissassociateCmdList()
+    auto ChipDevLz = static_cast<CHIPDeviceLevel0 *>(ChipDevice_);
+    ze_command_list_desc_t Desc = ChipDevLz->getCommandListComputeDesc();
     auto Status =
-        zeCommandListCreate(ZeCtx_, ZeDev_, &CommandListDesc_, &ZeCmdList);
+        zeCommandListCreate(this->ZeCtx, ChipDevLz->get(), &Desc, &ZeCmdList);
     CHIPERR_CHECK_LOG_AND_THROW(Status, ZE_RESULT_SUCCESS,
                                 hipErrorInitializationError);
-    return ZeCmdList;
+    NumCmdListsCreated_++;
```
linehill, line 917: Unused member variable? I don't see this variable being observed anywhere.

src/backend/Level0/CHIPBackendLevel0.cc
```diff
@@ -850,18 +886,55 @@ void CHIPQueueLevel0::addCallback(hipStreamCallback_t Callback,
...
+  //      "ICL is used - this should never be called");
+  LOCK(CmdListMtx) // CHIPQueueLevel0::ZeCmdListRegStack_
+  // TODO this check doesn't work since we use RCL for callbacks even when ICL
+  // if (NumCmdListsCreated_ < ZeCmdListRegStack_.size() + 1) {
+  //   logError("{} CHIPContextLevel0::returnCmdList({}) returning more cmd
+  //   lists "
+  //            "than created",
+  //            (void *)this, (void *)ZeCmdList);
+  //   assert(false);
+  // }
```
linehill, line 936: This sounds like there is a need for refactoring. It seems the issue can be solved by passing `ze_command_list_handle_t`s around wrapped in a structure which carries the needed information - E.g:

```c++
struct CommandList {
  ze_command_list_handle_t Handle;
  bool IsImmediate;
};
```

src/backend/Level0/CHIPBackendLevel0.cc
```diff
@@ -636,99 +670,100 @@ void CHIPCallbackEventMonitorLevel0::monitor() {
...
+
+    // if this is the first time we are stopping, set the time
+    if (TimeSinceStopRequested_ == 0)
+      TimeSinceStopRequested_ = CurrTime;
+
+    int EpasedTime = CurrTime - this->TimeSinceStopRequested_;
+    bool AllEventsCleared = Backend->Events.size() == 0;
+    if (AllEventsCleared) {
+      pthread_exit(0);
+    } else if (EpasedTime > BackendZe->getCollectEventsTimeout()) {
```
linehill, line 730: [Else-after-return](https://llvm.org/docs/CodingStandards.html#don-t-use-else-after-a-return). 
```suggestion
    if (AllEventsCleared)
      pthread_exit(0);

    if (EpasedTime > BackendZe->getCollectEventsTimeout()) {
```
There is another one in the below too.

src/backend/Level0/CHIPBackendLevel0.cc
```diff
@@ -850,18 +886,55 @@ void CHIPQueueLevel0::addCallback(hipStreamCallback_t Callback,
 
 ze_command_list_handle_t CHIPQueueLevel0::getCmdList() {
   if (static_cast<CHIPBackendLevel0 *>(Backend)->getUseImmCmdLists()) {
-    return ZeCmdList_;
+    logTrace("CHIPQueueLevel0::getCmdList() returning ICL {}",
+             (void *)ZeCmdListImm_);
+    return ZeCmdListImm_;
   } else {
```
linehill, line 892: Nit: else-after-return.

src/backend/Level0/CHIPBackendLevel0.hh
```diff
@@ -323,8 +352,30 @@ public:
...
 public:
+  /**
+   * @brief Either return the immediate command list for this queue or a regular
+   * command list depending on if ICL is used or not. If ICL is not used, the
+   * regular command list can be created new or returnend from the stack of
+   * previously created handles.
+   *
+   * @return ze_command_list_handle_t
+   */
+  ze_command_list_handle_t getCmdListReg();
```
linehill, line 370: The description does not match function definition.

src/backend/Level0/CHIPBackendLevel0.hh
```diff
@@ -323,8 +352,30 @@ public:
...
+
+  /**
+   * @brief reset the given command list and return it back to the stack
```
linehill, line 373:
```suggestion
   * @brief reset the given command list and return it back to the pool
```
Should also apply the change on the other occurrences of the "stack".

src/CHIPBindings.cc
```diff
@@ -2090,6 +2090,8 @@ hipError_t hipStreamDestroy(hipStream_t Stream) {
     RETURN(hipErrorStreamCaptureInvalidated);
   }
 
+  LOCK(ChipQueue->QueueMtx);
```
linehill, line 2093: What is the purpose of this lock?

src/backend/Level0/CHIPBackendLevel0.cc
```diff
@@ -636,99 +670,100 @@ void CHIPCallbackEventMonitorLevel0::monitor() {
...
+      int PrintedEntries = 0;
+      for (auto &Event : Backend->Events) {
+        auto EventLz = std::static_pointer_cast<CHIPEventLevel0>(Event);
+        logError("Uncollected Backend->Events: {} {} AssocCmdList {}",
+                 (void *)Event.get(), Event->Msg,
+                 (void *)EventLz->getAssocCmdList());
+        PrintedEntries++;
+        if (PrintedEntries > MaxPrintEntries)
+          break;
+      }
```
linehill, line 744:
```suggestion
      size_t MaxPrintEntries = std::min(Backend->Events.size(), 10);
      for (size_t i = 0; i < MaxPrintEntries; i++) {
        auto EventLz = std::static_pointer_cast<CHIPEventLevel0>(Backend->Events[i]);
        logError("Uncollected Backend->Events: {} {} AssocCmdList {}",
                 (void *)Event.get(), Event->Msg,
                 (void *)EventLz->getAssocCmdList());
      }
```

src/backend/Level0/CHIPBackendLevel0.hh
```diff
@@ -175,6 +183,17 @@ class CHIPStaleEventMonitorLevel0 : public chipstar::EventMonitor {
   int TimeSinceStopRequested_ = 0;
   int LastPrint_ = 0;
 
+  /**
+   * @brief Go through all events in Backend::Events, update their status, upon
+   * status change release dependencies and command lists, return to event pool
+   */
+  void checkEvents_();
```
linehill, line 190:
```suggestion
  void checkEvents();
```

src/backend/Level0/CHIPBackendLevel0.hh
```diff
@@ -175,6 +183,17 @@ class CHIPStaleEventMonitorLevel0 : public chipstar::EventMonitor {
...
+  /**
+   * @brief Check if stop was requested for this monitor, if so handle all
+   * outstanding events
+   */
+  void exitChecks_();
```
linehill, line 195:
```suggestion
  void exitChecks();
```

src/backend/Level0/CHIPBackendLevel0.cc
```diff
@@ -215,6 +211,28 @@ createSampler(CHIPDeviceLevel0 *ChipDev, const hipResourceDesc *PResDesc,
...
+void CHIPEventLevel0::disassociateCmdList() {
+  assert(AssocCmdList_ != nullptr && "command list not associated!");
+  assert(AssocContext_ != nullptr && "queue not associated!");
+  logTrace("CHIPEventLevel0({})::disassociateCmdList({})", (void *)this,
+           (void *)AssocCmdList_);
+  auto Status = zeCommandListReset(AssocCmdList_);
+  CHIPERR_CHECK_LOG_AND_THROW(Status, ZE_RESULT_SUCCESS, hipErrorTbd);
+  AssocContext_->returnCmdList(AssocCmdList_);
+  AssocCmdList_ = nullptr;
+  AssocContext_ = nullptr;
```
linehill, line 233: This looks like a logic that should be in CHIPEventLevel0::reset(). Now, it looks like there is event reset logic in two places.

src/backend/Level0/CHIPBackendLevel0.hh
```diff
@@ -323,8 +352,30 @@ public:
 class CHIPContextLevel0 : public chipstar::Context {
   OpenCLFunctionInfoMap FuncInfos_;
   std::vector<LZEventPool *> EventPools_;
+  std::mutex CmdListMtx;
+  size_t NumCmdListsCreated_ = 0;
+  size_t CmdListsRequested_ = 0;
+  size_t CmdListsReused_ = 0;
+  std::stack<ze_command_list_handle_t> ZeCmdListRegStack_;
```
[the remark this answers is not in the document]
pvelesko: I wanted to avoid this since there is a level-zero notion of an EventPool so I didn't want to confuse.

src/backend/Level0/CHIPBackendLevel0.hh
pvelesko: see comment above

src/backend/Level0/CHIPBackendLevel0.hh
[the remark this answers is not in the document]
linehill: How is level-zero's notion of object pools different to chipStar's notion?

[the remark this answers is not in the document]
pvelesko: Event Pool is a thing that you can create in level zero. There is no such thing as an Command List Pool and I don't want to imply so. Because of that, and because it's a literal stack object, I think think ZeCmdListRegStack_ is better.

src/backend/Level0/CHIPBackendLevel0.cc
pvelesko: This does not reset the event, rather only deals with release the associated command list when this event fires. Event can fire and then still remain unreset. If we go with a struct for containing command lists then ```  auto Status = zeCommandListReset(AssocCmdList_);
  CHIPERR_CHECK_LOG_AND_THROW(Status, ZE_RESULT_SUCCESS, hipErrorTbd);``` could go into the `CmdList.reset()`` and that's how I had it in my initial draft but decided to keep it as simple as possible for now.

src/backend/Level0/CHIPBackendLevel0.hh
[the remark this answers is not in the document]
pvelesko: We can go with event pool - maybe I was overthinking it.

src/backend/Level0/CHIPBackendLevel0.cc
pvelesko: Leftover comments - we never need to associate events with immediate command lists since those are self-managed.

src/CHIPBindings.cc
pvelesko: leftover from the 1st attempt

src/backend/Level0/CHIPBackendLevel0.cc
linehill: >This does not reset the event, rather only deals with release the associated command list when this event fires. Event can fire and then still remain unreset.

The events associated to command lists, which are disassociated in the event garbage collector, are going to be returned event pool soon and eventually to be reset, right? What are the issues that prevent the command list disassociate logic to live in CHIPEventLevel0::reset()?

pvelesko: > are going to be returned event pool soon and eventually to be reset, right? 

not necessarily right away. An event could fire (at which point we don't need the command list and can recycle it, thus disassociate it) but still remain refcounted somewhere for example in a dependency chain somewhere.

src/backend/Level0/CHIPBackendLevel0.cc
```diff
@@ -235,10 +253,23 @@ ze_event_handle_t CHIPEventLevel0::peek() {
...
+    logTrace("~CHIPEventLevel0({}) waiting for event to finish", (void *)this);
+    wait();
+  }
+
+  if (AssocCmdList_ || AssocContext_) {
+    logError("~CHIPEventLevel0({}) disassociating command list {}",
+             (void *)this, (void *)AssocCmdList_);
+    logError("~CHIPEventLevel0({}) disassociating queue {}", (void *)this,
+             (void *)AssocContext_);
+    // assert(false);
```
linehill, line 268: Nit: should clean unused code.

linehill (approved):
LGTM.

It seems there is an opportunity to combine logic of  disassociateCmdList() and CHIPEventLevel0::reset() together but that can be done later.

PR #675, opened 2023-11-08
Implement an env var storage class

pvelesko:
Collect all environment variables into a single class and avoid calls to read the environment more than once.


CMakeLists.txt
```diff
@@ -216,6 +216,7 @@ option(CHIP_EXT_FLOAT_ATOMICS "Use cl_ext_float_atomics." ON)
 # If your application does a lot of small allocations, it might be slow
 # https://github.com/CHIP-SPV/chipStar/issues/458 https://github.com/intel/compute-runtime/issues/631
 option(CHIP_L0_FIRST_TOUCH "First-touch workaround for Level Zero." OFF)
+set(CHIP_DEFAULT_JIT_FLAGS "-cl-kernel-arg-info -cl-std=CL3.0")
```
linehill, line 219: These are OpenCL compilation options. Aren't these passed to level zero BE too? Does level zero support these options?

pvelesko: Yes it's the same set of flags for both backends.

src/CHIPDriver.hh
```diff
@@ -118,4 +120,171 @@ extern std::atomic_ulong CHIPNumRegisteredFatBinaries;
...
+    if (!readEnvVar("CHIP_L0_COLLECT_EVENTS_TIMEOUT").empty())
+      L0CollectEventsTimeout = parseInt("CHIP_L0_COLLECT_EVENTS_TIMEOUT");
+  }
+
+  std::string parseJitFlags(const std::string &StrIn) {
+    auto str = readEnvVar(StrIn);
+    if (str.empty()) {
+      return CHIP_DEFAULT_JIT_FLAGS;
+    }
+    return str;
```
linehill, line 218: Nit: [braces](https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements)
```suggestion
    auto str = readEnvVar(StrIn);
    if (str.empty())
      return CHIP_DEFAULT_JIT_FLAGS;
    return str;
```

src/CHIPDriver.hh
```diff
@@ -118,4 +120,171 @@ extern std::atomic_ulong CHIPNumRegisteredFatBinaries;
...
+    if (!readEnvVar("CHIP_DUMP_SPIRV").empty())
+      DumpSpirv = parseBoolean("CHIP_DUMP_SPIRV");
+    JitFlags = parseJitFlags("CHIP_JIT_FLAGS_OVERRIDE");
+    if (!readEnvVar("CHIP_L0_IMM_CMD_LISTS").empty())
+      L0ImmCmdLists = parseBoolean("CHIP_L0_IMM_CMD_LISTS");
+    if (!readEnvVar("CHIP_L0_COLLECT_EVENTS_TIMEOUT").empty())
+      L0CollectEventsTimeout = parseInt("CHIP_L0_COLLECT_EVENTS_TIMEOUT");
+  }
+
+  std::string parseJitFlags(const std::string &StrIn) {
```
linehill, line 213: Could we avoid returning a copy? E.g. by returning a const reference or std::string_view?

src/CHIPDriver.hh
```diff
@@ -118,4 +120,171 @@ extern std::atomic_ulong CHIPNumRegisteredFatBinaries;
...
+    return std::stoi(str);
+  }
+
+  bool parseBoolean(const std::string &StrIn) {
+    auto str = readEnvVar(StrIn);
+    if (str == "1" || str == "on") {
+      return true;
+    } else if (str == "0" || str == "off") {
+      return false;
+    }
```
linehill, line 236:
```suggestion
    if (str == "1" || str == "on")
      return true;
    if (str == "0" || str == "off") {
      return false;
```

src/CHIPDriver.hh
```diff
@@ -118,4 +120,171 @@ extern std::atomic_ulong CHIPNumRegisteredFatBinaries;
...
+      return DeviceType(DeviceType::GPU);
+    } else if (str == "cpu") {
+      return DeviceType(DeviceType::CPU);
+    } else if (str == "accel") {
+      return DeviceType(DeviceType::ACCEL);
+    } else if (str == "fpga") {
+      return DeviceType(DeviceType::FPGA);
+    } else if (str == "") {
+      return DeviceType(DeviceType::DEFAULT);
+    }
```
linehill, line 255:
```suggestion
    if (str == "gpu")
      return DeviceType(DeviceType::GPU);
    if (str == "cpu")
      return DeviceType(DeviceType::CPU);
    if (str == "accel")
      return DeviceType(DeviceType::ACCEL);
    if (str == "fpga")
      return DeviceType(DeviceType::FPGA);
    if (str == "")
      return DeviceType(DeviceType::DEFAULT);
```

src/CHIPDriver.hh
```diff
@@ -118,4 +120,171 @@ extern std::atomic_ulong CHIPNumRegisteredFatBinaries;
...
+
+  BackendType parseBackendType(const std::string &StrIn) {
+    auto str = readEnvVar(StrIn);
+    if (str == "opencl") {
+      return BackendType(BackendType::OPENCL);
+    } else if (str == "level0") {
+      return BackendType(BackendType::LEVEL0);
+    } else if (str == "") {
+      return BackendType(BackendType::DEFAULT);
+    }
```
linehill, line 270: Nit: braces and else-after-return.

src/CHIPDriver.hh
```diff
@@ -118,4 +120,171 @@ extern std::atomic_ulong CHIPNumRegisteredFatBinaries;
...
+      return "fpga";
+    case DEFAULT:
+      return "default";
+    default:
+      return "unknown";
+    }
+  }
+
+  Type getType() const { return Type_; }
+};
```
linehill, line 151: Seems excessive. This could be simplified to enum class and free-standing `const std::string& toString(DeviceType Type)`.

src/CHIPDriver.hh
```diff
@@ -118,4 +120,171 @@ extern std::atomic_ulong CHIPNumRegisteredFatBinaries;
...
+class EnvVars {
+public:
+  int PlatformIdx = 0;
+  DeviceType Device = DeviceType::GPU;
+  int DeviceIdx = 0;
+  BackendType Backend = BackendType::DEFAULT;
+  bool DumpSpirv = false;
+  std::string JitFlags = CHIP_DEFAULT_JIT_FLAGS;
+  bool L0ImmCmdLists = true;
+  int L0CollectEventsTimeout = 0;
```
linehill, line 188: Should these be immutable after environment variables are parsed?

src/CHIPDriver.hh
```diff
@@ -118,4 +120,171 @@ extern std::atomic_ulong CHIPNumRegisteredFatBinaries;
  */
 extern thread_local hipError_t CHIPTlsLastError;
 
+class DeviceType {
+public:
+  enum Type { GPU, CPU, ACCEL, FPGA, DEFAULT };
```
linehill, line 125: Nit: [Naming convention](https://llvm.org/docs/CodingStandards.html#name-types-functions-variables-and-enumerators-properly).
```suggestion
  enum Type { GPU, CPU, Accelerator, FPGA, Default };
```

src/CHIPDriver.hh
```diff
@@ -118,4 +120,171 @@ extern std::atomic_ulong CHIPNumRegisteredFatBinaries;
...
+      return "unknown";
+    }
+  }
+
+  Type getType() const { return Type_; }
+};
+
+class BackendType {
+public:
+  enum Type { OPENCL, LEVEL0, DEFAULT };
```
linehill, line 155: Nit: naming convention.
```suggestion
  enum Type { OpenCL, Level0, Default };
```

src/CHIPDriver.hh
pvelesko: Making them const would complicate constructor but I'll make them private.

src/CHIPDriver.hh
pvelesko: I don't see any benefit to having related functions outside of the class definition.

src/CHIPDriver.hh
linehill: Sorry, I meant immutability respect to class users. Making them private also makes them immutable for the users.

src/CHIPDriver.hh
```diff
@@ -118,4 +120,182 @@ extern std::atomic_ulong CHIPNumRegisteredFatBinaries;
...
+    case CPU:
+      return "cpu";
+    case Accelerator:
+      return "accel";
+    case FPGA:
+      return "fpga";
+    case Default:
+      return "default";
+    default:
+      return "unknown";
```
linehill, line 146: We probably want attention about unhandled cases.
```suggestion
    default:
      assert(!"Unknown device type!");
      return "unknown";
```

src/CHIPDriver.hh
```diff
@@ -118,4 +120,182 @@ extern std::atomic_ulong CHIPNumRegisteredFatBinaries;
...
+  const char *str() const {
+    switch (Type_) {
+    case OpenCL:
+      return "opencl";
+    case Level0:
+      return "level0";
+    case Default:
+      return "default";
+    default:
+      return "unknown";
```
linehill, line 172:
```suggestion
    default:
      assert(!"Unknown backend type!");
      return "unknown";
```

src/CHIPDriver.hh
pvelesko: So does this current solution work for you then?

src/CHIPDriver.hh
```diff
@@ -118,4 +120,184 @@ extern std::atomic_ulong CHIPNumRegisteredFatBinaries;
...
+      L0ImmCmdLists_ = parseBoolean("CHIP_L0_IMM_CMD_LISTS");
+    if (!readEnvVar("CHIP_L0_COLLECT_EVENTS_TIMEOUT").empty())
+      L0CollectEventsTimeout_ = parseInt("CHIP_L0_COLLECT_EVENTS_TIMEOUT");
+  }
+
+  std::string_view parseJitFlags(const std::string &StrIn) {
+    auto str = readEnvVar(StrIn);
+    if (str.empty()) {
+      return CHIP_DEFAULT_JIT_FLAGS;
+    }
```
linehill, line 232:
```suggestion
    if (str.empty())
      return CHIP_DEFAULT_JIT_FLAGS;

```

src/CHIPDriver.hh
```diff
@@ -118,4 +120,184 @@ extern std::atomic_ulong CHIPNumRegisteredFatBinaries;
...
+    parseEnvironmentVariables();
+    logDebugSettings();
+  }
+
+  int getPlatformIdx() const { return PlatformIdx_; }
+  DeviceType getDevice() const { return Device_; }
+  int getDeviceIdx() const { return DeviceIdx_; }
+  BackendType getBackend() const { return Backend_; }
+  bool getDumpSpirv() const { return DumpSpirv_; }
+  std::string getJitFlags() const { return JitFlags_; }
```
linehill, line 203:
```suggestion
  const std::string &getJitFlags() const { return JitFlags_; }
```

src/CHIPDriver.hh
```diff
@@ -118,4 +120,184 @@ extern std::atomic_ulong CHIPNumRegisteredFatBinaries;
...
+public:
+  enum Type { GPU, CPU, Accelerator, FPGA, Default };
+
+private:
+  Type Type_;
+
+public:
+  DeviceType(Type TypeIn) : Type_(TypeIn) {}
+
+  std::string str() const {
```
linehill, line 133: To avoid a copy:
```suggestion
  const char *str() const {
```
Or alternatively:
```suggestion
  std::string_view str() const {
```

src/CHIPDriver.hh
```diff
@@ -118,4 +120,184 @@ extern std::atomic_ulong CHIPNumRegisteredFatBinaries;
...
+  std::string_view parseJitFlags(const std::string &StrIn) {
+    auto str = readEnvVar(StrIn);
+    if (str.empty())
+      return CHIP_DEFAULT_JIT_FLAGS;
+
+    return JitFlags_;
+  }
+
+  int parseInt(const std::string &StrIn) {
+    auto str = readEnvVar(StrIn);
```
linehill, line 237: To avoid a copy:
```suggestion
    const auto &Str = readEnvVar(StrIn);
```
Also, a nit for naming convention. Added `const`because Str seems not to be modified.

src/CHIPDriver.hh
```diff
@@ -118,4 +120,184 @@ extern std::atomic_ulong CHIPNumRegisteredFatBinaries;
...
+    }
+    if (!readEnvVar("CHIP_L0_IMM_CMD_LISTS").empty())
+      L0ImmCmdLists_ = parseBoolean("CHIP_L0_IMM_CMD_LISTS");
+    if (!readEnvVar("CHIP_L0_COLLECT_EVENTS_TIMEOUT").empty())
+      L0CollectEventsTimeout_ = parseInt("CHIP_L0_COLLECT_EVENTS_TIMEOUT");
+  }
+
+  std::string_view parseJitFlags(const std::string &StrIn) {
+    auto str = readEnvVar(StrIn);
+    if (str.empty())
```
linehill, line 230: Unnecessary copy due to `auto` instead of `auto&`:
```suggestion
    if (readEnvVar(StrIn).empty())
```

src/CHIPDriver.hh
```diff
@@ -118,4 +120,184 @@ extern std::atomic_ulong CHIPNumRegisteredFatBinaries;
...
+  int parseInt(const std::string &StrIn) {
+    auto str = readEnvVar(StrIn);
+    if (!isConvertibleToInt(str))
+      CHIPERR_LOG_AND_THROW("Invalid integer value: " + str,
+                            hipErrorInitializationError);
+    return std::stoi(str);
+  }
+
+  bool parseBoolean(const std::string &StrIn) {
+    auto str = readEnvVar(StrIn);
```
linehill, line 245: Naming convention, an unnecessary copy and opportunity for add `const` qualifier:
```suggestion
    const auto &Str = readEnvVar(StrIn);
```
There are many other occurrences of this in the file.

src/CHIPDriver.hh
pvelesko: Seems to me like this would be handled by the most basic return value optimization?

src/CHIPDriver.hh
pvelesko: Why wouldn't these instances get optimized away?

src/CHIPDriver.hh
linehill: The RVO eliminates copy for a value returned from a function - IOW, the value returned by `readEnvVar()` (may) not involve a copy. There is a copy due to assignment to `std::string str` (AFAIK, `auto` defaults to value not a reference).

src/CHIPDriver.hh
linehill: They could. Alas, optimizations are not guaranteed by C++ spec to occur.

src/CHIPDriver.hh
```diff
@@ -118,4 +120,183 @@ extern std::atomic_ulong CHIPNumRegisteredFatBinaries;
...
+      return true;
+    if (str == "0" || str == "off")
+      return false;
+    CHIPERR_LOG_AND_THROW("Invalid boolean value: " + str + "while parsing " +
+                              StrIn,
+                          hipErrorInitializationError);
+    return false; // This return is never reached
+  }
+
+  DeviceType parseDeviceType(const std::string &StrIn) {
```
linehill, line 255: Looks like there is a separation of concerns issue. The concern of constructing DeviceType from a string looks something the class should do. E.g. via `static DeviceType fromString(const std::string&);`.

src/CHIPDriver.hh
```diff
@@ -118,4 +120,183 @@ extern std::atomic_ulong CHIPNumRegisteredFatBinaries;
...
+    if (str == "")
+      return DeviceType(DeviceType::Default);
+
+    CHIPERR_LOG_AND_THROW("Invalid device type value: " + str +
+                              " while parsing " + StrIn,
+                          hipErrorInitializationError);
+    return DeviceType(DeviceType::GPU); // This return is never reached
+  }
+
+  BackendType parseBackendType(const std::string &StrIn) {
```
linehill, line 274: Same as for the parseDeviceType(). This function looks like it belongs to the BackendType class.

src/CHIPDriver.hh
pvelesko: Yeah this makes sense and is a good practice but seems like we're really splitting hairs here over a function that gets invoked once.

linehill: Why not learn a habit to apply good practices? The change here shouldn't take much time.

pvelesko: Yeah you're right I'll make sure to double check these things going forward.

linehill (approved):
LGTM. Thanks.

Issue #657, opened 2023-10-22
CHIPQueueOpenCL::MemMap bug Unit_hipMultiThreadStreams1_AsyncSame `invalid free()` 


pvelesko:
```
Thread 5 (Thread 0x7f964810e640 (LWP 175973)):
#0  0x00007f96498ebb3f in ioctl () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x00007f96488c6be0 in NEO::Drm::ioctl(NEO::DrmIoctl, void*) () from /home/pvelesko/install/intel/neo/2023.10.02/lib/intel-opencl/libigdrcl.so
#2  0x00007f96488c0d5b in NEO::Drm::waitHandle(unsigned int, long) () from /home/pvelesko/install/intel/neo/2023.10.02/lib/intel-opencl/libigdrcl.so
#3  0x00007f96488ab6a3 in NEO::BufferObject::wait(long) () from /home/pvelesko/install/intel/neo/2023.10.02/lib/intel-opencl/libigdrcl.so
#4  0x00007f96488aee48 in NEO::DrmGemCloseWorker::worker(void*) () from /home/pvelesko/install/intel/neo/2023.10.02/lib/intel-opencl/libigdrcl.so
#5  0x00007f9649865ac3 in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#6  0x00007f96498f7a40 in ?? () from /lib/x86_64-linux-gnu/libc.so.6

Thread 4 (Thread 0x7f9634eb5640 (LWP 176086)):
#0  0x00007f9649862117 in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x00007f9649864a41 in pthread_cond_wait () from /lib/x86_64-linux-gnu/libc.so.6
#2  0x00007f9648572250 in NEO::AsyncEventsHandler::asyncProcess(void*) () from /home/pvelesko/install/intel/neo/2023.10.02/lib/intel-opencl/libigdrcl.so
#3  0x00007f9649865ac3 in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#4  0x00007f96498f7a40 in ?? () from /lib/x86_64-linux-gnu/libc.so.6

Thread 3 (Thread 0x7f9649775600 (LWP 175972)):
#0  0x00007f9649862117 in ?? () from /lib/x86_64-linux-gnu/libc.so.6
[...]

PR #682, opened 2023-11-13
Refactor hipEventRecord

pvelesko:
* Invert record operation
Since event recording enqueues things onto a queue, invert the operation: event->record(queue) to queue->record(event). Now all enqueue operations are part of the queue class.

* Optimize event recording in Level Zero backend
Previous implementation has two global barriers which is already inefficient. This was addressed by using a dependency chain instead. Furthermore, there are a couple of tests that do a lot of event recording and when the number of parallel unit tests would get pushed up past ~8, the total runtime of these tests can sometimes explode causing event timeouts. It appears to me that sometimes these timeouts would result in the zombie processes we've observed which would then subsequently cause other tests to timeout (creating more of such processes). 

* Max out parallel unit tests
Tests now appear to run very consistently and I haven't observed any more hung kernel processes.


src/backend/Level0/CHIPBackendLevel0.cc
```diff
@@ -359,78 +363,52 @@ CHIPEventLevel0::CHIPEventLevel0(CHIPContextLevel0 *ChipCtx,
...
-    CHIPERR_LOG_AND_THROW("Queue passed in is null", hipErrorTbd);
+  ze_command_list_handle_t CommandList = ChipCtxLz_->getCmdListReg();
+
+  auto EventsToWaitOn = getSyncQueuesLastEvents();
+  auto EventToWaitOnHandles = getEventListHandles(EventsToWaitOn);
+
+  // create an Event for making a dependency chain
+  std::shared_ptr<chipstar::Event> TimestampWriteComplete =
+      static_cast<CHIPBackendLevel0 *>(Backend)->createCHIPEvent(
+          this->ChipContext_);
```
linehill, line 387: Casting does not seem to be needed.
```suggestion
  std::shared_ptr<chipstar::Event> TimestampWriteComplete =
      Backend->createCHIPEvent(this->ChipContext_);
```

src/backend/Level0/CHIPBackendLevel0.cc
```diff
@@ -476,7 +454,7 @@ bool CHIPEventLevel0::updateFinishStatus(bool ThrowErrorIfNotReady) {
 /** This Doesn't work right now due to Level Zero Backend hanging?
   unsinged long CHIPEventLevel0::getFinishTime() {
 
-    ze_kernel_timestamp_result_t res{};
+    ze_kernel_Timestampresult_t res{};
```
linehill, line 457: Is this change intended? Seems accidental.

src/backend/Level0/CHIPBackendLevel0.hh
```diff
@@ -255,11 +256,16 @@ protected:
   ze_command_list_desc_t CommandListDesc_;
   ze_command_queue_handle_t ZeCmdQ_;
   ze_command_list_handle_t ZeCmdListImm_;
-  std::mutex CommandListMtx_; /// prevent simultaneous access to ZeCmdListImm_
 
   void initializeCmdListImm();
 
 public:
+  virtual void recordEvent(chipstar::Event *ChipEvent) override;
```
linehill, line 263: FWIW, `override` implies `virtual`.
```suggestion
  void recordEvent(chipstar::Event *ChipEvent) override;
```

src/backend/OpenCL/CHIPBackendOpenCL.cc
```diff
@@ -597,13 +595,19 @@ CHIPBackendOpenCL::createCHIPEvent(chipstar::Context *ChipCtx,
...
-  return;
+  if (getLastEvent() == nullptr) {
+    std::shared_ptr<chipstar::Event> MarkerEvent = enqueueMarker();
+    ChipEventCL->takeOver(MarkerEvent);
+    ChipEventCL->setRecording();
+  } else {
+    std::shared_ptr<chipstar::Event> LastEvent = getLastEvent();
+    ChipEventCL->takeOver(LastEvent);
+    ChipEventCL->setRecording();
+  }
```
linehill, line 610: Could be expressed more simply.
```suggestion
  std::shared_ptr<chipstar::Event> LastEvent = getLastEvent();
  ChipEventCL->takeOver(LastEvent ? LastEvent : enqueueMarker());
  ChipEventCL->setRecording();
```

src/backend/Level0/CHIPBackendLevel0.cc
pvelesko: Indeed!

pvelesko:
@linehill  can you test on your end to see if you can reproduce OpenCL intel cpu failure on CI?

src/backend/Level0/CHIPBackendLevel0.cc
```diff
@@ -814,6 +792,19 @@ hipError_t CHIPKernelLevel0::getAttributes(hipFuncAttributes *Attr) {
 // CHIPQueueLevelZero
 // ***********************************************************************
 
+std::vector<ze_event_handle_t> CHIPQueueLevel0::getEventListHandles(
```
linehill, line 795: FWIW, this function could be a free function instead of a member function of CHIPQueueLevel0.

cmake/UnitTests.cmake
```diff
@@ -432,6 +431,7 @@ list(APPEND CPU_OPENCL_FAILED_TESTS "syncthreadsExitedThreads") # Timeout
 list(APPEND CPU_OPENCL_FAILED_TESTS "hipMultiThreadAddCallback") # SEGFAULT
 
 # iGPU OpenCL Unit Test Failures
+list(APPEND IGPU_OPENCL_FAILED_TESTS "Unit_hipMultiThreadStreams1_AsyncSame") # invalid free()
```
linehill, line 434: FYI, this seems to be a flaky case. On my Intel iGPU the test passes.

pvelesko: It's a race condition addressed in #657

src/backend/Level0/CHIPBackendLevel0.cc
pvelesko: I don't see this  in the actual code? This should cause a build failure

linehill:
> @linehill can you test on your end to see if you can reproduce OpenCL intel cpu failure on CI?

Do you mean the `Unit_hipMultiThreadStreams1_AsyncSame` case? It passes on Intel iGPU and Intel CPU through OpenCL BE.

src/backend/OpenCL/CHIPBackendOpenCL.cc
pvelesko: nice.

src/backend/Level0/CHIPBackendLevel0.cc
pvelesko: I prefer it being a method

PR #693, opened 2023-11-20
Add capability based HIP device library link 

linehill:
Add a mechanism to link in HIP device library implementations based on target capabilities quaried at runtime (rtdevlib). Using floating-point atomic add operations and their fall-backs as an use case.

Other changes:

* Add checks for atomicAdd() and atomicExch().

* OpenCL: Remove redundant OpenCL declarations.

* Use built-in float/double atomic_exchange*() without cl_ext_float_atomics (they don't require the extension).

* Don't give warnings about undefined symbols starting with `__chip_` which are given for symbols defined in rtdevlib.


linehill:
Todo:
- [ ] Investigate/fix issue occurring on POCL.

pjaaskel:
Although this is a bit intrusive, shall we consider this for 1.1 as it potentially helps portability?

pvelesko:
rebased on main

PR #706, opened 2023-11-27
Add sanity check for catching unexpected atomic built-ins

linehill:
AFAIK, all HIP atomic operations should map to corresponding OpenCL built-ins. This checker attempts to catch unexpected non-OpenCL atomic built-ins.


pvelesko:
Rebasing on main

pvelesko:
should this draft be merged? @linehill

linehill:
Sure, why not, this could be useful.

PR #718, opened 2023-12-01
OpenCL: Require SVM

pjaaskel:
Require CG SVM / USM from the OpenCL device. Without this, it lets such a device pass until the first SVM allocation, which the crashes uglily.

Also sneak in a formatter script stolen from PoCL.

### DIFF
--- a/src/backend/OpenCL/CHIPBackendOpenCL.cc
+++ b/src/backend/OpenCL/CHIPBackendOpenCL.cc
@@ -1579,36 +1579,53 @@ void CHIPBackendOpenCL::initializeImpl(std::string CHIPPlatformStr,
 
   StrStream.str("");
 
-  StrStream << "OpenCL Devices of type " << CHIPDeviceTypeStr
-            << " with SPIR-V_1 support:\n";
-  std::vector<cl::Device> SpirvDevices;
+  StrStream << "OpenCL Devices of type " << CHIPDeviceTypeStr << ":\n";
+
+  std::vector<cl::Device> SupportedDevices;
   std::vector<cl::Device> Dev;
   Err = SelectedPlatform.getDevices(SelectedDevType, &Dev);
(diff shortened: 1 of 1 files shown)


src/backend/OpenCL/CHIPBackendOpenCL.cc
```diff
@@ -1579,36 +1579,53 @@ void CHIPBackendOpenCL::initializeImpl(std::string CHIPPlatformStr,
...
-      StrStream << DeviceName << "\n";
-      SpirvDevices.push_back(D);
+
+    std::string DeviceName = D.getInfo<CL_DEVICE_NAME>();
+
+    StrStream << DeviceName << " ";
+    std::string SPIRVVer = D.getInfo<CL_DEVICE_IL_VERSION>(&Err);
+    if ((Err != CL_SUCCESS) ||
+        (SPIRVVer.rfind("SPIR-V_1.", 0) == std::string::npos)) {
+      StrStream << " no SPIR-V support." << std::endl;
```
linehill, line 1596: Should we avoid [std::endl](https://llvm.org/docs/CodingStandards.html#avoid-std-endl) even tough we are not dealing with an IO-stream for style consistency?

pjaaskel: I've understood a benefit (in addition to the flush) you get from std::endl is that it can also add a DOS/Windows line feed if running on Windows.

pjaaskel: https://stackoverflow.com/questions/1651936/changing-stdendl-to-put-out-crlf-instead-of-lf ok, \n char seems to get expanded to Windows linefeeds also. It shows that how much I've programmed on Windows ;) Anyhow, I don't think it matters if it flushes that stream here.

pjaaskel: OK. It was new info to me that the implementation is supposed to check the outputted chars and replace \n with \r\n for Windows.

Issue #713, opened 2023-11-30
Bad portability due to Intel extensions in generated SPIR-V binaries


linehill:
Currently HIP application compiled by chipStar don't port to non-Intel devices due to Intel extensions in SPIR-V binaries. This is known to be a problem with Mali GPU. An example from the VecAdd sample:
```
spirv-dis VecAdd-hip-spirv64-generic.out | grep -i intel
               OpCapability FPFastMathModeINTEL
               OpCapability MemoryAccessAliasingINTEL
               OpExtension "SPV_INTEL_memory_access_aliasing"
         %41 = OpAliasDomainDeclINTEL
         %42 = OpAliasScopeDeclINTEL %41
         %43 = OpAliasScopeListDeclINTEL %42
         %50 = OpAliasDomainDeclINTEL
         %51 = OpAliasScopeDeclINTEL %50
         %52 = OpAliasScopeListDeclINTEL %51
        %101 = OpAliasDomainDeclINTEL
        %102 = OpAliasScopeDeclINTEL %101
        %103 = OpAliasScopeListDeclINTEL %102
...
```

PR #521, opened 2023-06-24
Fix 513

pvelesko:
Fixes #513

PR #719, opened 2023-12-01
SPIR-V portability fix for single element vectors

linehill:
Adjust a HIP header for avoiding single element vectors to appear in LLVM IR. Single element vectors are not supported in SPIR-V without an extension (e.g. SPV_INTEL_vector_compute) which are enabled with hard coded --spirv-ext=+all option in clang. 

I'll make clang patch to disable all SPIR-V extension which will cause an assertion to be triggered in `llvm-spirv` if there are single element vectors in the input device code.

Related to issue #713.

(We had this patch previously but it got lost in #521)


pvelesko:
rebased on main

PR #721, opened 2023-12-05
Add mapping to OpenCL native_* on -ffast-math

linehill:
Map HIP device builtin functions to corresponding OpenCL native built-ins on -ffast-math. The rationale is based on `-ffast-math` specification [1] which gives permission to approximate transcendental functions (-fapprox-func).

[1] https://clang.llvm.org/docs/UsersManual.html#cmdoption-ffast-math


pvelesko:
Shouldn't we compile two versions of `devicelib.bc` - one with fast math and one without? Then link in the appropriate on hipcc?

linehill:
> Shouldn't we compile two versions of `devicelib.bc` - one with fast math and one without?

Sure, if there is a case where we can't select the fast-math version in the HIP headers.

We would probably need multiple devicelib versions for cases where other relaxed math options are used (e.g. `-fapprox-func`, `-ffinite-math-only`, `-fassociateve-math`, `fno-signed-zeros`, etc.). For those clang does not emit macros as it does for `-ffast-math`.

PR #729, opened 2023-12-18
Link __ballot() implementation at runtime

linehill:
Both the level zero and the OpenCL implementations may support SPIR-V counterpart of __ballot() below SPIR-V v1.3 with appropriate extension.

However, there does not seem to be a way to produce SPIR-V with ballot instructions on version v1.2 and below without using chipStar's fork of the llvm-spirv. To get around this limitation the ballot() is put into separate source, compiled to SPIR-V v1.3 and linked at runtime if the device supports it. Consequently, this patch should make the llvm-spirv fork unneeded.

### DIFF
--- a/src/backend/OpenCL/CHIPBackendOpenCL.cc
+++ b/src/backend/OpenCL/CHIPBackendOpenCL.cc
@@ -407,6 +407,10 @@ CHIPDeviceOpenCL::CHIPDeviceOpenCL(CHIPContextOpenCL *ChipCtx,
       *DevIn, CL_DEVICE_SINGLE_FP_ATOMIC_CAPABILITIES_EXT);
   Fp64AtomicAddCapabilities_ = getFPAtomicCapabilities(
       *DevIn, CL_DEVICE_DOUBLE_FP_ATOMIC_CAPABILITIES_EXT);
+
+  auto DevExts = DevIn->getInfo<CL_DEVICE_EXTENSIONS>();
+  HasSubgroupBallot_ =
+      DevExts.find("cl_khr_subgroup_ballot") != std::string::npos;
 }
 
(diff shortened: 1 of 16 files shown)


src/backend/OpenCL/CHIPBackendOpenCL.cc
```diff
@@ -407,6 +407,9 @@ CHIPDeviceOpenCL::CHIPDeviceOpenCL(CHIPContextOpenCL *ChipCtx,
       *DevIn, CL_DEVICE_SINGLE_FP_ATOMIC_CAPABILITIES_EXT);
   Fp64AtomicAddCapabilities_ = getFPAtomicCapabilities(
       *DevIn, CL_DEVICE_DOUBLE_FP_ATOMIC_CAPABILITIES_EXT);
+
+  auto DevExts = DevIn->getInfo<CL_DEVICE_EXTENSIONS>();
+  HasSubgroupBallot_ = DevExts.find("cl_khr_subgroup_ballot");
```
pjaaskel, line 412: I think the LLVM style doesn't add underscores for member variables, or does it?

linehill: Underscores in the names are probably discouraged in LLVM's coding standard. That said LLVM has them in some uncommon cases - for example:
```
  LazyValueInfo(AssumptionCache *AC_, const DataLayout *DL_,
                TargetLibraryInfo *TLI_)
      : AC(AC_), DL(DL_), TLI(TLI_) {}
```
It is Paulius' request that private and protected members have `_` suffix in the names. (There is also a [clang-tidy configuration](https://github.com/CHIP-SPV/chipStar/blob/d31f90490d6f9ff7ffc85979e792c4c6467c7895/.clang-tidy#L13-L16) for this too).

pjaaskel (approved):

PR #731, opened 2023-12-19
Interop sample fix for oneAPI-2024

Sarbojit2019:


pjaaskel:
Should we pull this in for 1.1? It doesn't seem to affect other parts but the interop.

pvelesko:
@pjaaskel I need to test locally first but yes I think so

pjaaskel (approved):

PR #737, opened 2023-12-21
level0: resolve #735

### LINKED ISSUE #735: sycl_interop fails: "Can't compile module. Level zero does not support multi-input compilation."
This is with the latest compute-runtime drivers and `intel-oneapi-dpcpp-cpp-2023.2.1`.

#731 doesn't help. @linehill "multi-input" rings the bell related to the SPIR-V device-feature-based linking. Could that cause issues like this?

linehill:
Fix hipInitFromNativeHandles() resetted properties in the backend which caused a linking issue seen in #735.

Resolves #735.


pjaaskel:
Confirmed the fix locally.

pjaaskel (approved):

PR #742, opened 2023-12-24
Add MKL back into testing for interop

pvelesko:
* CI now loads icpx and mkl in order to build HIP-SYCL Interop
* HIP-SYCL Interop samples now query env for whether immediate or regular command lists are used
* Fix FindLLVM - don't cache value if not found. You no longer need to rm CMakeCache.txt to reconfigure if llvm wasn't loaded


pvelesko:
@Sarbojit2019 Can you please put the fix for #743 here? thanks

PR #744, opened 2024-01-04
Fix bogus sizeof(sizeof(string)) expression in gcnArchName()

sobomax:
Fix bogus sizeof(sizeof(string)) expression, which results in gcnArchName() returning "unavailable" truncated to 8 characters, which is "unavaila".


linehill (approved):
LGTM. Thanks.

PR #747, opened 2024-01-10
Fixed Interop sample crash

Sarbojit2019:


pvelesko:
Implemented an env var check in #742

PR #752, opened 2024-01-13
Refactor WaitForThreadExit

pvelesko:
In the case that a user creates a bunch of threads and main() exits without calling join() on those threads, we must delay backend destruction so that those other threads can finish their work and don't cause a segfault. 

* Add a test
* Refactor using atomic operations in thread_local `chipstar::Queue` ctor/dtor


pjaaskel (changes requested):

src/CHIPBackend.cc
```diff
@@ -1189,30 +1189,26 @@ void chipstar::Backend::trackEvent(
...
 
-      logDebug("Backend::waitForThreadExit() per-thread queues still active "
-               "{}. Sleeping for 1s..",
-               NumPerThreadQueuesActive);
+    // logDebug and wait for 1 second
+    if (!AllThreadsExited) {
+      logWarn("Waiting for all per-thread queues to exit... This condition "
+              "would indicate that the main thread didn't call "
+              "join()");
+      sleep(1);
```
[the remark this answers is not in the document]
pvelesko: This PR is meant to do exactly this - main thread waits for an atomic counter to hit (which dets incremented/decremented by Queue ctor/dtor)

src/CHIPBackend.cc
```diff
@@ -1189,30 +1184,36 @@ void chipstar::Backend::trackEvent(
...
```
pjaaskel, line 1215: The sleep here feels wrong/hacky. Could it be a condition variable invoked when the device's ref count goes to zero?

pvelesko: That's what we do above - we have an atomic counter in the queue ctor/dtor.

The issue the main thread can exit before all the threads have reached the ctor. I agree that this is not optimal but I can't think of a better way to solve this. I tried creating another thread counter class and making it public thread_local hoping that as soon as a thread is created the constructor would get called but it still gets called later down in the callstack as indicated in the comment.

Perhaps there's a clang trick to force a certain object constructor to get execute before others? @pjaaskel

pvelesko: Sorry I thought you meant the first sleep statement but my question still stands.

As for this, this sleep is part of the look where we check the counter. I think we want this debug statement and if we remove the sleep here it will spam stdout. This section of the code is only run in the case that a user wrote a bad hip program where threads are not joined() explicitly.

pjaaskel: Yes, I just meant to have a cleaner approach for the wait instead of sleep-polling with a rather long sleep time - a (pthread) condition variable kind of approach. Not critical if you think it's good enough.

pvelesko:
Since sleep polling only happens in a very edge case - merging.

Issue #700, opened 2023-11-22
atomicAdd for unsigned long lost


pjaaskel:
hipCUB used to build when CHIP_ENABLE_NON_COMPLIANT_DEVICELIB_CODE` was defined, which brought in the atomicAdd() for longs (see 71b5c82fd0235c1) which is not defined in the HIP spec.

It's seems it's missing again:
```
/home/pjaaskel/src/hipCUB/build/deps/rocprim/include/rocprim/intrinsics/../intrinsics/atomic.hpp:51:16: error: no matching function for call to 'atomicAdd'
        return ::atomicAdd(address, value);
```

linehill:
The only atomic function I see the CHIP_ENABLE_NON_COMPLIANT_DEVICELIB_CODE covers is `atomicAdd(unsigned long*, unsigned long)` which is the undocumented one. However, the function seems to be callable with the recent chipStar.

```
$ cat atomicAddULong.hip
#include <hip/hip_runtime.h>
__device__ unsigned long x = 0;
__global__ void k() { atomicAdd(&x, 1); }
$ ../install/bin/hipcc -DCHIP_ENABLE_NON_COMPLIANT_DEVICELIB_CODE atomicAddULong.hip -c

$ echo $?
0
```

pjaaskel:
Seems I forgot that configuring chipStar doesn't then make hipcc automatically pass that macro (perhaps it should). Passing it separately moves the build forward, bumping to #701.

PR #763, opened 2024-01-26
Map device built-ins to compiler built-ins

linehill:
Map a set of HIP built-ins referenced by HeCBench benchmarks to compiler built-ins. The myocyte-hip benchmark seems to benefit from this the most - most likely due to:

* HIP built-in calls with compile time values which are constant-folded.

* HIP built-in calls which are simplified such as `pow(x, 2) --> x*x`.

### DIFF
--- a/bitcode/CMakeLists.txt
+++ b/bitcode/CMakeLists.txt
@@ -34,10 +34,6 @@ if(NOT CLANG_VERSION_LESS_15 AND CLANG_VERSION_LESS_16)
   list(APPEND EXTRA_FLAGS ${DISABLE_OPAQUE_PTRS_OPT})
 endif()
 
-if(CHIP_ENABLE_NON_COMPLIANT_DEVICELIB_CODE)
-  list(APPEND EXTRA_FLAGS "-DCHIP_ENABLE_NON_COMPLIANT_DEVICELIB_CODE=1")
-endif()
-
 if("${LLVM_VERSION}" VERSION_LESS 14.0)
   # Definitions for pre-upstreamed HIP-Clang.
   set(BC_TRIPLE "spirv64")
@@ -109,18 +105,15 @@ set(OCML_LIB_PATHS
   ${CMAKE_BINARY_DIR}/bitcode/ROCm-Device-Libs/oclc/oclc_isa_version_803.lib.bc
 )
 
-if(CHIP_ENABLE_NON_COMPLIANT_DEVICELIB_CODE)
-    # add support for calling certain LLVM builtins that are not supported by HIP/CUDA
-    add_custom_command(
-    OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/BC/c_to_opencl.bc"
-    DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/c_to_opencl.c"
-    COMMAND "${CMAKE_CXX_COMPILER}" ${BITCODE_C_COMPILE_FLAGS}
-    -o "${CMAKE_CURRENT_BINARY_DIR}/BC/c_to_opencl.bc"
-    -c "${CMAKE_CURRENT_SOURCE_DIR}/c_to_opencl.c"
-    COMMENT "Building c_to_opencl.bc"
-    VERBATIM)
-    list(APPEND DEPEND_LIST "${CMAKE_CURRENT_BINARY_DIR}/BC/c_to_opencl.bc")
-endif()
+add_custom_command(
+  OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/BC/c_to_opencl.bc"
+  DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/c_to_opencl.c"
+  COMMAND "${CMAKE_CXX_COMPILER}" ${BITCODE_C_COMPILE_FLAGS}
+  -o "${CMAKE_CURRENT_BINARY_DIR}/BC/c_to_opencl.bc"
+  -c "${CMAKE_CURRENT_SOURCE_DIR}/c_to_opencl.c"
+  COMMENT "Building c_to_opencl.bc"
+  VERBATIM)
+list(APPEND DEPEND_LIST "${CMAKE_CURRENT_BINARY_DIR}/BC/c_to_opencl.bc")
 
 # devicelib
 add_custom_command(
--- a/bitcode/c_to_opencl.c
+++ b/bitcode/c_to_opencl.c
@@ -22,12 +22,15 @@
 
 // See c_to_opencl.def for details.
 
-#define DEF_UNARY_FN_MAP(NAME_, TYPE_)                                         \
-  extern TYPE_ MAP_PREFIX##NAME_(TYPE_);                                       \
-  TYPE_ NAME_(TYPE_ x) { return MAP_PREFIX##NAME_(x); }
-#define DEF_BINARY_FN_MAP(NAME_, TYPE_)                                        \
-  extern TYPE_ MAP_PREFIX##NAME_(TYPE_, TYPE_);                                \
-  TYPE_ NAME_(TYPE_ x, TYPE_ y) { return MAP_PREFIX##NAME_(x, y); }
+#define DEF_UNARY_FN_MAP(FROM_FN_, TO_FN_, TYPE_)                              \
+  extern TYPE_ __chip_c2ocl_##FROM_FN_(TYPE_);                                 \
+  TYPE_ FROM_FN_(TYPE_ x) { return __chip_c2ocl_##FROM_FN_(x); }
+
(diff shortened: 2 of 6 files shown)


bitcode/CMakeLists.txt
```diff
@@ -109,18 +105,15 @@ set(OCML_LIB_PATHS
...
-    OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/BC/c_to_opencl.bc"
-    DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/c_to_opencl.c"
-    COMMAND "${CMAKE_CXX_COMPILER}" ${BITCODE_C_COMPILE_FLAGS}
-    -o "${CMAKE_CURRENT_BINARY_DIR}/BC/c_to_opencl.bc"
-    -c "${CMAKE_CURRENT_SOURCE_DIR}/c_to_opencl.c"
-    COMMENT "Building c_to_opencl.bc"
-    VERBATIM)
-    list(APPEND DEPEND_LIST "${CMAKE_CURRENT_BINARY_DIR}/BC/c_to_opencl.bc")
-endif()
+add_custom_command(
```
pvelesko, line 108: Why are we getting rid of the option? These are still non-compliant, right?

bitcode/devicelib.cl
```diff
@@ -847,14 +847,12 @@ EXPORT uint __chip_float_as_uint(float x) { return as_uint(x); }
...
-#ifdef CHIP_ENABLE_NON_COMPLIANT_DEVICELIB_CODE
-
 // See c_to_opencl.def for details.
-#define DEF_UNARY_FN_MAP(NAME_, TYPE_)                                         \
-  TYPE_ MAP_PREFIX##NAME_(TYPE_ x) { return NAME_(x); }
-#define DEF_BINARY_FN_MAP(NAME_, TYPE_)                                        \
-  TYPE_ MAP_PREFIX##NAME_(TYPE_ x, TYPE_ y) { return NAME_(x, y); }
+#define MAP_PREFIX __chip_c2ocl_
+#define DEF_UNARY_FN_MAP(FROM_FN_, TO_FN_, TYPE_)                              \
+  TYPE_ __chip_c2ocl_##FROM_FN_(TYPE_ x) { return TO_FN_(x); }
```
pvelesko, line 853: Meant to use MAP_PREFIX?

bitcode/devicelib.cl
```diff
@@ -847,14 +847,12 @@ EXPORT uint __chip_float_as_uint(float x) { return as_uint(x); }
...
+#define DEF_BINARY_FN_MAP(FROM_FN_, TO_FN_, TYPE_)                             \
+  TYPE_ __chip_c2ocl_##FROM_FN_(TYPE_ x, TYPE_ y) { return TO_FN_(x, y); }
 #include "c_to_opencl.def"
 #undef UNARY_FN
 #undef BINARY_FN
```
pvelesko, line 858: Can you remind me why we have this here? If we already have this in c_to_opencl.c which we compile and then link into the final .bc?

bitcode/c_to_opencl.c
```diff
@@ -22,12 +22,15 @@
...
-  TYPE_ NAME_(TYPE_ x) { return MAP_PREFIX##NAME_(x); }
-#define DEF_BINARY_FN_MAP(NAME_, TYPE_)                                        \
-  extern TYPE_ MAP_PREFIX##NAME_(TYPE_, TYPE_);                                \
-  TYPE_ NAME_(TYPE_ x, TYPE_ y) { return MAP_PREFIX##NAME_(x, y); }
+#define DEF_UNARY_FN_MAP(FROM_FN_, TO_FN_, TYPE_)                              \
+  extern TYPE_ __chip_c2ocl_##FROM_FN_(TYPE_);                                 \
+  TYPE_ FROM_FN_(TYPE_ x) { return __chip_c2ocl_##FROM_FN_(x); }
+
+#define DEF_BINARY_FN_MAP(FROM_FN_, TO_FN_, TYPE_)                             \
+  extern TYPE_ __chip_c2ocl_##FROM_FN_(TYPE_, TYPE_);                          \
```
pvelesko, line 30: Use MAP_PREFIX?

include/hip/devicelib/double_precision/dp_math.hh
```diff
@@ -25,28 +25,86 @@
 
 #include <hip/devicelib/macros.hh>
 
+#if defined __has_builtin && __has_builtin(__builtin_acos)
```
pvelesko, line 28: When would these be false?

pvelesko (commented):
some questions & small things

bitcode/c_to_opencl.def
```diff
@@ -38,30 +40,78 @@
...
+DEF_BINARY_FN_MAP(fmaxf, fmax, float)
+DEF_BINARY_FN_MAP(fmin, fmin, double)
+DEF_BINARY_FN_MAP(fminf, fmin, float)
+DEF_BINARY_FN_MAP(fmod, fmod, double)
+DEF_BINARY_FN_MAP(fmodf, fmod, float)
+DEF_BINARY_FN_MAP(hypot, hypot, double)
+DEF_BINARY_FN_MAP(hypotf, hypot, float)
+DEF_BINARY_FN_MAP(nextafter, nextafter, double)
 
 #undef MAPPING_PREFIX
```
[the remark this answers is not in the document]
linehill: MAP*_PREFIX mentions have been removed.

bitcode/devicelib.cl
linehill: Removed MAP_PREFIX (was left over).

bitcode/devicelib.cl
linehill: The reason is explained in the [c_to_opencl.def](https://github.com/CHIP-SPV/chipStar/blob/0dba21d5c36121b9f0323acb2319e2bbcb00fc83/bitcode/c_to_opencl.def#L23-L39).

bitcode/CMakeLists.txt
linehill: > Why are we getting rid of the option?

These are required for using compiler built-ins (see [c_to_opencl.def](https://github.com/CHIP-SPV/chipStar/blob/0dba21d5c36121b9f0323acb2319e2bbcb00fc83/bitcode/c_to_opencl.def#L23-L39) for explanation).

> These are still non-compliant, right?

Non-compliant respect to what?

include/hip/devicelib/double_precision/dp_math.hh
linehill: The guards are for just in case. It's unlikely that Clang would drop the built-ins, I think.

bitcode/CMakeLists.txt
pvelesko: So why not keep the option? In atomics.hh we still rely on this option:

```
#ifdef CHIP_ENABLE_NON_COMPLIANT_DEVICELIB_CODE
// At least rocPRIM tests call the unsigned long variant although it's not
// listed in the user manual. Annoyingly, size_t is typically defined as
// unsigned long.
// FIXME: We should check that unsigned long is 64bits for the host.
extern "C++" inline __device__ unsigned long atomicAdd(unsigned long *address,
                                                       unsigned long val) {
  return __chip_atomic_add_l(address, val);
}
#endif
```


> Non-compliant respect to what?

https://github.com/CHIP-SPV/chipStar/issues/700

"when CHIP_ENABLE_NON_COMPLIANT_DEVICELIB_CODE` was defined, which brought in the atomicAdd() for longs (see https://github.com/CHIP-SPV/chipStar/commit/71b5c82fd0235c12d0e20f2adeffa75eaa8eef12) which is not defined in the HIP spec."

linehill: If the concern here is that removing the option in the bitcode library would expose non-compliant functions to the users, that shouldn’t be happening because they are not declared in the HIP headers. Therefore, removing the CHIP_ENABLE_NON_COMPLIANT_DEVICELIB_CODE option check shouldn't have negative consequences, AFAICS.

bitcode/CMakeLists.txt
```diff
@@ -34,10 +34,6 @@ if(NOT CLANG_VERSION_LESS_15 AND CLANG_VERSION_LESS_16)
   list(APPEND EXTRA_FLAGS ${DISABLE_OPAQUE_PTRS_OPT})
 endif()
 
-if(CHIP_ENABLE_NON_COMPLIANT_DEVICELIB_CODE)
-  list(APPEND EXTRA_FLAGS "-DCHIP_ENABLE_NON_COMPLIANT_DEVICELIB_CODE=1")
-endif()
-
```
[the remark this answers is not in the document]
linehill: It was already broken and the issue filed in #700 hasn’t been fixed despite being marked resolved. The removed definition in question was only set for bitcode library compilation (now redundant) and it did not have an effect elsewhere.

pvelesko:
rebased

PR #776, opened 2024-02-18
default to Debug 

pvelesko:
Switch to CMake default Debug

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -263,7 +263,7 @@ jobs:
       - run: ${{ env.OPENCL_ENV }} clinfo
       - run: mkdir -p build
       - name: CMake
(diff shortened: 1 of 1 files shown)


pvelesko:
Switching to debug makes page locking fail as follows:
```
cuda-bandwidthTest: /home/runner/work/chipStar/chipStar/src/CHIPBindings.cc:2473: hipError_t hipHostMallocInternal(void **, size_t, unsigned int): Assertion `PageLockSuccess == 0 && "Failed to page lock memory"' failed.
```

Are reason for this to happen from PoCL/Ubuntu Runners perspective? @pjaaskel @Kerilk

pjaaskel:
I have no clue why -g would change anything related to page locking and how PoCL could have anything to do with it.

PR #785, opened 2024-02-26
OpenCL: Skip SVM pointer annotation if possible

linehill:
Calling clSetKernelExecInfo() for passing CL_KERNEL_EXEC_INFO_SVM_PTRS demonstrates notable overhead on PVC. For example, HeCBench/hybridsort-hip sees 4-16% slowdown.

This patch adds a rudimentary analysis for detecting whether device modules might have indirect buffer accesses. The OpenCL backend skips clSetKernelExecInfo() calls when it knows the module, the kernel belongs to, does not have any indirect accesses.

### DIFF
--- a/llvm_passes/HipIGBADetector.cpp
+++ b/llvm_passes/HipIGBADetector.cpp
@@ -14,7 +14,7 @@
 //
 // Where the result is one if the are potential IGBAs and otherwise zero.
 //
-// If there were be an IGPA in the module, there has to to be a load instruction
+// If there would be an IGPA in the module, there has to to be a load instruction
 // with a pointer operand which is either loaded from memory or crafted from an
 // integer (which OTOH is loaded from somewhere else). The analysis is very
 // simple and naive: we look for pointer load and inttoptr instructions in the
--- a/llvm_passes/HipIGBADetector.cpp
+++ b/llvm_passes/HipIGBADetector.cpp
@@ -1,0 +1,102 @@
+//===- HipIGBADetector.cpp --------------------------------------===//
+//
+// Part of the chipStar Project, under the Apache License v2.0 with LLVM
+// Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+// This pass analyzes module to find out if it has potential indirect global
+// buffer accesses (IGBA). The outcome of the analysis is stored in a magic
+// variable for the chipStar runtime:
+//
+//   uint8_t __chip_module_may_have_IGBAs = <result>;
+//
+// Where the result is one if the are potential IGBAs and otherwise zero.
+//
+// If there were be an IGPA in the module, there has to to be a load instruction
+// with a pointer operand which is either loaded from memory or crafted from an
+// integer (which OTOH is loaded from somewhere else). The analysis is very
+// simple and naive: we look for pointer load and inttoptr instructions in the
+// whole module. If we see any, we conclude there are potential IGBAs. Downsides
+// of this are that
+//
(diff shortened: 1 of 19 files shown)


llvm_passes/HipIGBADetector.cpp
```diff
@@ -0,0 +1,102 @@
...
+// Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+// This pass analyzes module to find out if it has potential indirect global
+// buffer accesses (IGBA). The outcome of the analysis is stored in a magic
+// variable for the chipStar runtime:
+//
+//   uint8_t __chip_module_may_have_IGBAs = <result>;
```
pjaaskel, line 13: Perhaps use __chipStar_ as the prefix. 'chip' is misleading.

llvm_passes/HipIGBADetector.cpp
```diff
@@ -0,0 +1,102 @@
...
+//
+// Where the result is one if the are potential IGBAs and otherwise zero.
+//
+// If there were be an IGPA in the module, there has to to be a load instruction
```
pjaaskel, line 17:
```suggestion
// If there would be an IGPA in the module, there has to to be a load instruction
```

tests/runtime/TestNegativeIGBADetection.hip
```diff
@@ -0,0 +1,14 @@
+// Check we detect that the device code does not have indirect global
```
pjaaskel, line 1: Pls add the copyright/license texts to new files.

pjaaskel (changes requested):
Looks quite. Small things.

llvm_passes/HipIGBADetector.cpp
linehill: We are using `__chip_` prefix in a lot of places so I think I'll keep it for style consistency.

tests/runtime/TestNegativeIGBADetection.hip
linehill: Done. Is it fine to use "short-form" copyright text? And was it the MIT license the chipStar is using?

pvelesko:
I see some failures that seem unrelated to this PR, let me investigate.

src/spirv.cc
```diff
@@ -737,6 +745,14 @@ class SPIRVmodule {
             SpillAnnotation.push_back(std::make_pair(ArgIndex, ArgSize));
           }
         }
+
+        // A magic variable created by HipIGBADetector.cpp.
+        if (Name == "__chip_module_may_have_IGBAs") {
```
pjaaskel, line 750: Would it be safer to reverse the logic of the flag (also for backwards compatibility)? That is, add the magic variable only if known that there are no IGBAs in the module.

pjaaskel (changes requested):

src/spirv.cc
linehill: SGTM. Updated.

pjaaskel (approved):

tests/runtime/TestNegativeIGBADetection.hip
linehill: The existing LICENSE file in project root seems to match to the MIT license.

pvelesko (changes requested):
merge conflicts

linehill:
Rebased now.

pvelesko:
Only this is failing now
```
dgpu_opencl_make_check_result.txt: FAIL
	358 - Unit_hipClassKernel_Friend (SEGFAULT)
```

This is fixed in #791 so let's merge that first

pvelesko:
conflits

linehill:
Rebased. Should have compiler error resolved seen in the CI in the merge prior this rebase.

PR #791, opened 2024-03-01
OpenCL BE: set CHIP_USE_INTEL_USM on by default

linehill:
PVC platforms seems to benefit a lot from the USM. Few picks:

* HeCBench/mandelbrot: 56x speed-up.

* HeCBench/floydwarshall: 14x speed-up.

* HeCBench/bitonic-sort: 33x speed-up.

* HeCBench/nw: 20x speed-up.

* HeCBench/fft: 11x speed-up.

* And bunch of other HeCBench cases with speed-ups between 1.16x and 6.5x.

The OpenCL backend still falls back to use SVM (the previous default) when the USM is not available.


pvelesko:
@linehill  still failing `hipKernelLaunchIsNonBlocking`

linehill:
Added work-around to a test timeout issue.

PR #793, opened 2024-03-04
Fix HIP float intrinsics were mapped double built-ins

linehill:
The [use-compiler-builtins](https://github.com/CHIP-SPV/chipStar/pull/763) patch unintentionally redirected HIP single precision intrinsic calls to double precision compiler/OpenCL built-ins.

### DIFF
--- a/include/hip/devicelib/single_precision/sp_intrinsics.hh
+++ b/include/hip/devicelib/single_precision/sp_intrinsics.hh
@@ -41,6 +41,7 @@
  */
 
 extern "C++" __device__ float native_cos(float x); // OpenCL
+extern "C++" __device__ float cos(float x);        // OpenCL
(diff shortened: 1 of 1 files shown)


pvelesko:
Linter error + Question: So the opencl intrinsics are declared here, shouldn't the mappping also change? Or was the mapping done in the previous PR without declaration which would have resulted in compilation issues?

If it's the latter we need to adjust testing

pvelesko (changes requested):
lint + question

linehill:
Lint fix.

> So the opencl intrinsics are declared here, shouldn't the mappping also change? Or was the mapping done in the previous PR without declaration which would have resulted in compilation issues?

This patch brings the mapping to the state it was before #763 PR.

PR #800, opened 2024-03-12
Add CHIP_DEVICE_TYPE to documentation

karlwessel:
This add the following two points to the documentation.
1. Document existance and usage of the `CHIP_DEVICE_TYPE` environment variable.
2. Adds one more note regarding availability of trace and debug log level information.

# Motivation
Thanks for this great library, works great!
I had to spent some hours at the beginning tracking down the root for the following two error messages using OpenCL backend: `Selected OpenCL device 0 is out of range`.
It took some time to find the undocumented environment variable which caused chipStar to only look for GPUs on my CPU-only system.


pvelesko:
Thank you for the contribution! @karlwessel

PR #816, opened 2024-03-27
Various small optimizations

linehill:
... for reducing kernel launch host overhead.

* Enable (optional) link-time optimizations for the libCHIP library.
* Add early exits in couple functions.
* Avoid redundant kernel argument copies. Along the way, fix CHIPGraphNodeKernel instances didn't copy kernel arguments fully (they only copied pointers to arguments but not their values).
* Eliminate map lookups in SPVFuncInfo::visit*Args().


pvelesko:
A test failed with what appears to be a race condition.

pvelesko (approved):
LGTM

PR #826, opened 2024-04-09
Add SKIP_TESTS_WITH_DOUBLES Option

pvelesko:


pjaaskel:
This would be better as a check.py option as it's possible to run the same build on different devices, some of them supporting FP64, some not.

pvelesko:
@pjaaskel I couldn't find another way to do this. Tried using `CTEST_PRE_TEST` but it didn't do anything. 

Since `check.py` uses ctest as well, there's no simple way of doing this. In either case, just toggling the option on or off and reconfiguring cmake is enough here - you don't have to actually recompile the tests.

pvelesko:
Merging this now now, we can implement a method that doesn't require CMake reconfigure later.

pjaaskel:
I meant that we'd restrict the list of XPASS tests (which is currently restricted only by those cpu/gpu cl/lz toggles) by checking from clinfo if the target supports doubles, and/or have a flag in check.py that drops the tests that are known to fail with doubles.

pvelesko:
> XPASS tests (which is currently restricted only by those cpu/gpu cl/lz toggles

I'm sorry what is XPASS? And what toggles are you talking about?

In any case, how would we know if the test is using doubles without inspecting the SPIR-V?

pjaaskel:
Sorry for being unclear - not enough coffee consumed yet. Toggles in check.py which limit the list of tests expected to pass. You could inspect SPIR-V as you do know and dump the test cases that are expected to fail? But yep - this is not important.

pvelesko:
> Toggles in check.py which limit the list of tests expected to pass.

I'll add this approach to my backlog. Thank you for the suggestion.

PR #844, opened 2024-05-05
Small Fixes

pvelesko:
* Add a test which runs valgrind to check that no additional thread safety issues are introduced
* CMake fixes enabling to recompile the code with modules unloaded
* CMake fixes to prevent repeated building of hipInfo and ccompat
* Various other small fixes


linehill (approved):
LGTM, thanks.

PR #845, opened 2024-05-05
Level Zero - Fix OOM & Improve Thread Safety

pvelesko:
* Fix the issue where not all events were being recycled by Level Zero event collector
* Refactor `CHIPEventLevel0::wait()` to get rid of race conditions reported by valgrind. The API states that calls to `zeEventHostSynchronize` and `zeEventQuery` are thread safe but valgrind reports race conditions.
* Implement a global shared mutex - `ApiMtx` which is to be locked by every HIP API call. This prevents multiple HIP commands from executing at the same time. Pretty coarse lock which we can relax over time since performance is affected only for multithreaded HIP applications of which I haven't seen any yet.

### DIFF
--- a/src/backend/Level0/CHIPBackendLevel0.cc
+++ b/src/backend/Level0/CHIPBackendLevel0.cc
@@ -421,23 +421,20 @@ bool CHIPEventLevel0::wait() {
 bool CHIPEventLevel0::updateFinishStatus(bool ThrowErrorIfNotReady) {
   isDeletedSanityCheck();
   std::string EventStatusOld, EventStatusNew;
-  {
-    LOCK(EventMtx); // chipstar::Event::EventStatus_
 
-    EventStatusOld = getEventStatusStr();
+  EventStatusOld = getEventStatusStr();
 
-    ze_result_t Status = zeEventQueryStatus(Event_);
-    if (Status == ZE_RESULT_NOT_READY && ThrowErrorIfNotReady) {
-      CHIPERR_LOG_AND_THROW("chipstar::Event Not Ready", hipErrorNotReady);
-    }
-    if (Status == ZE_RESULT_SUCCESS) {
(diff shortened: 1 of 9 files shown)


src/backend/Level0/CHIPBackendLevel0.hh
```diff
@@ -179,7 +178,7 @@ private:
   CHIPContextLevel0 *Ctx_;
   ze_event_pool_handle_t EventPool_;
   unsigned int Size_;
-  std::stack<std::shared_ptr<CHIPEventLevel0>> Events_;
+  std::stack<CHIPEventLevel0 *> Events_;
```
linehill, line 181: Doesn't seem like the CHIPEventLevel0 objects are destructed along with LZEventPool. Relevant part from the destructor:
```c++
  while (Events_.size())
    Events_.pop();
```

src/backend/Level0/CHIPBackendLevel0.cc
```diff
@@ -1478,10 +1465,13 @@ void CHIPQueueLevel0::finish() {
...
-                                        ChipEnvVars.getL0EventTimeout());
-  CHIPERR_CHECK_LOG_AND_ABORT(Status, ZE_RESULT_SUCCESS, hipErrorTbd,
-                              "zeCommandListHostSynchronize timeout out");
+  auto LastEvent = getLastEvent();
+  if (LastEvent)
+    LastEvent->wait();
+  // Status = zeCommandListHostSynchronize(ZeCmdListImm_,
+  //                                       ChipEnvVars.getL0EventTimeout());
+  // CHIPERR_CHECK_LOG_AND_ABORT(Status, ZE_RESULT_SUCCESS, hipErrorTbd,
+  //                             "zeCommandListHostSynchronize timeout out");
```
linehill, line 1474: Nit: dead code. Perhaps, remove it?

PR #873, opened 2024-06-17
wrap vector operator overloading with preprocessor directives to prev…

jjennychen:
…ent double definitions

Note: Added preprocessor directives to the operator overloadings inside and outside of struct; indented wrapped code to fit coding style.


pvelesko:
@jjennychen 

Cleaned up formatting to reduce the diff & incorporated this into #875

PR #876, opened 2024-06-20
Fix math function j1 typo in dp_math.hh

jjennychen:
Fixed the typo in double-precision math function j1 so that it calls __ocml_j1_f64 instead of __ocml_j0_f64.


pvelesko (approved):

pvelesko:
Nice catch!

PR #960, opened 2025-01-07
Don't split printf if %s is static

lights0123:
This is part of my work on [HipScript](https://hipscript.lights0123.com/), where I use this project to run HIP code in the browser with WebGPU.

Using `printf` with the `%s` specifier, like in `__assert_fail`, causes chipStar to split the call into several. For example:

```c
printf("File %s errored\n", "main.hip");
// turns into
printf("File ");
printf("main.hip");
printf(" errored\n");
```

This causes problems if multiple threads are printing, since these calls can interleave, producing output like:
```
File File main.hipmain.hip errored
errored

```
This PR will keep `%s` specifiers, but only if the argument can be converted to a global string. In my case, where I compile to Vulkan afterwards with [Clspv](https://github.com/google/clspv/), this allows the compiler to convert the string pointer to an integer ID.

### DIFF
--- a/llvm_passes/HipPrintf.cpp
+++ b/llvm_passes/HipPrintf.cpp
@@ -380,9 +380,11 @@ PreservedAnalyses HipPrintfToOpenCLPrintfPass::run(Module &Mod,
         // Skip the original format string arg and recreate it.
         OrigCallArgI++;
 
+        std::string toAdd;
+        std::vector<Value *> Args;
+
         for (auto FmtStr : FmtSpecPieces) {
           unsigned FormatSpecCount = NumFormatSpecs(FmtStr);
-          std::vector<Value *> Args;
 
           // TODO: handle a (compile time known) null ptr format string arg and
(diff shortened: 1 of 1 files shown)


llvm_passes/HipPrintf.cpp
```diff
@@ -380,9 +380,11 @@ PreservedAnalyses HipPrintfToOpenCLPrintfPass::run(Module &Mod,
         // Skip the original format string arg and recreate it.
         OrigCallArgI++;
 
+        std::string toAdd;
```
pjaaskel, line 383:
```suggestion
        std::string ToAdd;
```

pjaaskel: Also pls add a comment what is this variable for.

pjaaskel (changes requested):
Thanks. The improvement makes sense, but please check my feedback.

pvelesko:
printf unit tests failing fyi

PR #963, opened 2025-01-20
Prevent InvalidBitWidth Errors

pvelesko:
During optimization of loops, LLVM generates non-standard integer types such as i33 or i56. 

Tried to find which LLVM pass does this but it seems to happen quite early.

1. Upgrade non-standard sizes
2. Remove redundant trunc instructions


pjaaskel:
Related: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/1685

llvm_passes/HipPromoteInts.cpp
```diff
@@ -0,0 +1,115 @@
...
+            LLVM_DEBUG(dbgs() << "[HipPromoteInts] Found non-standard type in result: " << *I << "\n");
+            
+            unsigned NextStdSize = getPromotedBitWidth(IntTy->getBitWidth());
+            Type *PromotedType = Type::getIntNTy(M.getContext(), NextStdSize);
+            
+            LLVM_DEBUG(dbgs() << "[HipPromoteInts] Promoting from i" << IntTy->getBitWidth() 
+                      << " to i" << NextStdSize << "\n");
+            
+            // Update the instruction type
+            I->mutateType(PromotedType);
```
linehill, line 71: "Note that this is an extremely dangerous operation which can create completely invalid IR very easily. It is strongly recommended that you recreate IR objects with the right types instead of mutating them in place." [[source]](https://llvm.org/doxygen/classllvm_1_1Value.html#ac0f09c2c9951158f9eecfaf7068d7b20)

Please, do what the LLVM documentation suggests.

linehill (changes requested):

llvm_passes/HipPromoteInts.cpp
```diff
@@ -0,0 +1,298 @@
...
+                NewArg = Builder.CreateTrunc(NewArg, Call->getArgOperand(i)->getType());
+                
+            NewArgs.push_back(NewArg);
+        }
+        
+        CallInst* NewCall = CallInst::Create(Call->getFunctionType(),
+                                           Call->getCalledOperand(),
+                                           NewArgs,
+                                           Call->getName(),
+                                           Call);
```
linehill, line 197: This will create broken code as the called function is not promoted and still have non-standard integer parameters.

llvm_passes/HipPromoteInts.cpp
```diff
@@ -0,0 +1,298 @@
...
+        if (PromotedValues.count(V))
+            return PromotedValues[V];
+        
+        // If it's already the right type, return it
+        if (V->getType() == PromotedTy)
+            return V;
+            
+        // If it's the non-standard type, promote it
+        if (V->getType() == NonStdType) {
+            auto NewV = Builder.CreateZExt(V, PromotedTy);
```
linehill, line 77: This looks odd. If I see correctly, this creates an instruction with non-standard integer - but why?

llvm_passes/HipPromoteInts.cpp
```diff
@@ -0,0 +1,298 @@
...
+            Value* IncomingValue = Phi->getIncomingValue(i);
+            BasicBlock* IncomingBlock = Phi->getIncomingBlock(i);
+            
+            // If the incoming value is from our promotion chain, use the promoted value
+            Value* NewIncomingValue = PromotedValues.count(IncomingValue) ? 
+                                    PromotedValues[IncomingValue] : IncomingValue;
+            
+            // If the incoming value isn't promoted yet, promote it now
+            if (NewIncomingValue->getType() != PromotedType) {
+                NewIncomingValue = Builder.CreateZExt(NewIncomingValue, PromotedType);
```
linehill, line 103: I think this potentially inserts the new instruction before another PHI node which is ilegal.

llvm_passes/HipPromoteInts.cpp
```diff
@@ -0,0 +1,298 @@
...
+                                           Call);
+        NewCall->setCallingConv(Call->getCallingConv());
+        NewCall->setAttributes(Call->getAttributes());
+        
+        errs() << Indent << "  " << *I << "    ============> " << *NewCall << "\n";
+        PromotedValues[I] = NewCall;
+        Replacements.push_back(Replacement(I, NewCall));
+    }
+    else {
+        errs() << Indent << " Skipping: " << *I << "\n";
```
linehill, line 206: Put an assert here to alarm unhandled promotion cases.

llvm_passes/HipPromoteInts.cpp
pvelesko: Can you provide an example IR where this would happen?

llvm_passes/HipPromoteInts.cpp
pvelesko: We're using `PromotedTy` not `NonStd`:

```
1274:       getPromotedValue for: i33 4294967295
1274:         Promoting non-standard type: i33 4294967295 to i64 4294967295
```

linehill: The built instruction has an operand with non-standard integer type of `NonStdType` which makes it one this pass is supposed to eliminate.

pvelesko: That's not what I am seeing in the log provided, the resulting type is `i64`

llvm_passes/HipPromoteInts.cpp
linehill: If a call happens to be promoted, you would end up creating call with promoted argument list to a non-promoted function. Something like:
```
target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
target triple = "spirv64"

define spir_func void @bar_new(i32 noundef %x) local_unnamed_addr #0 {
entry:
  tail call spir_func void @foo(i32 noundef %x) #2
  ret void
}

declare spir_func void @foo(i42 noundef) local_unnamed_addr #1
```
Interestingly, LLVM's verifier does not complain about this, but this will fail in llvm-spirv.

llvm_passes/HipPromoteInts.cpp
linehill:
```diff
--- a/llvm_passes/HipPromoteInts.cpp
+++ b/llvm_passes/HipPromoteInts.cpp
@@ -75,6 +75,7 @@ void processInstruction(Instruction *I, Type *NonStdType, Type *PromotedTy, std:
         // If it's the non-standard type, promote it
         if (V->getType() == NonStdType) {
             auto NewV = Builder.CreateZExt(V, PromotedTy);
+            dbgs() << "XXX NewV = " << *NewV << "\n";
             PromotedValues[V] = NewV;
             return NewV;
         }
```
Compiling a source with the `testWarpCalc` kernel:
```
 ../install/bin/hipcc bitwidth-issue.hip -O2 -c
    %2 = zext i32 %0 to i33    ============>   %1 = zext i32 %0 to i64
XXX NewV =   %5 = zext i33 %4 to i64
      %7 = mul i33 %4, %2    ============>   %6 = mul i64 %5, %1
(...)
```
Here is an example of an instruction with non-standard integer (the operand) being created.

linehill (changes requested):
Please, run git-clang-format on the changes.

llvm_passes/HipPromoteInts.cpp
pvelesko: it's an intermediary instruction which I had to generate because LLVM seems to be very sensitive about changing instructions with different types. You probably understand this better but me but essentially when doing instruction replacement, their types must match. For this to be the case, I have to recreate an entire replacement chain first and only then do substitutions. To keep replacement chain valid, I have to deal with non-std types and promote them.

In the final step the replacement is done and we are left with only std types:
image


You can validate this yourself by running the test: 
`ctest -R hipcc-TestHipccInvalidBitWidth -V`

If this is not the correct way to do type promotion, please suggest a fix.

llvm_passes/HipPromoteInts.cpp
pvelesko: Same issue as above, I believe

llvm_passes/HipPromoteInts.cpp
pvelesko: This is also a part of the replacement chain. Can you provide an example where we are left with non-std types after the replacement is complete? Or suggest a way to perform the replacemet without intermediary values?

llvm_passes/HipPromoteInts.cpp
linehill: Maybe. If it is a placeholder value that will get deleted later in this pass, then its fine. If it sticks, you may end up with something like:
```
%x = phi ...
%y = zext ...
%z = phi ...
```
which is ill-formed. All PHI instructions must be come first before other instructions.

llvm_passes/HipPromoteInts.cpp
pvelesko: > replace the illegal operands with placeholders instructions (something like freeze(poison)) and keep track of them,
and after you have done constructing all the replacement instructions, replace all the placeholders with the actual replacement instructions.

Currently, I do this:
1. Create a replacement instruction
2. Track what it will have to replace via `Replacements` map
3. Perform all replacements

You are suggesting:
1. Modify the original instruction to use placeholder instruction like `freeze(poison)` 
2. Create a replacement instruction
3. Keep track of where these placeholders were used
4. Perform replacement for instructions and operands? 

this sounds to me more complex with the added benefit of being.. not having `NonStd` argument types in our intermediary IR (which has no effect after the replacements are complete?)


> You are kind of doing this in here minus the tracking and placeholder replacement part.

Is this not tracking?
```
Replacements.push_back(Replacement(I, NewInst));
```
And is this not replacement? 

```
                    for (const auto &R : Replacements) {
                        for (auto &U : R.Old->uses()) {
                            User *User = U.getUser();
                            if (!GlobalVisited.count(cast<Instruction>(User))) {
                                U.set(R.New);
                            }
                        }
                    }
```

pvelesko:
@linehill I spent a fair amount of looking already but to confirm: there is no way to instruct LLVM to NOT generate illegal integer types?

llvm_passes/HipPromoteInts.cpp
linehill: It wasn’t obvious how the replacements (of intermediate `zext` instruction) are carried after simulating a case through the algorithm. The above snippets aren't enough to explain it fully - the missing crucial part is the intermediate `zext` created here which sets up a def-use chain so that the instructions gets visited and processed later by `promoteChain()` and recorded into `Replacements`. 

So perhaps document how the algorithm works and leave a comment here how intermediate `zext` plays a role in it?

llvm_passes/HipPromoteInts.cpp
```diff
@@ -0,0 +1,309 @@
...
+    return TypeToPromote;  // Return original type if not an integer
+}
+
+struct Replacement {
+    Instruction* Old;
+    Value* New;
+    Replacement(Instruction* O, Value* N) : Old(O), New(N) {}
+};
+
+void processInstruction(Instruction *I, Type *NonStdType, Type *PromotedTy, std::string Indent,
```
linehill, line 60: Avoid a string copy?
```suggestion
void processInstruction(Instruction *I, Type *NonStdType, Type *PromotedTy, const std::string &Indent,
```

llvm_passes/HipPromoteInts.cpp
```diff
@@ -0,0 +1,309 @@
...
+        errs() << Indent << "  " << *I << "    ============> " << *NewCall << "\n";
+        PromotedValues[I] = NewCall;
+        Replacements.push_back(Replacement(I, NewCall));
+    }
+    else {
+        assert(false && "HipPromoteIntsPass: Unhandled instruction type");
+    }
+}
+
+bool promoteChain(Instruction *OldI, Type *NonStdType, Type *PromotedTy,
```
linehill, line 221: No caller observers the return value so might as change it to void. Also, this function does not need to be exported.
```suggestion
static void promoteChain(Instruction *OldI, Type *NonStdType, Type *PromotedTy,
```

llvm_passes/HipPromoteInts.cpp
```diff
@@ -0,0 +1,309 @@
...
+                      SmallPtrSetImpl<Instruction*> &Visited,
+                      SmallVectorImpl<Replacement> &Replacements,
+                      SmallDenseMap<Value*, Value*> &PromotedValues,
+                      unsigned Depth = 0) {
+    // If we've already processed this instruction, just return
+    if (!Visited.insert(OldI).second) {
+        // If we have a promoted value for this instruction, use it
+        if (PromotedValues.count(OldI)) {
+            errs() << std::string(Depth * 2, ' ') << "Already processed: " << *OldI << "\n";
```
linehill, line 230: This produces spam so perhaps remove it or warp it in LLVM_DEBUG. `dbgs()` is more appropriate for printing debug messages.

tests/compiler/TestHipccInvalidBitWidth.cpp
```diff
@@ -0,0 +1,59 @@
...
+    testWarpCalc<<<grid, block>>>(d_debug);
+    hipDeviceSynchronize();
+    
+    // Copy results back
+    hipMemcpy(h_debug, d_debug, numThreads * sizeof(int), hipMemcpyDeviceToHost);
+    
+    printf("Results for first few threads:\n");
+    printf("GlobalIdx\tValue\n");
+    
+    // Print first few entries
```
linehill, line 49: The result should be verified to detect regressions.

depChain.txt
linehill, line 1: Accidental file addition?

llvm_passes/HipPromoteInts.cpp
pvelesko: These are temporary - btw, how do you get  LLVM_DEBUG. dbgs() to print?

depChain.txt
pvelesko: WIP

linehill:
> @linehill I spent a fair amount of looking already but to confirm: there is no way to instruct LLVM to NOT generate illegal integer types?

Did you look into the pass responsible for spawning the non-standard integers? It might have a command line option for it.

pvelesko:
> Did you look into the pass responsible for spawning the non-standard integers? It might have a command line option for it.

I did but I can't find my notes on which one it was but iirc it was quite early and deals with general loop optimizations so disabling is not desired.

linehill:
> I did but I can't find my notes on which one it was but iirc it was quite early and deals with general loop optimizations so disabling is not desired.

When you find it again, check if it has an option that could stop it to generate non-standard integers - not to disable the whole pass.

pvelesko:
> > I did but I can't find my notes on which one it was but iirc it was quite early and deals with general loop optimizations so disabling is not desired.
> 
> When you find it again, check if it has an option that could stop it to generate non-standard integers - not to disable the whole pass.

It's `IndVarSimplifyPass` - none of the options I tried seem to help

llvm_passes/HipPromoteInts.cpp
linehill: By passing `-debug` to `opt` or `llc` tool, or `-mllvm -debug` to `clang` tool.

llvm_passes/HipPromoteInts.cpp
```diff
@@ -0,0 +1,377 @@
...
+      // If the incoming value is from our promotion chain, use the promoted
+      // value
+      Value *NewIncomingValue = PromotedValues.count(IncomingValue)
+                                    ? PromotedValues[IncomingValue]
+                                    : IncomingValue;
+
+      // If the incoming value isn't promoted yet, promote it now
+      if (NewIncomingValue->getType() != PromotedType) {
+        NewIncomingValue = Builder.CreateZExt(NewIncomingValue, PromotedType);
+      }
```
linehill, line 166: LLVM style nit:
```suggestion
      if (NewIncomingValue->getType() != PromotedType)
        NewIncomingValue = Builder.CreateZExt(NewIncomingValue, PromotedType);
```

llvm_passes/HipPromoteInts.cpp
```diff
@@ -0,0 +1,377 @@
...
+  } else if (isa<TruncInst>(I)) {
+    TruncInst *TruncI = cast<TruncInst>(I);
+    Value *SrcOp = TruncI->getOperand(0);
+    Value *PromotedSrc =
+        PromotedValues.count(SrcOp) ? PromotedValues[SrcOp] : SrcOp;
+
+    // Verify the source is actually of our promoted type
+    if (PromotedSrc->getType() != PromotedTy) {
+      PromotedSrc = Builder.CreateZExt(PromotedSrc, PromotedTy);
+    }
```
linehill, line 211: LLVM style nit.
```suggestion
    if (PromotedSrc->getType() != PromotedTy)
      PromotedSrc = Builder.CreateZExt(PromotedSrc, PromotedTy);
```

llvm_passes/HipPromoteInts.cpp
```diff
@@ -0,0 +1,377 @@
...
+  processInstruction(OldI, NonStdType, PromotedTy, Indent, Replacements,
+                     PromotedValues);
+
+  // Recursively process all users
+  for (User *U : OldI->users()) {
+    if (auto *UI = dyn_cast<Instruction>(U)) {
+      promoteChain(UI, NonStdType, PromotedTy, Visited, Replacements,
+                   PromotedValues, Depth + 1);
+    }
+  }
```
linehill, line 310: LLVM style nit:
```suggestion
  for (User *U : OldI->users())
    if (auto *UI = dyn_cast<Instruction>(U))
      promoteChain(UI, NonStdType, PromotedTy, Visited, Replacements,
                   PromotedValues, Depth + 1);
```

llvm_passes/HipPromoteInts.cpp
```diff
@@ -0,0 +1,377 @@
...
+    // First collect all instructions we need to promote
+    for (BasicBlock &BB : F) {
+      for (Instruction &I : BB) {
+        if (auto *IntTy = dyn_cast<IntegerType>(I.getType())) {
+          if (!isStandardBitWidth(IntTy->getBitWidth())) {
+            WorkList.push_back(&I);
+          }
+        }
+      }
+    }
```
linehill, line 333: LLVM style nit:
```suggestion
    for (BasicBlock &BB : F)
      for (Instruction &I : BB)
        if (auto *IntTy = dyn_cast<IntegerType>(I.getType()))
          if (!isStandardBitWidth(IntTy->getBitWidth()))
            WorkList.push_back(&I);
```

llvm_passes/HipPromoteInts.cpp
```diff
@@ -0,0 +1,377 @@
...
+          promoteChain(I, IntTy, PromotedType, GlobalVisited, Replacements,
+                       PromotedValues, 0);
+
+          // Update uses and cleanup as before
+          for (const auto &R : Replacements) {
+            for (auto &U : R.Old->uses()) {
+              User *User = U.getUser();
+              if (!GlobalVisited.count(cast<Instruction>(User))) {
+                U.set(R.New);
+              }
```
linehill, line 361: LLVM style nit:
```suggestion
              if (!GlobalVisited.count(cast<Instruction>(User)))
                U.set(R.New);
```

llvm_passes/HipPromoteInts.cpp
```diff
@@ -0,0 +1,377 @@
...
+                         unsigned Depth = 0) {
+  // If we've already processed this instruction, just return
+  if (!Visited.insert(OldI).second) {
+    // If we have a promoted value for this instruction, use it
+    if (PromotedValues.count(OldI)) {
+      LLVM_DEBUG(dbgs() << std::string(Depth * 2, ' ')
+                        << "Already processed: " << *OldI << "\n");
+      return;
+    }
+    return;
```
linehill, line 295: Could be simplified:
```suggestion
    if (PromotedValues.count(OldI))
      LLVM_DEBUG(dbgs() << std::string(Depth * 2, ' ')
                        << "Already processed: " << *OldI << "\n");
    return;
```

llvm_passes/HipPromoteInts.cpp
```diff
@@ -0,0 +1,377 @@
...
+ * During optimization of loops, LLVM generates non-standard integer types
+ * such as i33 or i56
+ *
+ * Algorithm Overview:
+ * ------------------
+ * The pass uses a two-phase approach to handle non-standard integer types:
+ *
+ * 1. Construction Phase:
+ *    - When encountering a non-standard integer type (e.g., i33), the pass
+first creates
```
linehill, line 20: I guess, clang-format messed up the formatting here.
```suggestion
 *    - When encountering a non-standard integer type (e.g., i33), the pass
        first creates
```

llvm_passes/HipPromoteInts.cpp
```diff
@@ -0,0 +1,377 @@
...
+              if (!GlobalVisited.count(cast<Instruction>(User))) {
+                U.set(R.New);
+              }
+            }
+          }
+
+          for (auto It = Replacements.rbegin(); It != Replacements.rend();
+               ++It) {
+            It->Old->eraseFromParent();
+          }
```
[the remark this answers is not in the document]
pvelesko: To preserve iterators traversing instructions from leaf to root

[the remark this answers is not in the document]
linehill: But why? I'm thinking you encountered restriction on the `Instruction::eraseFromParent()` that does not allow instructions with users to be erased and the reverse iteration order works-around it. I guess that explains it.

linehill (approved):
LGTM, thanks.

PR #983, opened 2025-03-24
H4I Refactor

pvelesko:
- Getting rid of old API `hipGetBackendName` 
- Updates for Unified Runtime headers
- Implement missing functions
- Add tests
- build spirv-tools if not found on system


colleeneb (approved):
It worked well for me after a few changes, since I use the `standalone` build:
I got initially:
```
CMake Error at test/CMakeLists.txt:6 (target_link_libraries):
  Target "mklshim_context_test" links to:

    hip::host

  but the target was not found.  Possible reasons include:
``` when I tried to build MKLShim.
I realized it needed `find_package(hip)    ` in the [CMakeLists.txt](https://github.com/CHIP-SPV/H4I-MKLShim/compare/6853b983aaf0fb50c7c4d9c1ba3e8e4bad71ca65...c114bbb6d63174045e5d4f8b04557eb1a809859e#diff-33394812ba204689144fd2f80832db83853ba1cb32403edb4e15fe4893e675fd). 

after that change, the include path for the standalone build needed to be adjusted, since it was pointing to `H4I-MKLShim/H4I-MKLShim/include`. To fix this I needed to do:
```
-                          ${CMAKE_SOURCE_DIR}/H4I-MKLShim/include)
+                          ${CMAKE_SOURCE_DIR}/include)
```
I'm not sure if that's needed in the non-standalone case or not though.

After that, I get a
```
/usr/bin/ld: cannot find -lSPIRV-Tools: No such file or directory
```
I guess libSPRIV-tools has to be in our environment paths for compiling anything now.
Once that is fixed, the stream/handle test and GAMESS worked well! Thanks a lot!!

File: shimCPURepro/backend_mismatch_reproducer.cpp [hunk 1]
```diff
@@ -0,0 +1,257 @@
...
+    // First check if sizes are different
+    if (handles1.size() != handles2.size()) {
+        std::cout << "DIFFERENCE: Handle count mismatch - " << name1 << ": " 
+                 << handles1.size() << ", " << name2 << ": " << handles2.size() << std::endl;
+        hasDifference = true;
+    }
+    
+    // Compare handles that both arrays have
+    for (size_t i = 0; i < minSize; i++) {
+        const char* name = (i < sizeof(HandleNames)/sizeof(HandleNames[0])) ?
```
Sarbojit2019, line 63: "sizeof(HandleNames)/sizeof(HandleNames[0])" is a fixed value why do we need to calculate in every iteration of the loop?

shimCPURepro/backend_mismatch_reproducer.cpp
```diff
@@ -0,0 +1,257 @@
...
+    
+    // Report on any extra handles
+    for (size_t i = minSize; i < maxSize; i++) {
+        const char* name = (i < sizeof(HandleNames)/sizeof(HandleNames[0])) ? 
+                       HandleNames[i] : "UNKNOWN";
+        
+        if (i < handles1.size()) {
+            std::cout << "DIFFERENCE: Extra handle in " << name1 << " - Handle[" 
+                     << i << "] " << name << ": 0x" << std::hex << handles1[i] << std::dec << std::endl;
+            hasDifference = true;
```
Sarbojit2019, line 82: hasDifference is always true as it is in 'if' and 'else' both. Do we need to have it inside loop or keep it outside with a check "if(max != min)"

shimCPURepro/backend_mismatch_reproducer.cpp
```diff
@@ -0,0 +1,257 @@
...
+    "QUEUE", 
+    "COMMAND_LIST"
+};
+
+// More detailed handle printing with names
+void print_handles(const char* prefix, const std::vector<uintptr_t>& handles) {
+    std::cout << "\n=== " << prefix << " (" << handles.size() << " handles) ===" << std::endl;
+    
+    for (size_t i = 0; i < handles.size(); i++) {
+        const char* name = (i < sizeof(HandleNames)/sizeof(HandleNames[0])) ?
```
Sarbojit2019, line 36: "sizeof(HandleNames)/sizeof(HandleNames[0])" looks to me a constant value. Why can't you calculate it before hand.

shimCPURepro/backend_mismatch_reproducer.cpp
```diff
@@ -0,0 +1,257 @@
...
+        if (handles1[i] != handles2[i]) {
+            std::cout << "DIFFERENCE: Handle[" << i << "] " << name << " - " 
+                     << name1 << ": 0x" << std::hex << handles1[i] << ", " 
+                     << name2 << ": 0x" << handles2[i] << std::dec << std::endl;
+            hasDifference = true;
+        }
+    }
+    
+    // Report on any extra handles
+    for (size_t i = minSize; i < maxSize; i++) {
```
Sarbojit2019, line 75: This seems to me a dead code. I would never expect it to get executed. Here is my reason:
max and min will be different only if both lists do have same size. You have already added a check on line no. 55. Hence after that min and max are always same. :)

Sarbojit2019 (commented):
Some suggestion feel free to make the change.

pvelesko:
@Sarbojit2019  dead code but that's a reproducer for another issue - OpenCL CPU runtime doesn't work w/ hipBLAS

PR #989, opened 2025-04-04
Refactor HipPromoteInt Pass

pvelesko:
- During testing of rocprim, some cases of unresolved nonstd types were found
- Extracted an IR reproducer from rocprim & added it as a test case
- Source of error in PHI node processing
- Add examples and more comments, especially regarding handling of PHI nodes
- PHI nodes special handling for incoming blocks being trunc or zext
- Defer handling of incoming values until they've been processed/promoted

### DIFF
--- a/llvm_passes/HipPromoteInts.cpp
+++ b/llvm_passes/HipPromoteInts.cpp
@@ -173,8 +173,58 @@ struct PendingPhiAdd {
   BasicBlock *IncomingBlock;
 };
 
+// Helper to adjust value V to type TargetTy using Builder
+static Value* adjustType(Value *V, Type *TargetTy, IRBuilder<> &Builder, const std::string& Indent = "") {
+    if (V->getType() == TargetTy) {
+        return V;
+    }
+
+    // TODO: IR Example
+    // Example 1: Source i32, Target i64
+    //   adjustType(%val_i32, i64, builder) -> creates '%zext = zext i32 %val_i32 to i64'
+    // Example 2: Source i64, Target i32
+    //   adjustType(%val_i64, i32, builder) -> creates '%trunc = trunc i64 %val_i64 to i32'
+    // Example 3: Source i64, Target <64 x i1> (Different types, same size)
+    //   adjustType(%val_i64, <64 x i1>, builder) -> creates '%bitcast = bitcast i64 %val_i64 to <64 x i1>'
+
+    unsigned SrcBits = V->getType()->getPrimitiveSizeInBits();
+    unsigned DstBits = TargetTy->getPrimitiveSizeInBits();
+
+    LLVM_DEBUG(dbgs() << Indent << "Adjusting type of " << *V << " from " << *V->getType() << " to " << *TargetTy << "\n");
+
+    Value* AdjustedV = nullptr;
+    if (DstBits < SrcBits) {
+        AdjustedV = Builder.CreateTrunc(V, TargetTy);
+         LLVM_DEBUG(dbgs() << Indent << "  Created Trunc: " << *AdjustedV << "\n");
+    } else if (DstBits > SrcBits) {
+        AdjustedV = Builder.CreateZExt(V, TargetTy);
+         LLVM_DEBUG(dbgs() << Indent << "  Created ZExt: " << *AdjustedV << "\n");
+    } else {
+        AdjustedV = Builder.CreateBitCast(V, TargetTy);
+         LLVM_DEBUG(dbgs() << Indent << "  Created BitCast: " << *AdjustedV << "\n");
+    }
+    return AdjustedV;
+}
+
 // Helper to get or create promoted value
-// TODO: explain when and on what this is called
+// This helper function is called whenever an instruction processing function
+// (e.g., `processBinaryOperator`, `processPhiNode`) needs an operand's value.
+// It ensures that if the operand `V` is part of the non-standard type chain,
+// an equivalent value with the `PromotedTy` is returned. It handles:
+// 1. Returning cached results from `PromotedValues`.
+// 2. Returning `V` directly if it's already the `PromotedTy`.
+// 3. Promoting `ConstantInt`s of `NonStdType` directly to `PromotedTy`.
+// 4. Creating `ZExt`/`Trunc`/`BitCast` instructions if `V` is an instruction
+//    result of `NonStdType` to convert it to `PromotedTy`.
+// 5. Returning other values (standard constants, arguments) directly.
+//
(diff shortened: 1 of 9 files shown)


llvm_passes/HipPromoteInts.cpp
```diff
@@ -125,23 +254,42 @@ static Value *getPromotedValue(Value *V, Type *NonStdType, Type *PromotedTy,
...
+  // Handle Constants specifically
+  // Example:
+  //   %c = add i33 %a, 1
+  // Becomes:
+  //   %c_promoted = add i64 %a_promoted, 1
+  if (auto *ConstInt = dyn_cast<ConstantInt>(V)) {
+    if (ConstInt->getType() == NonStdType) {
+      // Create a new ConstantInt with the promoted type
+      // zext resolves at compile time, it doesn't generate zext instructions
+      APInt PromotedValue = ConstInt->getValue().zext(PromotedTy->getIntegerBitWidth());
```
linehill, line 266: It's not correct to *always* zero-extend the values. You need to know what they are for. For example, if the value is used by a signed integer comparison instruction, the value needs to be sign-extended for it (and be aware that other instructions using the same constant may need it zero-extended).

pvelesko: Added an assert to check for this case for now, will implement proper fix once this case is encountered.

linehill: That’s not a good check. There are other potential users that need similar treatment. Right now, you are leaving a gaping hole for potential silent miscompilations.

pvelesko: ConstantExpr? Can you share what these potential users might be so I can fix this check?

linehill: You don’t need to know them [potential user]. If the code here doesn’t now that the users of the value are and how the value should be transformed for them for preserving the program semantics, it should stop there and sound an alarm (e.g. triggering by triggering an assertion).

pvelesko: I'm sorry but I am having a hard time following you. 

1. It's not correct to always zero-extend the values. You need to know what they are for (this depends on the type of user)
2. I implement a check to look at the immediate users and abort if it's used for signed comparison so I can just stick to using zext for now
3. You say that my check is not sufficient and that I need to abort on more types of users other than signed comparison
4. You say that I don't need to look at users and should abort if I don't know if I should zero or sign extend? But looking at the users is how you know what to do with it?

Anyways, I am refactoring this to shift the responsobility of promoting constants.

linehill: > 1. It's not correct to always zero-extend the values. You need to know what they are for (this depends on the type of user)

Yes, see the next point.

> 2. I implement a check to look at the immediate users and abort if it's used for signed comparison so I can just stick to using zext for now

No, that wouldn't be correct - see the following [example](https://alive2.llvm.org/ce/#g:!((g:!((g:!((h:codeEditor,i:(fontScale:14,j:1,lang:llvm,selection:(endColumn:22,endLineNumber:10,positionColumn:22,positionLineNumber:10,selectionStartColumn:22,selectionStartLineNumber:10,startColumn:22,startLineNumber:10),source:'define+i1+@src(i32+%25a0)+%7B%0A++%25t0+%3D+zext+i32+%25a0+to+i33%0A++%25t1+%3D+icmp+sgt+i33+%25t0,+u0x100000000%0A++ret+i1+%25t1%0A%7D%0A%0Adefine+i1+@tgt(i32+%25a0)+%7B%0A++%25t0+%3D+zext+i32+%25a0+to+i64%0A++%25promoted_c+%3D+zext+i33+u0x100000000+to+i64%0A++%25t1+%3D+icmp+sgt+i64+%25t0,+%25promoted_c%0A++ret+i1+%25t1%0A%7D%0A'),l:'5',n:'0',o:'LLVM+IR+source+%231',t:'0')),k:43.817790022749236,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((g:!((h:compiler,i:(compiler:alive,filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'1',trim:'1'),fontScale:14,j:1,lang:llvm,libs:!(),options:'',selection:(endColumn:31,endLineNumber:17,positionColumn:31,positionLineNumber:17,selectionStartColumn:31,selectionStartLineNumber:17,startColumn:31,startLineNumber:17),source:1),l:'5',n:'0',o:'alive-tv+(Editor+%231,+Compiler+%231)+LLVM+IR',t:'0')),k:62.79193205944798,l:'4',m:60.73113207547169,n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compiler:1,editor:1,fontScale:14,wrap:'1'),l:'5',n:'0',o:'%231+with+alive-tv',t:'0')),header:(),l:'4',m:39.26886792452831,n:'0',o:'',s:0,t:'0')),k:56.182209977250764,l:'3',n:'0',o:'',t:'0')),l:'2',n:'0',o:'',t:'0')),version:4).

> 3. You say that my check is not sufficient and that I need to abort on more types of users other than signed comparison.

Yep, because of this:
```
bool constValUsedForSignedCompare(const ConstantInt *CI) {
(...)

        // Check for other signed operations
        switch (UserInst->getOpcode()) {
        case Instruction::SDiv:
        case Instruction::SRem:
        case Instruction::AShr:
           LLVM_DEBUG(dbgs() << "Constant " << *CI << " used in signed operation: " << *UserInst << "\n");
          return true;
        default:
          // Other instructions are assumed to not require sign extension for this constant
          break;
        }
    }
  }
  return false;
}
```
What happens if you forgot an instruction here? Or future LLVM version brings new instructions that needs to be accounted? For sure, there would be a chance for miscompilation.

> 4. You say that I don't need to look at users and should abort if I don't know if I should zero or sign extend? But looking at the users is how you know what to do with it?

You were asking for the "potential users" that needs to checked too earlier:

> Can you share what these potential users might be so I can fix this check?

 ...  and I'm against it because of the problem in the above point 3). Instead of that you should proceed with the promotion transformation when you know what are you dealing with. For example, is the instruction to be promoted:

* an `ICmpInst` with unsigned predicate? --> zero extend its operands,
* an `ICmpInst` with signed predicate? --> sign extend its operands,
* is something else not accounted? --> Abort (rather than proceed with possibly broken code).

pvelesko:
since comments have been addressed, none of the rocPRIM tests are failing anymore and since the IR has been verified by alive2, merging.

PR #1023, opened 2025-07-22
[Level Zero] Fix Events & Mem Leaks

pvelesko:


colleeneb:
Thanks, this fixed the hang in the application we were testing on Aurora! 

I was hoping to use tests/runtime/TestEventRecordCircularDep.cpp to look at the trace and see if the race was visible there, but it’s not actually hanging for me, even after a huge number of iterations in the loop. :( But it does stop the application from hanging indeed!

pvelesko:
>I was hoping to use tests/runtime/TestEventRecordCircularDep.cpp to look at the trace and see if the race was visible there, but it’s not actually hanging for me, even after a huge number of iterations in the loop. :( But it does stop the application from hanging indeed!

Well it's not supposed to hang now with the fixes in but good to hear that it works.

colleeneb:
> Well it's not supposed to hang now with the fixes in but good to hear that it works.

Right, I was trying it with the older commits to see.

pvelesko:
> Right, I was trying it with the older commits to see.

I never got it to hang but the test in question did trigger circular dependency check I added in

PR #1039, opened 2025-08-20
Enable Resident Memory by Default

pvelesko:
+ Fix deallocation


colleeneb:
Thanks a lot, I'll test this out on Aurora today!

Two comments though:
1) I don't think we should call this "first touch" since it's not doing a first touch but rather making the memory resident, maybe changing the name to "make_memory_resident" or something?
2) I don't think we need to call zeContextEvictMemory if it's just called before zeMemFree, mem free is supposed to evict by default also `The application may free the memory without evicting; the memory is implicitly evicted when freed.` from https://oneapi-src.github.io/level-zero-spec/level-zero/latest/core/api.html#zecontextevictmemory , but did you see issues without it?

PR #1046, opened 2025-08-27
Fix event cleanup

pvelesko:


colleeneb:
Amazing, thanks! I see that there is a failing test in the CI (TestRecordEventBlocking.cpp). Any idea why? That one does work on Aurora, so I can't reproduce to dig into it.

PR #1050, opened 2025-09-24
Set HipVerifyPass off by default

linehill:
... due to pathological compilation time seen in CHIP-SPV/chipStar#1047. The pass can still be enabled through the CHIP_VERIFY_MODE environment variable regardless of build mode.

Note: it's not yet confirmed that this patch helps the issue.


colleeneb (approved):
Indeed this fixed the issue! @pvelesko saw the same speedup on the Aurora compute nodes as well.

PR #1052, opened 2025-09-30
Change log level from error to warning for zeEventQueryStatus duration check in CHIPEventLevel0.

### LINKED ISSUE #1048: "zeEventQueryStatus took us, exceeded 100us threshold" error message
In https://github.com/CHIP-SPV/chipStar/blob/c2d64dda33c8df7c064238003f37b637ffc5850c/src/backend/Level0/CHIPBackendLevel0.cc#L442, there is an error message if the time of the zeEventQueryStatus is greater than 100 us. 

We've seen this printed a lot with an application on Aurora with a lot of the prints which are slightly over 100us. The application runs correctly with the right results, but just prints this message many times.

LZ has no guarantees of when kernels will execute other than they will be done at a sync points. So LZ may not even submit any kernels until the zeEventQueryStatus call. Thus timing it and comparing to an arbitrary limit can just cause a lot of prints for a code that's executing correctly. Because a lot of things could happen during a zeEventQueryStatus call, this should not be a “logError” since it's really not an error. Could this be removed or changed to logWarn or logInfo? I’d argue there’s no need for the timing at all since again things could be launched in a zeEventQuery and it isn't guaranteed to be less than a specific amount of time, but just changing to logWarn/logInfo would be helpful too!

pvelesko:
Fixes #1048

### DIFF
--- a/src/backend/Level0/CHIPBackendLevel0.cc
+++ b/src/backend/Level0/CHIPBackendLevel0.cc
@@ -439,7 +439,7 @@ bool CHIPEventLevel0::updateFinishStatus(bool ThrowErrorIfNotReady) {
   auto duration = std::chrono::duration_cast<std::chrono::microseconds>(
       endTime - startTime);
   if (duration.count() > 100) {
-    logError("zeEventQueryStatus took {}us, exceeded 100us threshold",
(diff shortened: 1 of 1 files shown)


colleeneb (approved):
Thanks!

PR #1060, opened 2025-10-14
In order queues

pvelesko:
This PR refactors the queue synchronization system to replace the previous "last event tracking" approach with a more robust event dependency system. The changes eliminate race conditions and simplify the synchronization logic.

### 1. Queue Synchronization Refactor

**Removed Components:**
- `LastEvent_` member variable from `Queue` class
- `LastEventMtx` mutex for protecting last event access
- `updateLastEvent()` method for updating last event references
- `getLastEvent()` and `getLastEventNoLock()` methods
- `getSyncQueuesLastEvents()` method for collecting synchronization events

**Added Components:**
- `addDependenciesQueueSyncImpl()` template method for creating event dependencies
- Template-based approach allows different backends to implement their own event handle types
- More flexible dependency management system

### 2. Device Destruction Ordering Fix

**Before:**
```cpp
chipstar::Device::~Device() {
  LOCK(DeviceVarMtx);      // chipstar::Device::HostPtrToCompiledMod_
  LOCK(QueueAddRemoveMtx); // chipstar::Device::ChipQueues_
  // ... cleanup code ...
}
```

**After:**
```cpp
chipstar::Device::~Device() {
  // First handle queues with QueueAddRemoveMtx
  {
    LOCK(QueueAddRemoveMtx); // chipstar::Device::ChipQueues_
    // ... queue cleanup ...
  }

  // Then handle device variables with DeviceVarMtx
  {
    LOCK(DeviceVarMtx); // chipstar::Device::HostPtrToCompiledMod_
    // ... device variable cleanup ...
  }
}
```

### 3. Event Management Improvements

**Added Methods:**
- `checkEvents()` virtual method to base `Context` class (default no-op implementation)
- `checkEventsForAllContexts()` method to `Backend` class for processing events across all contexts

**Removed Calls:**
- `updateLastEvent(nullptr)` calls from queue removal paths
- `updateLastEvent(nullptr)` calls from cleanup routines

### 4. Queue Query Interface Changes

**Before:**
```cpp
bool query() {
  if (!LastEvent_)
    return true;

  if (LastEvent_->updateFinishStatus(false))
    if (LastEvent_->isFinished())
      return true;

  return false;
};
```

**After:**
```cpp
virtual bool query() = 0;  // Pure virtual method
```

### Event Dependency System

The new `addDependenciesQueueSyncImpl()` template method:

```cpp
template<typename EventHandle>
std::pair<std::vector<EventHandle>, LockGuardVector>
addDependenciesQueueSyncImpl(
    chipstar::Backend* BackendPtr,
    std::shared_ptr<chipstar::Event> TargetEvent,
    std::function<EventHandle(chipstar::Queue*, std::shared_ptr<chipstar::Event>&)> CreateMarkerInQueue)
```

**Key Features:**
- Creates marker events in other queues for synchronization
- Adds dependencies to target events
- Tracks events through backend event management
- Handles both default stream and blocking queue synchronization

### Synchronization Logic

**Default Streams (Legacy/Per-Thread):**
- Create markers in all blocking queues
- Add dependencies to ensure proper ordering

**Blocking Queues:**
- Create markers in default streams (legacy and per-thread)
- Synchronize with default stream execution

### DIFF
--- a/src/backend/OpenCL/CHIPBackendOpenCL.cc
+++ b/src/backend/OpenCL/CHIPBackendOpenCL.cc
@@ -1554,6 +1554,11 @@ void CHIPQueueOpenCL::addCallback(hipStreamCallback_t Callback,
       CL_COMPLETE, pfn_notify, Cb);
   CHIPERR_CHECK_LOG_AND_THROW_TABLE(clSetEventCallback);
 
+  // Flush the queue to ensure the callback fires on drivers like ARM Mali
+  // that require explicit flush for event callbacks to be triggered
+  clStatus = clFlush(this->get()->get());
+  CHIPERR_CHECK_LOG_AND_THROW_TABLE(clFlush);
+
   // Wait for the callback thread to enqueue its marker
   // This is blocking but necessary to get the marker event handle
   while (!markerReady.load()) {


src/backend/Level0/CHIPBackendLevel0.cc
pvelesko: @colleeneb

colleeneb:
Thanks! gamess worked fine with this PR.

The test `Unit_hipStreamPerThread_DeviceReset_1` was hanging for me on aurora (with this branch and LLVM-19). I'm investigating, will see if it's reproducible and let you know.

PR #1065, opened 2025-10-20
Disable making all memory resident

### LINKED ISSUE #1064: Memory is always resident
https://github.com/CHIP-SPV/chipStar/pull/1039
https://github.com/CHIP-SPV/chipStar/issues/1032

During last chat we decided that it's better to not keep all memory as resident and instead fix IBGA pass. It seems like the pass itself is correct but we are missing `zeKernelSetIndirectAccess` calls since on PVC there is a discrepancy where on-demand paging is reporting as available but not working.

pvelesko:
- On PVC on demand paging is reported as available but not working in reality. This results in skipped calls to zeKernelSetIndirectAccess
- Always call zeKernelSetIndirectAccess
- Default to non-resident memory in CmakeLists.txt
- Fixes #1064

### DIFF
--- a/src/backend/Level0/CHIPBackendLevel0.cc
+++ b/src/backend/Level0/CHIPBackendLevel0.cc
@@ -1256,12 +1256,18 @@ CHIPQueueLevel0::launchImpl(chipstar::ExecItem *ExecItem) {
   // Do we need to annotate indirect buffer accesses?
   auto *LzDev = static_cast<CHIPDeviceLevel0 *>(getDevice());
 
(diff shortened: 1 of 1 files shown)


colleeneb:
Thanks! This did work with the indirect-access-rocthrust examples. 

This changes from always setting memory to resident to always adding in a zeKernelSetIndirectAccess call for every kernel. I thought we were changing it to something like having a pass to determine indirect/direct accesses and if so, use or not use zeKernelSetIndirectAccess access. What's the benefit of always using zeKernelSetIndirectAccess vs. always setting memory to be resident?

PR #1089, opened 2025-11-03
More robust patch directory paths in configure_llvm.sh

Noerr:
Refactor script to find patch directories relative to script location rather than working directory where script is run.  Exit with failure message if patch directories not found rather than printing a warning.  It is crucial that patches be applied.


Noerr:
I'll also point out that,  `--only-necessary-spirv-exts` has no effect, as `ONLY_NECESSARY_SPIRV_EXTS` has no impact in the script.   Is this leftover from some old behavior?

PR #1149, opened 2026-02-19
HipIGBADetector: use UniformConstant for __chip_module_has_no_IGBAs

pvelesko:
Cherry-pick of 0bc8d63204afb3ec02a6d9d3f985913ef06c5995 onto main.

Original commit: HipIGBADetector: use UniformConstant for __chip_module_has_no_IGBAs

### DIFF
--- a/llvm_passes/HipIGBADetector.cpp
+++ b/llvm_passes/HipIGBADetector.cpp
@@ -120,9 +120,10 @@ static bool detectIGBAs(Module &M) {
       // chipStar runtime reads it.
       GlobalValue::ExternalLinkage, Init, MagicVarName, nullptr,
       GlobalValue::NotThreadLocal /* Default value */,
-      // Global-scope variables may not have Function storage class.
-      // TODO: use private storage class?
-      SPIRV_CROSSWORKGROUP_AS);
+      // UniformConstant is correct: we initialize it in this pass, and at
+      // runtime it is immutable (the runtime only reads it). CrossWorkgroup
+      // is for mutable device memory.
+      SPIRV_UNIFORMCONSTANT_AS);
 
   return true;
 }


llvm_passes/HipIGBADetector.cpp
```diff
@@ -120,9 +120,10 @@ static bool detectIGBAs(Module &M) {
       // chipStar runtime reads it.
       GlobalValue::ExternalLinkage, Init, MagicVarName, nullptr,
       GlobalValue::NotThreadLocal /* Default value */,
-      // Global-scope variables may not have Function storage class.
-      // TODO: use private storage class?
-      SPIRV_CROSSWORKGROUP_AS);
+      // UniformConstant is correct: we initialize it in this pass, and at
+      // runtime it is immutable (the runtime only reads it). CrossWorkgroup
+      // is for mutable device memory.
+      SPIRV_UNIFORMCONSTANT_AS);
```
linehill: "the runtime only reads it" statement is ambiguous: Does it mean the host driver reads it? Or the kernel? Or both? The statement is true for the former but not for the latter.

Switching storage class from CrossWorkgroup to UniformConstant is concerning here because the space for the latter can be scarce (e.g. 64kb on RTX 3080 through OpenCL). This can have increased change for launch failures for users already consuming most of the constant memory.

linehill: I see the change here is motivated by an issue mentioned [here](https://github.com/CHIP-SPV/chipStar/blob/ba865d42ac60001e81c1369ba9fec5056fc3ef50/llvm_passes/HipIGBADetector.cpp#L127-L132):

> CrossWorkgroup variables with initializers may be optimized or lowered incorrectly in the in-tree SPIR-V backend, producing invalid forward references.

Please, open a bug report on llvm-project and add link to the comment.

PR #1161, opened 2026-02-27
Switch over to hip-tests by default + HIP7 Fixes

pvelesko:


colleeneb:
Thanks! It looks like far less tests than before, it was something over 1000 previously. Were a lot of tests changed or dropped?

pvelesko:
> Thanks! It looks like far less tests than before, it was something over 1000 previously. Were a lot of tests changed or dropped?

 1.8k tests total. I ran on Aurora and hip-tests had the old way of detecting tests which required compilation on a GPU node.

PR #1162, opened 2026-03-02
Change to only add in marker events when the other queue is not empty and remove event generation for kernel launches

colleeneb:
This PR is a consequence of noticing a large slowdown in hipKernelLaunch times for the in-order queue code vs the out-of-order queue code for a specific application that has 21 streams and does a lot of kernel launching.

Profiling output from out-of-order run (older chipStar):
```
> iprof -b hip ./a.out 
                      Name |     Time | Time(%) | Calls |  Average |      Min |      Max |
  __hipUnregisterFatBinary | 100.67ms |  42.06% |     1 | 100.67ms | 100.67ms | 100.67ms |
           hipLaunchKernel |  52.57ms |  21.96% | 10000 |   5.26us |   3.21us |   3.27ms |
[...]
```

Output from in-order-queue chipStar:
```
> iprof -b hip ./a.out 

                      Name |     Time | Time(%) | Calls |  Average |      Min |      Max |
           hipLaunchKernel |    1.43s |  85.15% | 10000 | 143.09us |  34.16us |   3.15ms |
  __hipUnregisterFatBinary | 100.77ms |   6.00% |     1 | 100.77ms | 100.77ms | 100.77ms |
[...]
```

So the hipLaunchKernel time went up by 10x.

The slowdown was tracked down to the fact that we add marker events to all the other queues (like here: https://github.com/CHIP-SPV/chipStar/blob/cdd85e1800d9b946ee50931a17c1364093797c65/src/CHIPBackend.hh#L2165). This overhead is not very noticeable if there aren't many streams or if launches don't involve the NULL stream. However, in this application's case, there are many streams and a lot of kernel launches in the NULL stream, so a lot of implicit syncing.  If we know a queue has nothing in it or it has already sync'd, we don't need to add a marker event since there's nothing to wait on. Thus this PR adds in a check `&& !q->isEmptyQueue()` to only add in the mark event if the queue is empty. The check itself calls `zeCommandListHostSynchronize(ZeCmdListImm_, 0)` to check if the command list is finished or not, and returns true if it is and false if it isn't. We can change this LZ call to just checking the bool `IsEmptyQueue_` which we track in the code (added via `IsEmptyQueue_.store(...);` calls). However, there is a bug with LZ at the moment (https://github.com/argonne-lcf/AuroraBugTracking/issues/124) so we need to wait for that to be resolved before switching to that from the API call. 

Additionally, the launchKernel routine is generating and tracking an event for the kernel launch (https://github.com/CHIP-SPV/chipStar/blob/cdd85e1800d9b946ee50931a17c1364093797c65/src/backend/Level0/CHIPBackendLevel0.cc#L1296) this should no longer be necessary for in-order queues, as we are not tracking last events anymore. One part of removing this event involved removing https://github.com/CHIP-SPV/chipStar/blob/cdd85e1800d9b946ee50931a17c1364093797c65/src/backend/Level0/CHIPBackendLevel0.cc#L1353 . If I read the `git blame` right, this was originally from @linehill . Is this still needed? If so, what case was it? All the tests are passing without it and I'd love to remove the event generated unless we need it.

With these changes it is much better:

```
iprof -b hip ./a.out

                      Name |     Time | Time(%) | Calls |  Average |      Min |      Max |
           hipLaunchKernel | 476.67ms |  71.09% | 10000 |  47.67us |  12.67us |   2.62ms |
  __hipUnregisterFatBinary | 100.71ms |  15.02% |     1 | 100.71ms | 100.71ms | 100.71ms |
[...]
```
It is still slower than before, but once we switch to a bool from the LZ API call it will help even more.

For reference, here's a code that will show the slowdown:
```
#include <hip/hip_runtime.h>
#include <iostream>
#include <vector>

#define HIP_CHECK(expression) \
{ \
    hipError_t status = expression; \
    if (status != hipSuccess) { \
        std::cerr << "HIP error " << hipGetErrorString(status) << " at " << __FILE__ << ":" << __LINE__ << std::endl; \
        exit(1); \
    } \
}

// Kernel to fill array                                                                                                                                                                                                                          
__global__ void fillKernel(float* d_arr, int N) {
  for (int i = 0; i < N; i++ )
    d_arr[i] = i;
}

int main(int argc, char* argv[]) {


  int N = 100;
  int num_streams = 20;
  int num_arrays = 100;
  size_t size = N * sizeof(float);
  std::vector<float> h_arr(N);
  float **d_arr = (float **) malloc( num_arrays * sizeof(float *));

  printf( "running with N=%d and num_arrays=%d", N, num_arrays);

  HIP_CHECK(hipSetDevice(0));
  HIP_CHECK(hipDeviceSynchronize());

  for(int i=0; i< num_arrays;i++) {
    HIP_CHECK(hipMalloc(&d_arr[i], size));
    }

  hipStream_t *stream_array = (hipStream_t *) malloc( num_streams*sizeof(stream_array));
  for( int i=0; i< num_streams; i++) {
    HIP_CHECK(hipStreamCreate(&stream_array[i]));
  }

  HIP_CHECK(hipMemcpy( h_arr.data(), d_arr[0], size, hipMemcpyHostToDevice));
  HIP_CHECK(hipMemcpy( h_arr.data(), d_arr[1], size, hipMemcpyHostToDevice));

  for( int i=0;i < 100; i++ ) {
    for(int i=0; i< num_arrays;i++) {
        hipLaunchKernelGGL(fillKernel, 1,1, 0, 0, d_arr[i], N);
      }

    HIP_CHECK(hipMemcpy( d_arr[0], h_arr.data(), size, hipMemcpyDeviceToHost));
    HIP_CHECK(hipDeviceSynchronize());
  }

  for( int i=0; i< num_streams; i++) {
    HIP_CHECK(hipStreamDestroy(stream_array[i]));
  }
  for(int i=0; i< num_arrays;i++) {
    HIP_CHECK(hipFree(d_arr[i]));
}
  return 0;
}
```
```

### DIFF
--- a/src/CHIPBackend.hh
+++ b/src/CHIPBackend.hh
@@ -2208,7 +2208,7 @@ protected:
       // Create marker for per-thread default stream if used
       if (ChipDevice_->isPerThreadStreamUsedNoLock()) {
         auto PerThreadDefaultQueue = ChipDevice_->getPerThreadDefaultQueueNoLock();
-        if (PerThreadDefaultQueue && !LegacyDefaultQueue->isEmptyQueue() ) {
+        if (PerThreadDefaultQueue && !PerThreadDefaultQueue->isEmptyQueue() ) {
           std::shared_ptr<chipstar::Event> ChipMarkerEvent;
           EventHandle handle = CreateMarkerInQueue(PerThreadDefaultQueue, ChipMarkerEvent);
 

### CODE REVIEW

**Comment by colleeneb** [comment]
@linehill I confirm by hand-testing that TestLargeKernelArgLists runs and passes.

```
> hipcc spill.cpp 
> ./a.out 
PASSED
```
I also tried increasing the size of LargeStruct from `4 * 1024 - 16` to `127 * 1024 - 16`, and this also passed. `128 * 1024 - 16` did fail with a segfault on kernel launch, but I see this same behavior with an older chipStar without this PR.

PR #1232, opened 2026-04-13
fix: poll zeCommandListHostSynchronize to detect GPU hangcheck reset (#1191)

### LINKED ISSUE #1191: Concurrent stream operations deadlock at repeat=6+ (HeCBench concurrentKernels)
## Description

HeCBench `concurrentKernels` benchmark deadlocks when `repeat >= 6` on Intel Arc A770. The benchmark creates multiple concurrent HIP streams with events and inter-stream synchronization. Scaling is linear up to repeat=5, then hangs indefinitely.

SYCL (oneAPI icpx 2025.0.4) shows the same pattern but hits the wall slightly later (repeat=8), suggesting a shared driver-level limitation that chipStar triggers earlier.

## Scaling data (Intel Arc A770)

| repeat | SYCL | HIP (chipStar) |
|--------|------|----------------|
| 1 | PASS (2.86s) | PASS (3.81s) |
| 2 | PASS (5.64s) | PASS (7.62s) |
| 3 | PASS (8.44s) | PASS (11.44s) |
| 4 | PASS (11.22s) | PASS (15.24s) |
| 5 | PASS (14.00s) | PASS (19.05s) |
| 6 | PASS (16.81s) | **HANG** |
| 7 | PASS (19.57s) | **HANG** |
| 8+ | **HANG** | **HANG** |

## Reproducer

```bash
cd HeCBench/src/concurrentKernels-hip
make CC=hipcc && ./main 6   # hangs
make CC=hipcc && ./main 5   # passes
```

## Environment
- chipStar v1.2.1 (main, 062410dee254)
- Intel Arc A770, Level Zero backend
- LLVM 21.1.7
- oneAPI icpx 2025.0.4 (for SYCL comparison)

pvelesko:
Fixes #1191

Replaces `zeCommandListHostSynchronize(UINT64_MAX)` with a polling loop using a 1-second interval. When i915 hangcheck kills in-flight GPU batches, Level Zero returns `ZE_RESULT_ERROR_DEVICE_LOST` on the next poll instead of blocking forever. `hipDeviceSynchronize` then returns an error rather than hanging indefinitely.

### DIFF
--- a/src/backend/Level0/CHIPBackendLevel0.cc
+++ b/src/backend/Level0/CHIPBackendLevel0.cc
@@ -1872,14 +1872,26 @@ void CHIPQueueLevel0::finish() {
   }
   LOCK(CommandListMtx);
 
-  // host wait for command list to complete
-  if( ZeCmdListImmCopy_ != ZeCmdListImm_) {
-    zeStatus = zeCommandListHostSynchronize(ZeCmdListImmCopy_, UINT64_MAX);
+  // Poll with a finite timeout per iteration so that a GPU reset (e.g. from
+  // the i915 hangcheck / heartbeat mechanism) can be detected and reported
+  // instead of blocking forever (issue #1191).
+  //
+  // With UINT64_MAX, zeCommandListHostSynchronize() never returns after the
+  // driver kills in-flight batches during a GPU reset.  Using a 1-second
+  // poll interval allows Level Zero to surface ZE_RESULT_ERROR_DEVICE_LOST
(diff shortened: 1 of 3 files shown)


colleeneb:
Interesting -- this is not a problem on Aurora, the test in https://github.com/CHIP-SPV/chipStar/issues/1191 works fine on Aurora. I would not want to add in this poll. Could we add a guard so we don't have to do this on Aurora?

Issue #1246, opened 2026-04-29
HIPRTC cache: hiprtcGetLoweredName returns empty string after cache hit


Noerr:
## Summary

When `CHIP_MODULE_CACHE_DIR` is set (which is the default — `$HOME/.cache/chipStar`), a hipRTC compilation that uses `hiprtcAddNameExpression()` + `hiprtcGetLoweredName()` succeeds on the first run but fails on the **second** run with:

```
hipErrorLaunchFailure (Failed to find kernel via kernel name: )
```

The kernel name in the error message is **empty**. Setting `CHIP_MODULE_CACHE_DIR=""` (disabling the cache) makes the failure disappear because every run does a fresh compile.

The bug surfaces in any HIP application that compiles a non-`extern "C"` C++ kernel via hipRTC and uses the standard `AddNameExpression` / `GetLoweredName` pattern to learn the C++-mangled symbol name before calling `hipModuleGetFunction`. (Our local case: WarpM, an Aurora ALCC project.)

This is related to but **distinct from** issue #1142 ("HIPRTC cache appears write-only"). On platforms where `std::hash<std::string>` happens to be deterministic across runs (Aurora's libstdc++ is one), the cache *does* hit on subsequent runs — and that hit is what triggers the empty-name failure described here. PR #1231 was a proposed FNV-1a fix for #1142; even with that fix applied, the present bug would remain because the root cause is in the cache-load path itself, not in the cache key.

## Environment

- ChipStar commit `b508f598` (deployed at ALCF Aurora as `chipStar/llvm22/20260402-22/release`).
- Bug also reproduces against current `origin/main` HEAD — the relevant code paths are unchanged since `b508f598`.
- Backend: Level Zero on Intel Data Center GPU Max 1550 (Aurora).
- `CHIP_LAZY_JIT=on`, `CHIP_BE=level0`.

## Reproducer

A self-contained, single-file C++ reproducer (no external dependencies beyond hipRTC + libCHIP) is attached as a gist below. It uses a non-templated, non-`extern "C"` `__global__ void demo_kernel(int*, int)` and exercises the same API sequence as a real workload. Run twice with the same `CHIP_MODULE_CACHE_DIR`.

```cpp
#include <hip/hip_runtime.h>
#include <hip/hiprtc.h>

static const char *RTC_SOURCE = R"(
__global__ void demo_kernel(int *data, int n) {
  int i = blockIdx.x * blockDim.x + threadIdx.x;
[...]

Issue #1142, opened 2026-02-07
hipRTC module cache grows without bound and never hits


Noerr:
## Summary

The chipStar module cache (`~/.cache/chipStar/`) is write-only for hipRTC-compiled kernels. Each run writes new cache entries that are never loaded on subsequent runs, causing unbounded disk growth with no cross-run speedup.

Statically compiled HIP programs (typical hipcc approach) are not affected -- their cache works reasonably.

## Root Cause

hipRTC's SPIR-V output is non-deterministic across runs for two independent reasons:

### 1. Random temp directory paths embedded in SPIR-V

`hiprtcCompileProgram()` creates a random temporary directory and its absolute path leaks into the compiled SPIR-V through four places in `spirv_hiprtc.cc`:

```cpp
// createCompileCommand():
Append("-I" + WorkingDirectory.string());   // -I flag with random path
Append(SourceFile.string());                // absolute source path
Append(OutputFile.string());                // absolute output path

// createSourceFile() — lowered names magic variable:
File << ... << LoweredNamesFile << ";";     // absolute path as string literal
```

Clang embeds these paths in SPIR-V metadata (`DICompileUnit`, module ID). The `_chip_name_expr_output_file` string literal (used by the `HipEmitLoweredNames` pass) is compiled directly into the SPIR-V. When users pass `-g`, clang also embeds the working directory as `DW_AT_comp_dir`.

### 2. LLVM non-deterministic code generation

Even after fixing all path-related issues, LLVM produces non-deterministic SPIR-V for some kernels. Observed differences include basic block names with different numeric suffixes (`loadstoreloop264` vs `loadstoreloop252`) and different SPIR-V instruction IDs for the same logical blocks. This is a known class of LLVM issue caused by hash table iteration order depending on pointer addresses:

- https://github.com/llvm/llvm-project/issues/123791 (non-deterministic SPIR-V output, open)
- https://github.com/mgrang/non-determinism (42 iteration-order bugs found in LLVM)

In testing with 5 real-world kernels, 2 produced deterministic SPIR-V after path fixes; 3 remained non-deterministic (36–1327 differing bytes in ~125K object files).

## Observed Behavior

```
$ rm -rf ~/.cache/chipStar/*
$ CHIP_LOGLEVEL=info ./my_hiprtc_program    # run 1
# Log shows "Kernel compilation took X seconds" (no "Loaded from cache")
$ ls ~/.cache/chipStar/ | wc -l
3
[...]

PR #1247, opened 2026-04-29
fix: HIPRTC cache — persist name-expression map and harden cache file format

Noerr:
## What this PR does

Fixes a HIPRTC cache bug where `hiprtcGetLoweredName()` returns an empty string after a cache hit, causing the caller to pass `""` to `hipModuleGetFunction()` and downstream failures of the form `Failed to find kernel via kernel name: `. See accompanying issue **#1246** for the symptom-side report and reproducer details.

The PR also folds in three closely-related correctness improvements that come naturally with the fix and are best landed together because they share the cache file format / cache-key code path. See per-commit messages for details.

## Bug summary (one paragraph)

`chipstar::Program::NameExpressions_` is populated in two phases. `hiprtcAddNameExpression()` inserts `(name, "")` (empty value); `compile()` later runs the `HipEmitLoweredNames` LLVM pass and `getLoweredNameExpressions()` fills in the `.second` mangled-symbol values. On a cache hit, `hiprtcCompileProgram()` returns `HIPRTC_SUCCESS` immediately after `loadHiprtcCache()` succeeds — `compile()` never runs, the map stays empty-valued, and `hiprtcGetLoweredName()` returns a pointer to the empty `.second`. This is why a cache-warmed second run of any program using `AddNameExpression` + `GetLoweredName` (e.g. any non-`extern "C"` C++ kernel that needs its mangled name) silently breaks.

## Commits

1. **`fix: persist name-expression -> lowered-name map in HIPRTC cache`** — the primary fix.
   - Cache file format extended to a small framed binary layout with magic `CHC1`, length-prefixed SPIR-V, then a `(name_expression, lowered_name)` count + entries table.
   - `loadHiprtcCache()` rejects any file lacking the magic — pre-fix cache entries are auto-evicted on first use, so users don't need to manually clear `~/.cache/chipStar/`.
   - `saveHiprtcCache()` now writes to `<key>.tmp` and `fs::rename()`s on success — a process killed mid-write no longer leaves a partially-written cache file behind that subsequent runs treat as valid.
   - `computeHiprtcCacheKey()` now also hashes the registered name-expression keys, closing a separate aliasing gap (two programs with identical source/options but different `AddNameExpression` sets would have aliased to the same cache entry).
   - `hiprtcGetLoweredName()` returns `HIPRTC_ERROR_INTERNAL_ERROR` (with a clear log line) if a found map entry has an empty lowered name, instead of silently returning `HIPRTC_SUCCESS` with a pointer to `""`. Defensive — catches any future regression of this class at the API boundary.

2. **`fix: replace std::hash with FNV-1a for HIPRTC cache key`** — re-introduces the FNV-1a 64-bit helper from PR #1231 (closed unmerged). A stable, portable on-disk hash is a prerequisite of the new versioned cache file format — without it the version header doesn't help on platforms where `std::hash<std::string>` is randomized per process. The `fnv1a64()` helper is byte-identical to the one in #1231 to honor the prior author's intent.

3. **`test: regression for HIPRTC cache + name-expression interaction`** — adds `tests/hiprtc/TestHiprtcCacheNameExpressions.cc`. Compiles the same RTC source twice in one process, asserts that the second compilation's `hiprtcGetLoweredName` returns the same non-empty mangled name as the first. Pre-fix this assertion fails; post-fix it passes.

## Relationship to prior PRs and issues

- **#1142** (issue, "HIPRTC cache appears write-only") — referenced by this PR but **not directly fixed** by it. That issue describes the symptom of cache files being written but never loaded, which is a separate failure mode (most likely caused by `std::hash<std::string>` being non-deterministic across runs on some libstdc++/libc++ implementations). This PR includes the FNV-1a change from #1231 which would address #1142 on those platforms, but the primary bug fixed here (empty mangled-name on cache hit) is a different issue that surfaces on platforms where `std::hash<std::string>` *is* deterministic (e.g. Aurora's libstdc++).
- **#1231** (PR, "fix: use FNV-1a instead of std::hash for HIPRTC cache key", closed unmerged on Apr 14, 2026) — this PR re-introduces that change as commit 2/3 (`fnv1a64()` helper) because it is logically prerequisite to the new versioned cache format. If there is a reason the original FNV-1a PR was rejected that I missed, happy to drop that commit and address separately — but I think the cache-versioning context makes the case for landing it together.

## Validation

- Built the patched library against the same toolchain used for ChipStar's `chipStar/llvm22/20260402-22/release` deployment (chipStar LLVM 22 fork, commit equivalent to deployed commit `b508f598`, rebased onto current `main`).
- Standalone reproducer (run twice as separate processes with a shared `CHIP_MODULE_CACHE_DIR`) — pre-fix Run 2 fails with empty mangled name; post-fix all runs succeed. Cache file size grows by exactly the planned trailer overhead (49 bytes for one `(expr, lowered)` entry where `expr="demo_kernel"` and `lowered="_Z11demo_kernelPii"`).
- New regression test (`TestHiprtcCacheNameExpressions.cc`) added in this PR — passes on the patched build, would fail on `b508f598`.
- The `cache files written` log line `hiprtc: Saved/Loaded SPIRV from cache (key=…, names=N)` now reports the persisted entry count for visibility.

## Affected file summary

```
src/spirv_hiprtc.cc                                | +181 -25
tests/hiprtc/CMakeLists.txt                        |   +1
tests/hiprtc/TestHiprtcCacheNameExpressions.cc     | +112 (new)
```

No public API changes; cache file format is private.

---

## Appendix: cache-key aliasing in detail

The PR's commit 1 also makes `computeHiprtcCacheKey()` hash the registered name expressions. The reasoning, in case a reviewer wants the chain of inference:

1. **`hiprtcAddNameExpression()` is observable on output.** When the user calls `hiprtcAddNameExpression(prog, "foo")`, ChipStar synthesizes a magic variable like `_chip_name_expr_<n>` and injects it into the augmented source it hands to clang. The `HipEmitLoweredNames` LLVM pass then uses those magic variables to extract the C++-mangled symbol that `"foo"` lowered to. Concretely: a program that registered just `"foo"` produces a different augmented source — and thus different SPIR-V symbol-table contents — than the same-original-source program that also registered `"bar"`.

2. **The cache key is computed from the original (pre-augmentation) source**, plus the API-supplied virtual headers (`hiprtcCreateProgram`'s `Headers`/`IncludeNames` parameters), plus options. (Note: "headers" here is *only* the in-memory virtual headers the caller passes at program-create time — `computeHiprtcCacheKey` is blind to filesystem-resolved `#include` files reached through `-I` paths or system header dirs. That is a separate soundness gap, out of scope for this PR.) So two programs A and B with identical original source / virtual-headers / options but different `AddNameExpression` sets compute the **same** cache key.

3. **Same key → same cache file → "aliasing".** If A runs first and writes the cache, then B starts and computes the same key, B sees a cache hit and loads A's SPIR-V — even though A's compilation didn't include B's name expressions in the augmented source.

### What goes wrong without this fix

Pre-fix (cache stores only SPIR-V):
- A registers `foo`, compiles, saves SPIR-V containing `foo`'s mangled symbol.
- B registers `foo` AND `bar`, hits cache, loads A's SPIR-V — which has no info about `bar`.
- B's `NameExpressions_` map already had `foo→""` and `bar→""` from the pre-compile registration; the cache-hit path skips `getLoweredNameExpressions()`, so both stay empty. Both lookups fail (the primary bug fixed by this PR's commit 1). The aliasing here is masked by the larger bug.

Post-fix without also putting names in the cache key:
- A registers `foo`, compiles, saves `(SPIR-V, [foo→_Z3fooPi])`.
- B registers `foo, bar`, hits cache, loads `(SPIR-V, [foo→_Z3fooPi])`. The new `loadHiprtcCache` fills in `foo` in B's map. **`bar` stays unmapped.**
- B's `hiprtcGetLoweredName(prog, "bar", ...)` returns the new `HIPRTC_ERROR_INTERNAL_ERROR` from the defensive check (instead of silently returning a pointer to `""`) — so the failure is loud and clearly diagnosable, but it is *still* a failure, and B's compile succeeded so the user has no obvious reason to expect it.
- The right behaviour: B should have gotten a cache *miss* (because its registered name set differs from A's), recompiled fresh, and produced a SPIR-V that maps both `foo` and `bar`.

### Significance

- **Real-world likelihood**: low. Most code (WARPM included, where this PR's diagnosis originated) registers a fixed, code-determined set of kernel names per RTC program. Two programs with identical source+options but *different* registered name sets would have to be intentionally constructed.
- **Severity if it happens, post-fix**: loud (`HIPRTC_ERROR_INTERNAL_ERROR` from `GetLoweredName`) — not silent. So even without putting names in the cache key, this PR's defensive check at the API boundary already converts the worst-case symptom from "silent kernel name mystery" to "explicit error."
- **Why fix it anyway**: cache keys should reflect every input that affects the meaningful output. Registered name expressions affect the lowered names extracted from the SPIR-V, so they belong in the key. It's a small change and prevents a class of "weird intermittent error" reports that would otherwise be hard to diagnose.

The cost is one additional small loop in `computeHiprtcCacheKey()` (iterating the already-sorted `std::map<std::string, std::string>` of name expressions and hashing the keys). Map values are empty at hash time (filled only after compilation), so only the keys contribute to the hash, making the key computation deterministic.

### DIFF
--- a/src/CHIPBackend.hh
+++ b/src/CHIPBackend.hh
@@ -848,10 +848,12 @@ class Program {
   /// Include headers.
   std::map<std::string, std::string> Headers_;
 
-  /// Name expressions added before compilation as key to the
-  /// map. After compilation they point to their lowered/mangled
-  /// names. The map value may also be empty meaning the lowered name
-  /// is unknown.
+  /// Name expressions added before compilation are inserted as keys with an
+  /// empty value. After a successful compilation (or a successful HIPRTC
+  /// cache load) the value holds the lowered/mangled name. An empty value
+  /// after compilation indicates a pipeline error: hiprtcGetLoweredName
+  /// treats it as HIPRTC_ERROR_INTERNAL_ERROR rather than returning a
+  /// pointer to an empty string.
   std::map<std::string, std::string> NameExpressions_;
 
   std::string ProgramLog_; ///< Captured compilation log.
--- a/src/spirv_hiprtc.cc
+++ b/src/spirv_hiprtc.cc
@@ -383,13 +383,31 @@ hiprtcResult hiprtcAddNameExpression(hiprtcProgram Prog,
   return HIPRTC_SUCCESS;
 }
 
+/// FNV-1a 64-bit hash — portable, deterministic across runs and platforms.
+/// std::hash<std::string> is implementation-defined and may produce different
+/// values across program runs (some libc++ randomize it) or across different
+/// compiler/stdlib versions, making the cache effectively write-only on those
+/// platforms. The same change was previously proposed in #1231 (closed
+/// unmerged); we re-introduce it here because a stable on-disk cache key is
+/// part of the new versioned cache file format ('CHC1' magic) — without a
+/// stable hash, the version header doesn't help.
+static uint64_t fnv1a64(const std::string &s) {
+  uint64_t hash = UINT64_C(14695981039346656037);
+  for (unsigned char c : s) {
+    hash ^= c;
+    hash *= UINT64_C(1099511628211);
+  }
+  return hash;
+}
+
 /// Compute a cache key for HIPRTC output based on source, headers, options,
 /// and registered name expressions.
-/// The key is a hash of all inputs that affect the SPIRV output AND the set
-/// of name expressions that need a lowered-name mapping. Without including
-/// name expressions, two compilations with the same source/options but a
-/// different set of registered name expressions would alias to the same
-/// cache entry, leaving some lowered-name lookups unmapped on cache hit.
+/// The key is a portable, stable hash of all inputs that affect the SPIRV
+/// output AND the set of name expressions that need a lowered-name mapping.
+/// Without including name expressions, two compilations with the same
+/// source/options but a different set of registered name expressions would
+/// alias to the same cache entry, leaving some lowered-name lookups unmapped
+/// on cache hit.
 static std::string computeHiprtcCacheKey(const chipstar::Program &Program,
                                          int NumOptions,
                                          const char *const *Options) {
@@ -412,8 +430,7 @@ static std::string computeHiprtcCacheKey(const chipstar::Program &Program,
   for (auto &[expr, lowered] : Program.getNameExpressionMap()) {
     combined += expr + "\n";
   }
-  std::hash<std::string> hasher;
-  return std::to_string(hasher(combined));
+  return std::to_string(fnv1a64(combined));
 }
 
 // Cache file format (binary, little-endian):
--- a/src/spirv_hiprtc.cc
+++ b/src/spirv_hiprtc.cc
@@ -27,10 +27,13 @@ THE SOFTWARE.
 #include "Utils.hh"
 #include "logging.hh"
 
+#include <cstdint>
 #include <cstdlib>
+#include <cstring>
 #include <fstream>
 #include <regex>
 #include <set>
+#include <vector>
 #include <chrono>
 
 struct CompileOptions {
@@ -380,8 +383,13 @@ hiprtcResult hiprtcAddNameExpression(hiprtcProgram Prog,
   return HIPRTC_SUCCESS;
 }
 
-/// Compute a cache key for HIPRTC output based on source, headers, and options.
-/// The key is a hash of all inputs that affect the SPIRV output.
+/// Compute a cache key for HIPRTC output based on source, headers, options,
+/// and registered name expressions.
+/// The key is a hash of all inputs that affect the SPIRV output AND the set
+/// of name expressions that need a lowered-name mapping. Without including
+/// name expressions, two compilations with the same source/options but a
+/// different set of registered name expressions would alias to the same
+/// cache entry, leaving some lowered-name lookups unmapped on cache hit.
 static std::string computeHiprtcCacheKey(const chipstar::Program &Program,
                                          int NumOptions,
                                          const char *const *Options) {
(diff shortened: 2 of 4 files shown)


pvelesko:
> ere is a reason the original FNV-1a PR was rejected that I missed, happy to drop that commit and address separately — but I think the cache-versioning context makes the case for landing it together.

I couldn't reproduce the original issue and simply forgot to tag you! thanks for the PR let's run CI

Noerr:
**The PR is entirely HIPRTC-scoped — zero impact on the typical combined-source build-time path.**

Two separate cache layers exist, and they don't overlap:

| Layer | What's cached | Where | Code path |
|---|---|---|---|
| **Build-time HIP** (typical case) | SPIR-V → native (L0/OCL) module | `<CHIP_MODULE_CACHE_DIR>/<hash>` (top-level) | `src/backend/Level0/CHIPBackendLevel0.cc:2930+`, `src/backend/OpenCL/CHIPBackendOpenCL.cc:981+` |
| **HIPRTC** (this PR) | source → SPIR-V | `<CHIP_MODULE_CACHE_DIR>/hiprtc/<key>` (subdir) | `src/spirv_hiprtc.cc` |

Concretely:
- `chipstar::Program` (the class whose `NameExpressions_` comment we updated) is referenced *only* from `src/spirv_hiprtc.cc` — confirmed by grep. The build-time path doesn't use it.
- `loadHiprtcCache`/`saveHiprtcCache`/`computeHiprtcCacheKey`/`fnv1a64`/`CHC1` magic all live in `spirv_hiprtc.cc` and are called only from `hiprtcCompileProgram`.
- The L0 backend has its *own* cache hasher (`std::hash` at `CHIPBackendLevel0.cc:2927`) and its own tmp/rename pattern (`<hash>_tmp`). The PR didn't touch either. So combined-source apps still go through the unchanged build-time cache.

**Three small caveats worth knowing about, none affecting the typical case:**

1. `CHIP_MODULE_CACHE_DIR` is shared between the two layers as a *root* directory, but they live in disjoint sub-paths (`<dir>/<hash>` vs `<dir>/hiprtc/<key>`). No collision.
2. Users who *do* use HIPRTC and have an existing on-disk HIPRTC cache will see a one-time cache miss after upgrade (old format lacks `CHC1` magic → auto-evicted, recompiled, re-cached). Build-time cache entries are untouched.
3. The pre-existing concurrent-writer hazard in the L0 backend (`<hash>_tmp`, no PID) still exists — this PR only fixed it on the HIPRTC side. That's a separate issue, out of scope for this PR, but worth filing as a follow-up if anyone cares about parallel-build-then-parallel-first-run scenarios.

In short: combined-source HIP users see no behavior change, no performance change, no cache invalidation, no disk-format change. The PR is scoped to the HIPRTC API surface.

pvelesko (approved):
Thank you for your contribution! @Noerr

PR #1271, opened 2026-05-27
cmake: add hipfft-config.cmake for find_package(hipfft)

pvelesko:
Install a CMake config file so apps can find_package(hipfft CONFIG) and link hip::hipfft when chipStar's hipFFT is installed.


pvelesko:
Closing the hipfft-config.cmake should ship with H4I-HipFFT or chipFFT

PR #1285, opened 2026-06-08
Add CHIP_ENABLE_DEVICE_PROGRAM_SCOPE_GLOBALS to gate program-scope globals

### LINKED ISSUE #1279: HIP programs fail on Rusticl
... because, AFAIK, Rusticl doesn't support program scope global variables (or, IOW, crossWorkgroup variables in global scope in SPIR-V). Following debug message appears on a debug build of Rusticl while running HIP programs:

```
SPIR-V parsing FAILED:
    In file ../../mesa/src/compiler/spirv/vtn_variables.c:2394
    Initializer for CrossWorkgroup variable 6 not yet supported in Mesa.
    692 bytes into the SPIR-V binary
```

This occurs on HIP device programs without global variables like on this simple case: `__global__ void k(int *Out) { *Out = 123; }`. That's because chipStar unconditionally inserts code for device-side dynamic memory allocations which introduces a global variable `__chip_var___chipspv_device_heap` into SPIR-V modules.

pvelesko:
Adds a CMake flag (default ON) to optionally omit the device heap and clock-counter program-scope CrossWorkgroup globals, which some OpenCL drivers such as rusticl/radeonsi cannot consume (fixes #1279); when disabled, device malloc/free are unavailable, clock* returns 0, and the device-malloc-dependent compile tests are skipped.


pjaaskel:
Would it possible to follow the original philosophy of if the CUDA feature (here the device-side malloc) is not used, it won't require the OpenCL/SPIR-V feature (PSG)? It might become cumbersome to hand-enable the features as the program needs it.

pvelesko:
it's enabled by default. for rusticl, you disable it.I don't think it's possible to make it gated since we can't tell if malloc() or clock() will be inside the kernel ahead of time @pjaaskel

pjaaskel:
Alright.

pjaaskel:
Hmm: Could it check if the _target device_ supports PSG at runtime and if it doesn't, do not produce the PSG structures/heap (or fail more gracefully if detecting the application needs them)? If the SPIR-V has malloc() calls it should scream in that case?

pvelesko:
I'll merge this for now so that we can get rusticl CI online and investigate options for improvement.

PR #1293, opened 2026-06-11
[L0] share one immediate CL per hardware queue to eliminate dispatch overhead

pvelesko:
Reduce Level Zero kernel dispatch overhead by sharing a single immediate command list per hardware queue instead of allocating one per submission.


pvelesko:
`hipcc_verify_detects_dropped_kernels` is failing now since the driver update on CI machine. merging

PR #1296, opened 2026-06-11
[L0] allow >4 GiB single allocations on Intel GPUs

pvelesko:
Permit single device allocations larger than the driver-reported 4 GiB cap on Intel Arc / Data Center GPUs via ZE_RELAXED_ALLOCATION_LIMITS and NEO unrestricted-size env vars.

### DIFF
--- a/src/CHIPBackend.cc
+++ b/src/CHIPBackend.cc
@@ -1360,13 +1360,20 @@ void *chipstar::Context::allocate(size_t Size, size_t Alignment,
     return HostPtr;
   }
 
+  // NOTE: Drivers (Intel L0/OpenCL on discrete GPUs) commonly report a
+  // conservative CL_DEVICE_MAX_MEM_ALLOC_SIZE / ze_device_properties.maxMemAllocSize
+  // (often 4 GiB) even though larger single allocations are achievable via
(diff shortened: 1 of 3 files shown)


pvelesko:
Closing without merging.

The two backends can't be unblocked the same way:

- **Level Zero**: `ZE_RELAXED_ALLOCATION_LIMITS_EXP_FLAG_MAX_SIZE` is a proper public extension and lifts the 4 GiB single-allocation cap cleanly, no env vars needed.
- **OpenCL**: there is no per-allocation equivalent. `clDeviceMemAllocINTEL`/`clSVMAlloc` enforce `CL_DEVICE_MAX_MEM_ALLOC_SIZE` and the only way to bypass it on NEO is the `NEOReadDebugKeys=1` / `AllowUnrestrictedSize=1` debug keys.

`NEOReadDebugKeys=1` is a broad 'enable all debug keys' gate with negative side effects, and forcing it on globally via a library static constructor isn't acceptable for the default build. Since we can't make both backends work without that env-var hack, we'd rather live with the driver-reported 4 GiB single-allocation limit for now than ship the debug-key workaround.

Can revisit if/when Intel exposes a non-debug OpenCL mechanism for unrestricted-size allocations.

PR #1297, opened 2026-06-11
[backend] disable IGC RetryManager recompilation

pvelesko:
Disable IGC's RetryManager so kernels are not silently recompiled with reduced optimization, avoiding the associated performance cliff.

### DIFF
--- a/src/backend/Level0/CHIPBackendLevel0.cc
+++ b/src/backend/Level0/CHIPBackendLevel0.cc
@@ -3197,6 +3197,20 @@ void CHIPModuleLevel0::compile(chipstar::Device *ChipDev) {
   auto Flags = ChipEnvVars.hasJitOverride() ? ChipEnvVars.getJitFlagsOverride()
                                             : ChipEnvVars.getJitFlags() + " " +
                                                   Backend->getDefaultJitFlags();
+  // On Intel GPUs, IGC's RetryManager can convert high-register-pressure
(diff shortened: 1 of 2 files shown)


pvelesko:
Closing for now — deferring in favor of the upstream IGC RetryManager fix that's in progress.

From dg2 (Arc A380) testing: `DisableRecompilation=1` does fix the kernel drop, but **only when applied to the module's first compile**. IGC latches its SIMD/flag state process-globally at first compilation, so applying the flag at a *rebuild* — the natural "detect dropped kernels → recompile with the flag" recovery — is a no-op. Confirmed that an in-process rebuild, a fresh Level Zero context, and a mid-process `setenv` all fail to bring the kernels back. So this PR's unconditional-at-first-compile approach is actually the right shape; what doesn't work is any *reactive* recovery after a default compile has already happened.

If we want reactive recovery later, it likely needs a **separate process**: compile the SPIR-V (+ rtdevlib) with the flag in a fresh process, then load the resulting native binary back via `ZE_MODULE_FORMAT_NATIVE` (a pure load, no IGC codegen, so the parent process's latch doesn't apply). That path recovered all dropped kernels on the A380.

pvelesko:
What is not clear is whether disabling recompilation would have negative performance effects across all kernels. Since it only kicks on during spills for very large kernels. Anyways, the bug is already reported to Intel and should be fixed in the next driver drop.

PR #1298, opened 2026-06-11
[driver] atexit _exit(0) shim to bypass libigdrcl.so dl_fini hang

pvelesko:
Register an atexit handler that drains chipStar then calls _exit(0), bypassing the Intel OpenCL ICD's deadlocking _dl_fini teardown (opt out with CHIP_SKIP_FINI=0).


pvelesko:
Can't reproduce anymore so closing for now.

PR #1319, opened 2026-06-29
Adding in to ctest run of the simple benchmarks

colleeneb:
This adds in runs for the simple benchmarks in `tests/benchmarks`.

PR #1321, opened 2026-07-01
OpenCL: fix O(N) idle-stream launch overhead (isEmptyQueue fast path)

pvelesko:
The OpenCL backend hard-coded `isEmptyQueue()` to `false`, so each default/blocking-stream sync created a cross-queue marker in *every* idle blocking stream — O(N) launch overhead (the `manySmallKernels*` benchmarks hit ~45x at 64 streams on every OpenCL device, while Level Zero was flat). Wired the already-maintained `IsEmptyQueue_` flag into the override so idle queues are skipped, matching Level Zero; also set the flag in `memPrefetchImpl`.

Includes the benchmark-enabling commit from #1319 so these run in CI, plus a fix to `hostRegisterMemcpyOverhead` to compare the median (not the outlier-dominated mean), which was failing on cpu_opencl.

Verified locally on B570 and i9-13900K CPU OpenCL: all three benchmarks pass (was O(N), now flat ≤1.4x).

includes https://github.com/CHIP-SPV/chipStar/pull/1319

@colleeneb

### DIFF
--- a/src/backend/Level0/CHIPBackendLevel0.hh
+++ b/src/backend/Level0/CHIPBackendLevel0.hh
@@ -413,6 +413,15 @@ public:
   // if it's true (not busy), update IsEmptyQueue_ and return true. else, return false.
   bool isEmptyQueue() override {
 #ifdef CHIP_LZ_API_QUERY_QUEUE_EMPTY
+    // Fast path: a queue that never received work (or was already observed
+    // empty and not resubmitted) has IsEmptyQueue_ == true and is definitively
+    // empty -- return immediately without a Level Zero round-trip. Without this,
+    // a null-stream launch that scans N idle blocking streams issues N
+    // zeCommandListHostSynchronize() calls, making per-launch cost O(N) in the
+    // number of idle streams (see tests/benchmarks/manySmallKernels). The driver
+    // query below stays authoritative for queues that actually have work.
+    if (IsEmptyQueue_.load(std::memory_order_relaxed))
+      return true;
     return (zeCommandListHostSynchronize(ZeCmdListImm_, 0) == ZE_RESULT_SUCCESS);
 #else
     bool is_empty = IsEmptyQueue_.load(std::memory_order_relaxed);
--- a/src/backend/Level0/CHIPBackendLevel0.hh
+++ b/src/backend/Level0/CHIPBackendLevel0.hh
@@ -413,13 +413,10 @@ public:
   // if it's true (not busy), update IsEmptyQueue_ and return true. else, return false.
   bool isEmptyQueue() override {
 #ifdef CHIP_LZ_API_QUERY_QUEUE_EMPTY
-    // Fast path: a queue that never received work (or was already observed
-    // empty and not resubmitted) has IsEmptyQueue_ == true and is definitively
-    // empty -- return immediately without a Level Zero round-trip. Without this,
-    // a null-stream launch that scans N idle blocking streams issues N
-    // zeCommandListHostSynchronize() calls, making per-launch cost O(N) in the
-    // number of idle streams (see tests/benchmarks/manySmallKernels). The driver
-    // query below stays authoritative for queues that actually have work.
+    // Fast path: an idle queue (IsEmptyQueue_ == true) is definitively empty,
+    // so return without a Level Zero round-trip. This keeps a null-stream launch
+    // that scans N idle streams at O(1) per stream instead of O(N)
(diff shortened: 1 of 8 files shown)


src/backend/Level0/CHIPBackendLevel0.hh
```diff
@@ -413,6 +413,15 @@ public:
   // if it's true (not busy), update IsEmptyQueue_ and return true. else, return false.
   bool isEmptyQueue() override {
 #ifdef CHIP_LZ_API_QUERY_QUEUE_EMPTY
+    // Fast path: a queue that never received work (or was already observed
+    // empty and not resubmitted) has IsEmptyQueue_ == true and is definitively
```
colleeneb: The change looks good but the comment is a bit verbose, could we simplify it? I can push a change to the comment if it's ok.

tests/benchmarks/hostRegisterMemcpyOverhead.hip
```diff
@@ -32,6 +32,15 @@ int main() {
   auto bench = [&](const char *label) {
     std::vector<double> timings(N_ITERS);
     CHECK(hipDeviceSynchronize());
+    // Untimed warmup so neither measurement pays one-off setup costs (first-touch
+    // paging, staging-buffer allocation, DMA-engine/queue init). Without it the
+    // bench that runs first (WITH register) is timed colder than the second,
+    // producing a spurious ~10% "slowdown" on GPUs where hipHostRegister gives
```
colleeneb: Do we need a warmup if we're taking the minimum? (i.e. shouldn't the minimum never add in the slower initial runs?

tests/benchmarks/hostRegisterMemcpyOverhead.hip
```diff
@@ -43,26 +52,39 @@ int main() {
...
-  std::cout << "overhead factor (avg): " << avg_with / avg_without << "x\n";
+  std::cout << "overhead factor (min): " << min_with / min_without << "x\n";
 
   free(HostBuf);
   CHECK(hipFree(DevBuf));
-// with hiphost register it should be faster, but as long as it's not slower
-// it's ok
-  if (avg_with / avg_without > 1.1) {
+// hipHostRegister only speeds copies up on datacenter GPUs (~2x on Aurora/PVC
+// via zexDriverImportExternalPointer). On client Arc (Alchemist/Battlemage) the
```
colleeneb, line 74: So using hipHostRegister slows down non-PVC Intel GPUs? maybe we can add in a "do-nothing" macro so we can avoid doing anything extra for them so it's not slower? (since it should not affect the time too much in theory)

tests/benchmarks/manySmallKernels.hip
```diff
@@ -86,22 +98,41 @@ int main(int argc, char **argv) {
   const int counts[] = {0, 1, 2, 4, 8, 16, 32, 64};
   const int N        = sizeof(counts) / sizeof(counts[0]);
   const int WARMUP   = 20;
-  const int ITERS    = 200;
+  const int ITERS    = 100;
+  // Repeat the whole sweep several times and keep the minimum p10 per point.
+  // Each point is thus measured in SWEEPS independent time windows spread
+  // across the run. A transient system-load window (the dominant source of
```
colleeneb: If the tests are running in serial now, are the extra sweep needed? (i.e. I'd rather keep it simpler if possible)

tests/benchmarks/hostRegisterMemcpyOverhead.hip
pvelesko: Yeah it's slower on client GPUs. I'd prefer to just keep the same path maybe the driver will get better?

tests/benchmarks/hostRegisterMemcpyOverhead.hip
pvelesko: fair point.

tests/benchmarks/manySmallKernels.hip
pvelesko: ok

tests/benchmarks/hostRegisterMemcpyOverhead.hip
colleeneb: ah, ok. fine to keep it, but any chance of tightening the tolerance from 1.5 to 1.3 or 1.2?

colleeneb (approved):
LGTM

Issue #1335, opened 2026-07-17
hipRTC cache does not consider changes to filesystem `#include` content from source pre-processing


Noerr:
### Problem

The hipRTC cache added in 5e24fc344fba (`computeHiprtcCacheKey`, `src/spirv_hiprtc.cc`) computes its key from `fnv1a64(source + hiprtcCreateProgram headers + options + name-expression keys)`. The "headers" here are strictly the in-memory virtual headers passed to `hiprtcCreateProgram` (`Program::getHeaders()`, populated only by the `addHeader()` call in `hiprtcCreateProgram`). Headers resolved from `-I` paths on the filesystem are **not** included in the hash, and there is no preprocessing step (no `-E` / `-no-integrated-cpp`) that would fold their content in. (Verified against `main` as of this writing.)

If a user's RTC source contains `#include "mykernel.h"` where `mykernel.h` is found via a `-I` path (not passed as an in-memory header to `hiprtcCreateProgram`), changing that file will not invalidate the cache. The stale compiled SPIR-V will be served silently.

### How ROCm handles this

AMD's Comgr cache avoids this problem by splitting compilation into two phases (`-no-integrated-cpp`):

1. **Preprocessing** (not cached) — clang expands all `#include` and `#define` directives
2. **Compilation** (cached) — the input is the fully preprocessed source, so the content hash naturally captures all header content

From `comgr-compiler.cpp`:
```cpp
// The -no-integrated-cpp is used to split the preprocessing stage from the
// rest of the compilation jobs. The cache doesn't handle source-code input,
// but can handle preprocessed input (to avoid dealing with includes).
Args.push_back("-no-integrated-cpp");
```

### Suggested fix

Before computing the cache key, run clang's preprocessor (`-E`) on the source to expand all includes, then hash the preprocessed output:

```
cache_key = hash(preprocessed_source + options + compiler_version)
```

This adds one clang invocation (preprocessor only — fast compared to full compilation), but guarantees correctness for all `#include` sources.

Alternatively, document that the cache only covers in-memory headers and provide `CHIP_MODULE_CACHE_DIR=""` as an escape hatch for users with filesystem includes that change.

### References

- hipRTC cache commit: 5e24fc344fba
- Feature request with analysis: #1143
- ROCm Comgr cache: [ROCm/llvm-project/amd/comgr](https://github.com/ROCm/llvm-project/tree/amd-staging/amd/comgr)

pvelesko:
[...]

PR #1336, opened 2026-07-17
hipRTC cache: include compiler version in the cache key

Noerr:
## Problem

`computeHiprtcCacheKey()` (`src/spirv_hiprtc.cc`) currently hashes only
`source + hiprtcCreateProgram headers + options + name-expression keys`. It does
**not** include any toolchain identity.

After a user upgrades LLVM/Clang (or chipStar) and reruns the same program, the
source hashes to the same key, so the warm hipRTC cache returns SPIR-V compiled
by the **old** frontend — silently. Depending on what changed between toolchains
this can mean stale codegen, missed optimizations, or a hard-to-diagnose
miscompile, with no signal to the user that a stale artifact was served.

Note this is chipStar's to fix even though POCL's own kernel cache *does*
version-guard (its key seed includes `LLVM_VERSION`): on a hipRTC cache hit
chipStar returns the cached SPIR-V and never re-runs the frontend, so POCL never
sees fresh input. This is the sibling of #1335 (filesystem `#include` content not
in the key); both are cache-key soundness gaps in the same function.

## Fix

Fold the toolchain identity into the key:

- **LLVM/Clang version** — `llvm-config --version`, already computed by
  `cmake/FindLLVM.cmake` as `LLVM_VERSION`, exposed to C++ as a new
  `CHIP_LLVM_VERSION_STRING` in `chipStarConfig.hh`.
- **chipStar version** — the existing `CHIPSTAR_VERSION`.

Three-line surface: one `set()` in `CMakeLists.txt`, one `#cmakedefine` in
`chipStarConfig.hh.in`, and the key-composition lines in `computeHiprtcCacheKey`.

## Testing (LLVM 21.1.8 / POCL main / Apple M4)

- **Regression** — cold run logs `Saved SPIRV to cache`, warm run logs
  `Cache hit — skipped clang compilation` at the same key. Normal caching intact.
- **Invalidation** — with a warm cache from "21.1.8", simulated an upgrade by
  bumping `CHIP_LLVM_VERSION_STRING` to a fake value and recompiling. The next
  run computed a **new** key and re-saved (a miss) instead of reusing the stale
  entry — confirming a toolchain change now invalidates the cache. Reverting the
  version restores the original key.

## Open question for reviewers

Is the **build-time** LLVM version the right granularity, or would you prefer a
**runtime** identity (e.g. hashing the resolved `clang`/`hipcc` binary, or
`clang --version`) so that swapping the compiler binary without rebuilding
chipStar also invalidates? Build-time matches POCL/Comgr precedent and is
zero-cost; runtime is more robust to out-of-tree toolchain swaps. Also open to
adding a git SHA if you'd want per-commit granularity.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -819,6 +819,8 @@ set(CHIP_DEBUG_BUILD 0)
 if(CMAKE_BUILD_TYPE STREQUAL "Debug")
   set(CHIP_DEBUG_BUILD 1)
 endif()
+# Toolchain version for the hipRTC compilation cache key (see spirv_hiprtc.cc).
+set(CHIP_LLVM_VERSION_STRING "${LLVM_VERSION}")
 configure_file(chipStarConfig.hh.in include/chipStarConfig.hh @ONLY)
 
 # PROJECT CONFIGURATION HEADER
--- a/chipStarConfig.hh.in
+++ b/chipStarConfig.hh.in
@@ -29,6 +29,11 @@
 
 #cmakedefine CHIPSTAR_VERSION "@CHIPSTAR_VERSION@"
 
+// Full version string of the LLVM/Clang toolchain chipStar was built against
+// (from `llvm-config --version`). Used as part of the hipRTC compilation cache
+// key so that an LLVM/Clang upgrade invalidates cached SPIR-V.
+#cmakedefine CHIP_LLVM_VERSION_STRING "@CHIP_LLVM_VERSION_STRING@"
+
 // not implemented yet
 #undef OCML_BASIC_ROUNDED_OPERATIONS
 
--- a/src/spirv_hiprtc.cc
+++ b/src/spirv_hiprtc.cc
@@ -23,6 +23,7 @@ THE SOFTWARE.
 
 #include <hip/hiprtc.h>
 // #include "macros.hh"
+#include "chipStarConfig.hh"
 #include "CHIPBackend.hh"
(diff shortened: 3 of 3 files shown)


Noerr:
### Relationship to the non-RTC module cache

For context on why the non-RTC path doesn't need an analogous change:

chipStar's module cache (`generateCacheName`, `src/backend/OpenCL/CHIPBackendOpenCL.cc`) keys on `hash(SPIR-V bytes + build options + device name + IGC env vars)`. For statically-compiled HIP it already captures the frontend compiler **implicitly**: the SPIR-V bytes *are* the compiler's output, so a toolchain upgrade means a rebuilt binary with different SPIR-V and therefore a different cache key.

The hipRTC cache can't rely on that — `computeHiprtcCacheKey` keys on the HIP **source string**, not the compiled SPIR-V — so identical source compiled by a newer toolchain collides on the same key and a stale artifact is served. This PR adds the toolchain version explicitly so the RTC cache gets the same guarantee the module cache already has for free.

Scope note: the POCL/LLVM **backend** version (SPIR-V → device binary) is not part of chipStar's key in either path; that's covered separately by POCL's own kernel cache (its key seed includes `LLVM_VERSION` / `POCL_BUILD_TIMESTAMP`) and the poclbinary's internal build-hash. Out of scope for this PR, which is about the HIP→SPIR-V frontend.

pvelesko (approved):

PR #1338, opened 2026-07-17
hipRTC cache: capture #include content via preprocessing (#1335)

### LINKED ISSUE #1335: hipRTC cache does not consider changes to filesystem `#include` content from source pre-processing

Noerr:
## Problem (issue #1335)

`computeHiprtcCacheKey` keys on the raw HIP source, so it never sees the *content* of headers pulled in by `#include` — in particular headers resolved from `-I` filesystem paths. Editing such a header does not change the key, so a stale, previously-cached SPIR-V is served silently.

## Fix

When the source may include headers, preprocess it (`clang -E -P`) and key on the preprocessed translation unit, which embeds all `#include`d content (matching what ROCm/Comgr do). `-P` suppresses line markers so no absolute temp paths leak into the key — verified two independent cold preprocesses produce byte-identical output.

**Self-contained fast path.** To avoid adding a preprocess subprocess to *every* call, a conservative scan skips it when the source has no `#include`/`#import` — then the raw source is already a complete key (in-memory headers are hashed separately; toolchain headers are covered by the compiler-version component from #1336). The scan de-splices `\`-newline continuations and matches an unanchored `#[ \t]*(include|import)`, so it does not miss indented or line-split directives; a match inside a comment/string only triggers a harmless extra preprocess, never a stale hit.

## Measurements (LLVM 21.1.8 / POCL main / Apple M4)

| Source | Warm cost | Notes |
|--------|----------:|-------|
| Self-contained (no `#include`) | ~0.02s | fast path — no temp dir, no preprocess |
| Has `#include` | ~0.56s | preprocess runs; key reflects header content |
| Has `#include`, `-I` header edited | miss | **invalidates correctly** (new key) |

Invalidation validated end-to-end: a program that `#include`s a filesystem header via `-I` gets a cache hit on rerun, and a **new key (miss)** after the header is edited — the exact scenario #1335 describes.

## Open questions for reviewers

1. **Include-path cost (~0.56s).** The preprocess currently reuses the full compile flags, so it expands `hip_runtime.h` (~3MB) whose content is already covered by the compiler-version key, plus hipcc-wrapper overhead. Trimming the forced includes / invoking `clang` directly would cut this substantially, but changes which `#if` branches the preprocessor sees — a correctness subtlety I left out of this PR. Worth a follow-up?
2. **Scan precision vs. safety.** The regex handles all realistic code; two pathological constructs (a comment between `#` and `include`; `\`+space+newline) could slip through. If you want zero-doubt safety I can switch to de-splice-then-substring (`"include"`/`"import"`), which has no realistic false negatives at the cost of more harmless false positives.

Closes #1335.

### DIFF
--- a/src/spirv_hiprtc.cc
+++ b/src/spirv_hiprtc.cc
@@ -159,7 +159,8 @@ static std::string escapeWithSingleQuotes(const std::string &Str) {
 static std::string createCompileCommand(const CompileOptions &Options,
                                         const fs::path &WorkingDirectory,
                                         const fs::path &SourceFile,
-                                        const fs::path &OutputFile) {
+                                        const fs::path &OutputFile,
+                                        bool PreprocessOnly = false) {
 
   std::string CompileCommand;
 
@@ -209,7 +210,21 @@ static std::string createCompileCommand(const CompileOptions &Options,
   if (!Options.HasO)
     Append("-O2");
 
-  Append("-c");
+  if (PreprocessOnly) {
+    Append("-E");
+    // Preprocess-only with options for determinism:
+    //  -P suppresses line markers, which would otherwise embed absolute
+    //     temp paths and make the output non-deterministic.
+    Append("-P"); 
+    // The source lives in a per-run (random) temp dir. This maps the
+    // temp dir to "." so those expand deterministically. assert() macro
+    // expansion to __FILE__ is probably the most likely case where this mapping
+    // is needed. (Supported by Clang >= 10 / GCC >= 8, within chipStar's
+    // toolchain baseline.)
+    Append("-ffile-prefix-map=" + WorkingDirectory.string() + "=.");
+  } else {
+    Append("-c");
+  }
 
   Append(SourceFile.string());
   Append("-o");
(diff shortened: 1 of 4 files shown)


Noerr:
rebased to main now that #1336 is merged.

pvelesko:
could you please add a test @Noerr

Noerr:
> could you please add a test @Noerr

I have added a CI test to verify the RTC cache behavior @pvelesko

Noerr:
Status update on the test @pvelesko:

The test passes on macOS (native + translator) and on the Linux rusticl and pocl-7.1/llvm-19 jobs. It fails only on the **LLVM 22 / x86 Intel GPU** jobs (native-release, native-debug, translator-debug).

I added diagnostic logging to surface the cause (it was previously swallowed). On those runners the `-E` preprocess pass this PR uses to build the cache key crashes the compiler:

```
hiprtc: preprocessing for the cache key failed:
clang++: warning: -lCHIP: 'linker' input unused [-Wunused-command-line-argument]
clang++: warning: -locml_host_math_funcs: 'linker' input unused ...
clang++: warning: argument unused during compilation: '-no-hip-rt' ...
Segmentation fault (core dumped)
```

The unused-argument warnings are benign (identical on macOS); the failure is `clang++` itself segfaulting during the preprocess-only (`-E -P -ffile-prefix-map`) invocation used to build the cache key. macOS LLVM 22 runs that same preprocess on the same source without issue, so this looks specific to the LLVM 22 build on the Intel x86 stack.

The change fails safe there — on the crash, caching is disabled for that program (never stale) — but the include-in-key behavior is effectively inert on that platform, and each include-bearing RTC compile now spawns a crashing preprocess subprocess before falling back.

Do you have any insight from that CI platform into why the preprocess step might be segfaulting there?

pvelesko:
Superseded by #1375 (squashed, rebased on main, authorship preserved).

PR #1352, opened 2026-07-20
props: query warpSize from the device instead of hard-coding it

pvelesko:
Both backends set \`props.warpSize\` to the build-time \`CHIP_DEFAULT_WARP_SIZE\` and only warned when the device did not support it, so on a device whose subgroup width differs (e.g. Mali) the reported \`warpSize\` was wrong and disagreed with the device-side \`warpSize\` variable (\`get_sub_group_size()\`).

Set it from the device query (\`CL_DEVICE_SUB_GROUP_SIZES_INTEL\` / Level Zero \`subGroupSizes\`): prefer \`CHIP_DEFAULT_WARP_SIZE\` when supported (chipStar pins warp-sensitive kernels to it), otherwise the device native size. Drivers that do not enumerate subgroup sizes at the device level (Mali OpenCL) keep the build-time default as a fallback; querying those needs a per-kernel probe (`clGetKernelSubGroupInfo`), left as a follow-up.

Pairs with the device-side runtime `warpSize` change. Verified: Intel Level Zero + OpenCL and rusticl (W6400, supports {32,64}) all report warpSize 32.

### DIFF
--- a/src/backend/Level0/CHIPBackendLevel0.cc
+++ b/src/backend/Level0/CHIPBackendLevel0.cc
@@ -2797,17 +2797,24 @@ void CHIPDeviceLevel0::populateDevicePropertiesImpl() {
   // as per gen architecture doc
   HipDeviceProps_.regsPerBlock = 4096;
 
-  HipDeviceProps_.warpSize = CHIP_DEFAULT_WARP_SIZE;
-  if (std::find(DeviceComputeProps.subGroupSizes,
-                DeviceComputeProps.subGroupSizes +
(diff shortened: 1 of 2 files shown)


pvelesko:
Closing as redundant: with warpSize kept as a compile-time constant (CHIP_DEFAULT_WARP_SIZE, set to 8 for Mali builds via CHIP_MALI_GPU_WORKAROUNDS in #1340), host props.warpSize should simply equal that constant to guarantee host==device. This device-query approach was only motivated by the abandoned runtime-warpSize proxy and adds no value now.

PR #1376, opened 2026-07-28
Move vtable function pointers into the generic address space

### LINKED ISSUE #1373: Device-side virtual functions: vtable function pointers emitted in global instead of generic address space
Virtual dispatch through a base pointer in device code fails to compile:

    InvalidModule: Invalid SPIR-V module: Casts from private/local/global address space are allowed only to generic
      <badref> = addrspacecast ptr @_ZN7DerivedD2Ev to ptr addrspace(1)

Repro:

    struct Base    { __device__ virtual ~Base(){} __device__ virtual int f() const { return 1; } };
    struct Derived : Base { __device__ ~Derived() override {} __device__ int f() const override { return 2; } };
    __global__ void k(int* out){ Derived d; Base* b = &d; out[0] = b->f(); }

This is a codegen problem, not a missing extension. Clang emits the vtable with its slots in the global address space, so each entry is a function cast into CrossWorkGroup, which is invalid whatever extensions are enabled (`--spirv-ext=+all` does not help). `--target=spirv64` and `--target=spir64-unknown-unknown` produce the same layout, so this is not chipStar specific.

With slots in generic (AS 4) instead, the existing `SPV_INTEL_function_pointers` is sufficient and both runtimes accept the result. Measured on Arc B570, module doing an indirect call through a vtable style constant:

| vtable slot AS | `--spirv-ext` | llvm-spirv | clBuildProgram | zeModuleCreate |
|---|---|---|---|---|
| global AS(1), current codegen | none | InvalidModule | | |
| global AS(1) | `+SPV_INTEL_function_pointers` | InvalidModule | | |
| generic AS(4) | none | cannot translate function pointer | | |
| generic AS(4) | `+SPV_INTEL_function_pointers` | OK | ACCEPTED | ACCEPTED |

Note OpenCL accepts it even though `cl_intel_function_pointers` is absent from `CL_DEVICE_EXTENSIONS`.

So two changes are needed: emit vtable function pointers in the generic address space, and pass `+SPV_INTEL_function_pointers` when translating.

Impact: blocks Intrepid2 entirely and 3 Trilinos STK test TUs (`utest_VirtualFunction.cpp`, `UnitTestConstruction.cpp`, `ngpMultiStateFieldTests.cpp`). Classes with `__host__ __device__ virtual` members are unaffected as long as no dispatch happens on device, which is STK's normal pattern and why its libraries build fine.

### LINKED ISSUE #1381: __cxa_pure_virtual undefined on the device: a pure virtual member silently kills every kernel in the program
Clang emits a reference to `__cxa_pure_virtual` in the vtable of every class that still has an unoverridden pure virtual member. Nothing defines that symbol on the SPIR-V platform, so the consumer rejects the whole device module:

    error : unresolved external symbol symbol #__cxa_pure_virtual at offset 56 in data segment #GLOBAL_VARIABLES

Every kernel in the program then fails to build. Nothing of that reaches the application: `hipDeviceSynchronize()` and `hipGetLastError()` keep returning `hipSuccess` and the results keep their initial values. All that shows up is chipStar's own stderr chatter:

    CHIP warning : Missing definition for '__cxa_pure_virtual'
    CHIP error : hipErrorNotInitialized (Program build failed.) in CHIPBackendOpenCL.cc:1268:compile
    CHIP error : Caught Error: hipErrorNotInitialized

Repro (needs #1373 fixed first, otherwise the module fails earlier):

```c++
struct Base { __device__ virtual ~Base() {} __device__ virtual int f() const = 0; };
struct Derived : Base { __device__ int f() const override { return 2; } };
__global__ void k(int *Out) { Derived D; Base *B = &D; Out[0] = B->f(); }
```

Verified on Arc B570, OpenCL, LLVM 22.0-native: the kernel writes 0.0 and reports success.

`tests/compiler/TestDeviceVirtualFunctions.hip` does not cover this because its base class gives every virtual member a body. A compile-only test would not catch it either, since the symbol is resolved at the device module build inside the driver, not at `hipspv-link`.

Fix: define `__cxa_pure_virtual` next to `__assert_fail` with the same semantics ROCm's device libs use. Reaching a pure virtual slot is a hard error, so report and abort.

### LINKED ISSUE #1382: Dead C++ vtables in device modules abort llvm-spirv (Tpetra, Xpetra, Galeri)
Clang emits the vtable, VTT, construction vtables and RTTI of every polymorphic class that merely appears in device code, even when no device code ever dispatches through them. Trilinos is full of these. `Tpetra::RowMatrix` is host-only polymorphic with virtual bases, so each HIP explicit instantiation carries a dead `_ZTV`/`_ZTT`/`_ZTC` set whose slots are almost all null and whose virtual base offsets are `inttoptr` constants.

Translating those aborts llvm-spirv and kills the device link of `Tpetra_RowMatrix_{INT_INT,DOUBLE_INT,LONG_LONG_INT}_LONG_LONG_HIP.cpp` plus several Xpetra and Galeri instantiations:

    SPIRVWriter.cpp:1359: transConstantUse: Assertion
    `(C->getType()->isPointerTy() || ExpectedType->isTypeUntypedPointerKHR())
     && "Only pointer type mismatches should be possible"' failed.

Pre-existing rather than new: on current main the same TUs already fail, earlier and with the #1373 message, `InvalidModule: Casts from private/local/global address space are allowed only to generic`.

`llvm-reduce` cuts it to a three line module: a construction vtable holding `inttoptr` virtual base offsets. The tables are dead, there is not one indirect call left in the module, but GlobalDCE keeps them because they sit in comdats.

`--spirv-ext=+SPV_KHR_untyped_pointers` does make llvm-spirv succeed, but the B570 driver then rejects the extension, verified with a direct `clCreateProgramWithIL` probe, so that is not a way out.

Fix: drop unreferenced Itanium vtable globals from the device module before translation. A chipStar device module is self-contained, so nothing outside it can link against these symbols. Match only the vtable prefixes so runtime visible globals such as `__chip_var_*` are untouched, and iterate to a fixed point, since erasing a VTT releases the construction vtables it pointed at.

pvelesko:
Adds an LLVM pass that moves C++ vtable function pointer slots into the generic address space, the only address space SPIR-V permits function pointers in, and repairs the slot loads and indirect call sites that read them.

Also folds in the three follow-ups that make it usable on real C++: constant GEPs are rebuilt when a vtable is retyped (classes with virtual bases), unreferenced vtable globals are dropped before translation, and `__cxa_pure_virtual` is defined on the device.

Closes #1373
Fixes #1381
Fixes #1382

The pass is a temporary workaround. The root cause is in clang, which types every vtable component with the default globals address space (`clang/lib/CodeGen/CGVTables.cpp`, `getVTableComponentType`); the upstream fix is https://github.com/llvm/llvm-project/pull/212452 and the pass comes out once every supported LLVM carries it. `HipFunctionPointerAS.cpp` carries a `WORKAROUND(...)` marker to that effect.

### DIFF
--- a/include/hip/spirv_hip.hh
+++ b/include/hip/spirv_hip.hh
@@ -259,6 +259,22 @@ __assert_fail(const char *assertion, const char *file, unsigned int line,
   abort();
 }
 #endif // defined(_WIN32) || defined(_WIN64) || defined(__APPLE__)
+
+// Clang materialises a reference to __cxa_pure_virtual in the vtable of every
+// class that still has an unoverridden pure virtual member. The symbol has to
+// resolve on the device even when the slot can never be reached, otherwise the
+// SPIR-V consumer rejects the whole module with
+//   unresolved external symbol #__cxa_pure_virtual in data segment
+// and every kernel in the program fails to build. ROCm supplies it from its
+// device libs; on the SPIR-V platform there is no equivalent, so define it here
+// with the same semantics: calling a pure virtual function is a hard error.
+// No printf here on purpose: it would put a device printf into every
+// translation unit, and each one costs HipVerify's pre-lowering checkpoints a
+// crashing llvm-spirv run in debug builds (about 13 s per TU).
+__device__ __attribute__((noinline)) __attribute__((weak)) void
+__cxa_pure_virtual() {
+  abort();
+}
 } // extern "C"
 #endif // defined(__clang__) && defined(__HIP__)
 
--- a/include/hip/spirv_hip.hh
+++ b/include/hip/spirv_hip.hh
@@ -202,6 +202,20 @@ __assert_fail(const char *assertion, const char *file, unsigned int line,
   abort();
 }
 #endif // defined(_WIN32) || defined(_WIN64) || defined(__APPLE__)
+
+// Clang materialises a reference to __cxa_pure_virtual in the vtable of every
+// class that still has an unoverridden pure virtual member. The symbol has to
+// resolve on the device even when the slot can never be reached, otherwise the
+// SPIR-V consumer rejects the whole module with
+//   unresolved external symbol #__cxa_pure_virtual in data segment
+// and every kernel in the program fails to build. ROCm supplies it from its
+// device libs; on the SPIR-V platform there is no equivalent, so define it here
+// with the same semantics: calling a pure virtual function is a hard error.
+__device__ __attribute__((noinline)) __attribute__((weak)) void
+__cxa_pure_virtual() {
+  printf("Device-side pure virtual function called.\n");
+  abort();
+}
 } // extern "C"
 #endif // defined(__clang__) && defined(__HIP__)
 
--- a/llvm_passes/CMakeLists.txt
+++ b/llvm_passes/CMakeLists.txt
@@ -112,6 +112,7 @@ add_library(LLVMHipPasses MODULE
     HipDefrost.cpp
     HipDynMem.cpp
     HipEmitLoweredNames.cpp
+    HipFunctionPointerAS.cpp
     HipGlobalVariables.cpp
     HipIGBADetector.cpp
     HipKernelArgSpiller.cpp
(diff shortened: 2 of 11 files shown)


pvelesko:
This branch as pushed regresses Trilinos STK from 8/14 to 7/14 on Arc B570: `CL_OUT_OF_RESOURCES` in `desul::Impl::init_lock_arrays_hip`, and `stk_coupling` goes from passing to a timeout. Do not merge as-is.

Folding in the two commits that fix it, a constant GEP repair for retyped vtables and dropping unreferenced vtable globals (#1382), plus the device `__cxa_pure_virtual` definition (#1381). With those, STK is at 13/14.

pvelesko:
Necessary but not sufficient for RDC device virtual dispatch: with this applied, a `-fgpu-rdc` kynema-ugf build still leaves 40 `_ZTV` vtables and 18 virtual member bodies unresolved and the module build fails with `ZE_RESULT_ERROR_INVALID_MODULE_UNLINKED`. Filed as #1401 so this does not read as device virtual dispatch being finished.

PR #1389, opened 2026-07-29
Make DeviceVarMtx recursive to avoid teardown self-deadlock

zqb-all:
getOrCreateModule() holds DeviceVarMtx across compile(). If the backend exits the process during compile (e.g. the SPIR-V translator calls exit() on an invalid module), the atexit path deallocateDeviceVariables() re-locks DeviceVarMtx on the same thread and deadlocks.


pvelesko:
closing in favor of https://github.com/CHIP-SPV/chipStar/pull/1485 which avoids recursive lock + adds test. 

thank you for reporting the issue and PR

PR #1394, opened 2026-07-29
Do not deadlock when the device compiler exits during a module build (#1393)

### LINKED ISSUE #1393: Runtime deadlocks in its own exit handler when the device compiler calls exit() during a module build
If the backend compiler terminates the process from inside a module build, chipStar hangs forever instead of dying. The whole run then burns the full CI timeout with no diagnostic, which is indistinguishable from a slow test.

`Device::getOrCreateModule()` takes `DeviceVarMtx` and holds it across the backend compile. If the driver calls `std::exit()` during that compile, the atexit handler runs on the same thread and `deallocateDeviceVariables()` blocks on the non recursive mutex the build still holds.

Backtrace of the wedged process:

    #7  chipstar::Device::deallocateDeviceVariables ()  CHIPBackend.cc:1314   <- blocks on DeviceVarMtx
    #8  CHIPUninitializeCallOnce ()                     CHIPDriver.cc:244
    #12 CHIPUninitialize ()                             CHIPDriver.cc:254
    #13 __hipUnregisterFatBinary ()                     CHIPBindings.cc:6631
    #15 __run_exit_handlers ()                          stdlib/exit.c:108
    #17 SPIRV::SPIRVErrorLog::checkError(...)           libpocl.so           <- calls std::exit()
    #19 SPIRV::SPIRVModuleImpl::parseSPIRV
    #20 llvm::readSpirv
    #21 pocl_regen_spirv_binary
    #25 POclBuildProgram
    #26 CHIPModuleOpenCL::compile ()                    CHIPBackendOpenCL.cc:1263
    #20 chipstar::Device::getOrCreateModule ()          CHIPBackend.cc:1477   <- HOLDS DeviceVarMtx
    #22 chipstar::Device::prepareDeviceVariables ()
    #31 hipLaunchKernelInternal ()

Backend independent: the same wedge reproduces through Level Zero `zeModuleCreate`.

The trigger in practice is PoCL, whose `clBuildProgram` terminates the host process for a SPIR-V module it cannot accept, filed as pocl/pocl#2258 with the root cause upstream at KhronosGroup/SPIRV-LLVM-Translator#3939. Those are being fixed, but chipStar should not turn a misbehaving driver into an unkillable hang regardless. This showed up as a CI timeout on the macOS lane rather than a clear failure.

Observed PoCL behaviour by version: 2.16.0 exits 10, 2.15.0 exits 11, 7.1 and 2025.01.14 hit an LLVM assert and abort without hanging, 6.0 stack smashes.

A fix and a regression test follow. The test interposes `clBuildProgram` and `zeModuleCreate` in the executable so libCHIP dispatches into them and the interposer calls `exit(42)`, then forks and re execs so the parent bounds the child rather than burning the harness timeout.

pvelesko:
Fixes #1393. `Device::getOrCreateModule()` held `DeviceVarMtx` across the backend compile, so a driver calling `std::exit()` there ran the atexit handler on the same thread and `deallocateDeviceVariables()` blocked on the mutex the build still held. A thread local marker around the backend build lets `CHIPUninitializeCallOnce` skip teardown and log why, so the process exits with the compiler code instead of hanging.

Regression test added: it interposes `clBuildProgram` and `zeModuleCreate` so libCHIP dispatches into them, then forks and re execs so the parent bounds the child rather than burning the harness timeout. Before the fix it reports the deadlock after 45.8 s; after, it passes in 1.6 s. Skips cleanly where the interposer cannot fire.

Trade off worth noting: this path skips teardown, so the process exits with the compiler exit code and leaks whatever the runtime held. That is acceptable since the process is already dying, but it is deliberate rather than incidental.

Regression runs: dgpu opencl 1014/1014, dgpu level0 1045/1046, the single failure being `TestStaticLibRDC` which passes standalone in 75.9 s and timed out only under concurrent load.

Upstream causes are filed separately at pocl/pocl#2258 and KhronosGroup/SPIRV-LLVM-Translator#3939.

### DIFF
--- a/src/CHIPBackend.cc
+++ b/src/CHIPBackend.cc
@@ -1473,6 +1473,12 @@ chipstar::Module *chipstar::Device::getOrCreateModule(const SPVModule &SrcMod) {
 
   logDebug("Compile module {}", static_cast<const void *>(&SrcMod));
 
+  // The backend compiler runs with DeviceVarMtx held, and some backends end
+  // the process with exit() rather than reporting a rejected module. Let
+  // CHIPUninitializeCallOnce() detect that so it does not deadlock on the lock
+  // held here.
+  BackendModuleBuildScope BuildScope;
+
   auto start = std::chrono::high_resolution_clock::now();
   auto *Module = compile(SrcMod);
   auto end = std::chrono::high_resolution_clock::now();
--- a/src/CHIPDriver.cc
+++ b/src/CHIPDriver.cc
@@ -203,10 +203,33 @@ extern void CHIPInitialize() {
   std::call_once(Initialized, &CHIPInitializeCallOnce);
 }
 
+/// Nesting depth of backend module builds on this thread.
+static thread_local unsigned BackendModuleBuildDepth = 0;
+
+bool isThreadInBackendModuleBuild() { return BackendModuleBuildDepth > 0; }
+
+BackendModuleBuildScope::BackendModuleBuildScope() { ++BackendModuleBuildDepth; }
+
+BackendModuleBuildScope::~BackendModuleBuildScope() { --BackendModuleBuildDepth; }
+
 void CHIPUninitializeCallOnce() {
   logDebug("Uninitializing CHIP...");
   if (ChipEnvVars.getSkipUninit()) {
     logWarn("Uninitialization skipped");
+    return;
+  }
+  if (isThreadInBackendModuleBuild()) {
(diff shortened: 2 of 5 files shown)


pvelesko:
Duplicate of #1389, which fixes the same self-deadlock in 5 lines by making DeviceVarMtx recursive. That is the better fix: teardown proceeds normally instead of being skipped. Closing in favour of it.

PR #1443, opened 2026-08-06
Key the module cache on everything that shapes the device binary

pvelesko:
The module cache key misses several things that change what the driver emits, so a warm cache can serve a kernel built from different inputs. That is why CI disables caching entirely.

Missing from the OpenCL key:

- the runtime device library, linked by appendRuntimeObjects() after the cache is consulted, so editing bitcode/atomicAddFloat_emulation.cl kept serving the kernel built against the old one
- CHIP_JIT_FLAGS_OVERRIDE, which the x86 CI sets for the Zero-RK stage
- the Intel-conditional link step flags
- whether the compile took the compile+link or direct clBuildProgram path
- the driver version; the device compiler is the JIT here, and CL_DRIVER_VERSION was queried nowhere, so an IGC upgrade invalidated nothing
- compiler environment variables without an IGC_ prefix, notably OverrideDefaultFP64Settings which the x86 CI exports on every job

Not keyed on a chipStar version or build id on purpose. That changes every commit and would evict entries for changes that cannot affect a kernel, which is the common case since most chipStar work is host side and leaves the SPIR-V byte identical. Keying the real inputs invalidates exactly when the output moves.

Also fixes two things that made caching unsafe regardless of the key: the cache file concatenated every device's binary with no sizes and the reader handed the whole blob back as device 0's, corrupting multi-device programs; and the writer wrote straight to the final path so a concurrent reader could see a partial file. Entries are now framed behind a 'CHM1' magic and written via temp plus rename.

Level Zero gains the driver version and widened environment, and its two copies of the key computation are merged. Both backends move off std::hash<std::string>, which is implementation-defined, onto the fnv1a64 the HIPRTC cache already uses.

Adds TestModuleCacheInvalidation, which varies compilation inputs across runs under a fresh cache and asserts both directions: unchanged inputs must hit, changed inputs must miss. Verified it fails against the unfixed collector and passes once fixed.

This does not enable caching in CI. That is a separate decision and wants the measured hit rate first.

### DIFF
--- a/src/Utils.cc
+++ b/src/Utils.cc
@@ -278,35 +278,56 @@ bool startsWith(std::string_view Str, std::string_view WithStr) {
          Str.substr(0, WithStr.size()) == WithStr;
 }
 
-std::string collectIGCEnvironmentVariables() {
-  std::vector<std::string> igcVars;
-  logDebug("Collecting IGC environment variables...");
-  
+uint64_t fnv1a64(const std::string &S) {
+  uint64_t Hash = UINT64_C(14695981039346656037);
+  for (unsigned char C : S) {
+    Hash ^= C;
+    Hash *= UINT64_C(1099511628211);
+  }
+  return Hash;
+}
+
+std::string collectCompilerEnvironmentVariables() {
+  // Prefixes of variables that reach the device compiler. IGC_ covers IGC's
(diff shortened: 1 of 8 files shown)


pvelesko:
Merged into #1442; the cache key commits now live there.

PR #1473, opened 2026-08-24
Key the module cache on NEO debug keys, not the whole environment

### LINKED ISSUE #1470: Module cache never hits on Aurora (Causing fails of OpenSn performance test in Aurora CI)
In PR #1442, `collectCompilerEnvironmentVariables()` hashes the entire environment into the cache key when `NEOReadDebugKeys` is set: https://github.com/CHIP-SPV/chipStar/blob/cc98ae0f15881cae839c8748cc569f78ba900006/src/Utils.cc#L327

The hashing makes sense to do. Unfortunately on Aurora `NEOReadDebugKeys` is on by default (we use one the of the debug variables), and the environment of every mpiexec in a run will be different since HPE's launcher sets launcher-related variables to different value for every mpiexec, and things like the hostname it's running on will be different. Thus the key differs every launch and the cache will never hit.

This lead to Aurora CI OpenSn performance test failing since the case runs twice and relies on the second run hitting the cache to avoid ~8s JIT time. 

The environment variables that differ between runs on Aurora are: `PBS_*`, `HOSTNAME`, `HOST`, `ALPS_APP_ID`, `PALS_APID`, `PALS_SPOOL_DIR`, `TMPDIR`, `PMIX_*`. We could add in something so these are ignored. I would expect the same issue to show up on other systems that use schedulers like SLURM, so maybe we would hit it in the future too however.

pvelesko:
Fixes #1470

Under NEOReadDebugKeys the environment digest hashed every variable in the environment. On Aurora that gate is on by default and the launcher gives every mpiexec fresh PBS_*, PALS_*, PMIX_*, HOSTNAME and TMPDIR values, so the key was unique per launch and the module cache could never hit.

Match the shape of a NEO debug key instead: all 750 names in debug_variables_base.inl are CamelCase identifiers with no underscore and at least one lower-case letter, which no scheduler or launcher variable is. Real debug keys still invalidate; per-launch state no longer does.

Test: the invalidation driver gains two runs with the gate on and differing PBS_JOBID/PALS_APID/HOSTNAME, both of which must hit.

### DIFF
--- a/src/Utils.cc
+++ b/src/Utils.cc
@@ -24,6 +24,7 @@
 
 #include "logging.hh"
 
+#include <cctype>
 #include <cstdlib>
 #include <cstring>
 #include <fstream>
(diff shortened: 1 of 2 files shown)


pvelesko:
@colleeneb  aurora-ci flow seems to be broken but tested manually and it fixes the opensn issue

colleeneb:
Thanks! Yeah, it looks like it failed on the reporting step, but I restarted it and it's reporting passed now. Thanks for the quick fix!

Issue #1489, opened 2026-08-28
Level Zero: device atomics on hipMallocManaged memory abort with AtomicAccessViolation on PVC


pvelesko:
On Aurora (GPU Max 1550, NEO 25.18.33578.42, Level Zero, aurora-2026-08-28 at 08a106ad) a device atomic on `hipMallocManaged` memory aborts the process (`hipHostMalloc` is the same allocation kind, so it is expected to fault identically; not yet run):

```
Segmentation fault from GPU at 0xff94a0df8f9000, type: 2 (AtomicAccessViolation), level: 0 (PTE), access: 2 (Atomic), banned: 1, aborting.
```

Found with Kokkos `Kokkos_ContainersUnitTest_HIP` test `hip.scatterview_devicetype`: a ScatterView in `Kokkos::HIPManagedSpace` (`hipMallocManaged`) updated with `atomic_add`, `atomic_min`, `atomic_max`. `hip.scatterview` (`hipMalloc`) passes.

Cause: `CHIPContextLevel0::allocateImpl` backs `hipMemoryTypeUnified` and `hipMemoryTypeHost` with `zeMemAllocHost` (src/backend/Level0/CHIPBackendLevel0.cc:2681, :2700). On PVC host USM is mapped into the GPU page tables without atomic permission. Host USM was chosen in feac32955d56 for #131 and #1110 (Arc page migration).

Proposed fix: query `zeDeviceGetMemoryAccessProperties`; when `hostAllocCapabilities` lacks `ZE_MEMORY_ACCESS_CAP_FLAG_ATOMIC` and `sharedSingleDeviceAllocCapabilities` has `ZE_MEMORY_ACCESS_CAP_FLAG_CONCURRENT_ATOMIC`, use `zeMemAllocShared` for managed memory. B570 and dg2 lack concurrent shared USM and keep host USM. Report `hostNativeAtomicSupported` from the same flags.

Reproducers: `hostatomic_pinned.hip`, `hostatomic_managed.hip`, `probe_caps.hip`.

Related: #131, #1110, #446, https://github.com/alpaka-group/alpaka/issues/2641 (same fault on Aurora). No intel/compute-runtime report found.

PR #1513, opened 2026-08-29
llvm_passes: bypass the core-private cache on volatile global accesses

### LINKED ISSUE #1508: volatile global loads are plain OpLoad served from L1, Kokkos UnorderedMap_insert miscounts on PVC
Found with Kokkos `Kokkos_ContainersUnitTest_HIP` test `hip.UnorderedMap_insert` on Aurora PVC (Level Zero, IGC): the test expects 900 distinct keys and gets 901. `insert()` walks a hash list with `Kokkos::volatile_load` after the writer publishes a slot with a plain store, `__threadfence()` and a relaxed CAS. On chipStar the volatile load is `OpLoad` with the Volatile memory operand, which IGC serves from the core's L1, so a losing inserter reads a stale key and links its own slot behind it. CUDA lowers the same load to `ld.volatile`, which the PTX ISA (8.4.2, volatile Operation) defines as "equivalent to a relaxed memory operation with system-scope".

A standalone reproducer of the idiom (`cas_claim`) miscounts on PVC 1 in 40 runs with the Kokkos idiom and 3 in 40 when an early volatile read pulls the line into L1 first; with relaxed atomic loads instead it is 0 in 190, with release/acquire 0 in 190, and the no-fence control fails 28 in 40. Lowered device IR and SPIR-V before the fix:

```
%6 = load volatile i32, ptr addrspace(4) %add.ptr, align 4
%37 = OpLoad %uint %add_ptr Volatile|Aligned 4
```

Fix on branch `2026-08-28-volatile-relaxed` (commit c6b80b8477a1): `HipLowerVolatileAccessesPass` rewrites volatile 32 and 64 bit loads and stores through global or generic pointers into `load atomic volatile ... syncscope("device") monotonic`, emitted as `OpAtomicLoad %uint %p %uint_1 %uint_0` (Device, Relaxed). Tests: `TestFixVolatileLoadLoweringSPIRV`, `TestFixKokkosUnorderedMapInsert`, `hipVolatileAccesses-volatile-accesses`.

Related: the LLVM SPIR-V backend selects every `load atomic` as `OpLoad` (`SPIRVInstructionSelector::selectLoad`), so the native lane loses the atomicity (also for `__hip_atomic_load`); the translator lane, which Aurora uses, is correct. Kokkos's SYCL backend already implements `volatile_load` as a relaxed device-scope atomic load. Checked issues open and closed for volatile, UnorderedMap, volatile_load, OpAtomicLoad: #550, #1453, #1436 are adjacent, none covers this.

pvelesko:
CUDA gives a volatile access to global memory the meaning of PTX `ld.volatile` / `st.volatile`, an access that bypasses the core's private cache so a later reader sees the value, and HIP code such as `Kokkos::volatile_load` relies on it. chipStar emitted such accesses as plain `OpLoad` / `OpStore` with the `Volatile` memory operand, which only forbids eliminating or merging the access, so IGC still serves it from L1 and readers see stale data. On a PVC and a B570 an unmarked volatile reader misses a writer work-group's store 35840 times out of 35840.

`HipLowerVolatileAccessesPass` rewrites every volatile load and store through a global or generic pointer so it bypasses that cache. There are two forms and the build picks one, because no single form is accepted everywhere.

By default the access keeps its type, its volatility and its non atomic nature, and its pointer carries `CacheControlLoadINTEL` / `CacheControlStoreINTEL` at cache level 0 with `UncachedINTEL`. That is uncached at L1 and cached at L3, which is the CUDA meaning. It needs `SPV_INTEL_cache_controls` in the extension list the HIPSPV driver hands the SPIR-V producer, so `llvm-patches/llvm-23` adds it to both allow lists, and a configure probe asks the driver through `clang -###` whether the extension is there.

With `CHIP_ATOMICS_CACHE_BYPASS_WORKAROUND=ON`, or when that probe says the toolchain cannot allow the extension, the access becomes a relaxed device scope atomic instead. This is for consumers that reject a module carrying `OpCapability CacheControlsINTEL`: rusticl fails `clBuildProgram` with `-11 spirv_to_nir failed`, and Mali rejects the `OpExtension` at `clCreateProgramWithIL`.

The atomic form is the fallback rather than the default because an atomic is only legal where the allocation supports one. Level Zero lets a device report atomics as unsupported per allocation kind, and a PVC on Aurora reports `hostAllocCapabilities` as RW with no ATOMIC. Measured there per allocation kind and per operation, the atomic lowering still fixes the staleness and covers device and managed memory, and loses only `hipHostMalloc`, where every access aborts with `AtomicAccessViolation ... banned: 1` because IGC implements even `OpAtomicLoad` as an `atomic_or` read modify write. That is issue #1489. Cache controls cover that kind too, which is why they are the default where the toolchain and the consumer allow them.

Only naturally aligned 32 and 64 bit scalars are rewritten. OpenCL SPIR-V allows atomics on 32 bit types only, 64 bit under a capability, so nothing narrower or wider has a legal atomic form, and both lowerings share the filter so a source builds the same set of accesses either way.

Fixes #1508


pvelesko:
Aurora PVC manual run: PASS. `check.py ./ dgpu level0` at head `7f4e3239`, 1066/1066 tests passed, 0 failed, 0 not run.

PR #1516, opened 2026-08-29
runtime: bind storage for unregistered device variables in every module

### LINKED ISSUE #1500: Device-only globals are registered for the first SPIR-V module only: SyntheticVars de-duplication is process-global
Found by code reading while tracing device-only globals (the abort flag `__chipspv_abort_called`, and the new `__chipspv_abort_msg` from the #1487 fix); not yet reproduced at runtime.

`chipstar::Device::registerModuleVariables` (`src/CHIPBackend.cc:1385-1440`) discovers device-only globals that have no `__hipRegisterVar` entry from their `__chip_var_info_*` shadow kernels and synthesizes an `SPVVariable` for them. The de-duplication at `src/CHIPBackend.cc:1400-1424` uses a process-global `static std::vector<std::unique_ptr<SPVVariable>> SyntheticVars` and skips a name that was already synthesized for any module.

So in a program with two or more SPIR-V modules (any non-RDC build with several translation units, or a library plus the application), a device-only global present in every module is registered for the first module only. In the other modules its bind kernel never runs and `__chip_var_<name>` stays 0, so device code dereferences a null pointer: for `__chipspv_abort_called` the abort flag store goes to address 0, and `handleAbortRequest` finds no variable and cannot detect aborts from those modules.

Expected: the synthetic variable list is per module (or the de-duplication key includes the module), so every module that carries the global gets its own storage bound.

A test would build two `.hip` translation units without `-fgpu-rdc`, each with a kernel that calls `abort()`, and check that an abort in the kernel from the second module is detected.

pvelesko:
clang does not register comdat device variables (a function local static `__device__` variable, a C++17 inline `__device__` variable, a `__device__` static data member of a class template) with `__hipRegisterVar`, so the runtime discovers them from the `__chip_var_info_<X>` shadow kernels. That discovery de duplicated by name across all modules of the process, so when a second module carried the same variables it never got storage bound and `__chip_var_init_all` wrote the initializer through a zero address slot.

The runtime now binds storage for every unregistered variable in each module that carries it, and `registerModuleVariables` / `getOrCreateModule` share one path for that. The reproducer links two translation units with the same comdat variables, checks that every `__chip_var_info_<X>` kernel of each module has storage bound, and that a plain registered variable still round trips through `hipMemcpyToSymbol`, `hipMemcpyFromSymbol` and `hipGetSymbolAddress`.

Found with the Trilinos STK NGP tests on Aurora PVC, where every STK test executable faulted at module initialisation with `Segmentation fault from GPU at 0x0`.

Fixes #1500

### DIFF
--- a/src/CHIPBackend.cc
+++ b/src/CHIPBackend.cc
@@ -397,6 +397,38 @@ chipstar::DeviceVar *chipstar::Module::getGlobalVar(const char *VarName) {
   return *VarFound;
 }
 
+void chipstar::Module::addUnregisteredDeviceVariables() {
+  // A __chip_var_info_<X> shadow kernel means HipGlobalVariables.cpp lowered X
+  // into a __chip_var_<X> address slot: the module's code reads X through the
+  // slot and __chip_var_init_all writes X's initializer through it, so the
+  // slot must be bound to storage before the module's first launch whether or
+  // not the host registered X. clang emits no __hipRegisterVar for a
+  // function-local static __device__ variable, for an inline __device__
+  // variable or a template static data member that host code does not ODR-use
+  // (clang/lib/CodeGen/CGCUDANV.cpp, handleVarRegistration), or for any
+  // variable of a code object loaded through hipModuleLoadData. Such
+  // variables have no host pointer; they are looked up by name only.
+  static int NoHostPtr = 0;
+  for (auto *Kernel : ChipKernels_) {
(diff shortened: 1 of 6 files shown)


pvelesko:
Aurora PVC manual run: PASS. `check.py ./ dgpu level0` at head `af954e90`, 1058/1058 tests passed, 0 failed, 0 not run.

PR #1517, opened 2026-08-29
Keep the interior offset in hipHostGetDevicePointer and range-check registered host memory correctly

### LINKED ISSUE #1505: hipHostGetDevicePointer returns the allocation base for interior host pointers
`hipHostGetDevicePointer(&dev, hostPtr, 0)` on a pointer inside a `hipHostMalloc(hipHostMallocMapped)` or `hipHostRegister(hipHostRegisterMapped)` buffer returns the device address of the allocation base, dropping the offset of `hostPtr` within the allocation. ROCm returns the device virtual address plus the pointer's offset in its memory object (`hipamd/src/hip_memory.cpp`, `hipHostGetDevicePointer`: `getMemoryObject(hostPointer, offset)` then `virtualAddress() + offset`).

Found while running Trilinos STK `stk_mesh_utest` on Aurora PVC: STK does one `hipHostMalloc` per bucket and calls `hipHostGetDevicePointer` on an interior pointer for every field in the bucket (stk_mesh `FieldDataManager`, `get_host_pointer_for_device`), so every field but the first gets a device pointer to the wrong bytes.

Three related places in the runtime:

1. `src/CHIPBindings.cc` `hipHostGetDevicePointer` returns `AllocInfo->DevPtr` (the base) and, in the lazy `hipHostRegister` branch, registers the interior pointer as the record's host base.
2. `src/CHIPBackend.cc` `getAllocInfoCheckPtrRanges` measures the range from `DevPtr` even when the matched key is the host start, so interior pointers of `hipHostRegister`ed memory are rejected outright (`DevPtr` is null until the backing exists).
3. `src/CHIPBackend.cc` `RegisteredVarCopy` only syncs a host registered range when a kernel argument equals `DevPtr` exactly, so a kernel given an interior device pointer reads stale backing.

Reproducer `tests/runtime/TestFixHostGetDevicePointerOffset.hip` (branch 2026-08-28-hostgetdevptr-offset, commit 12bf18c541a8). Before the fix, on the Intel CPU OpenCL runtime:

```
hipHostMalloc: host base+4096 -> device base+0
  mismatch: kernel read 0x01234567 through the device pointer, host holds 0xdf0a2967 at that offset
FAIL: hipHostGetDevicePointer(&DevFirst, Base + Offsets[0], 0) -> hipErrorInvalidValue (1)
```

Fixes: commits 984bd4c38013, 1f69abe96032, 1e71dbdf9db8 on that branch; all three are needed for the reproducer to pass. Checked existing issues open and closed for `hipHostGetDevicePointer`: only #1374, which is about the symbol resolving, not its result. No upstream issue applies, the ROCm reference implementation already handles the offset.

pvelesko:
`hipHostGetDevicePointer()` on a pointer inside a `hipHostMalloc(Mapped)` or `hipHostRegister(Mapped)` buffer returned the device base of the allocation instead of base plus offset, and for a registered range an interior pointer queried before the base failed the lookup outright because `getAllocInfoCheckPtrRanges()` measured the record's extent from the still null `DevPtr` rather than from the key it matched. This keeps the interior offset in the returned device pointer (as ROCm does), range-checks against the matched key, registers the host base rather than the queried pointer when the device backing is created lazily, and makes `RegisteredVarCopy()` sync a registered range when a kernel argument points anywhere inside its backing rather than only at its base. Found with Trilinos STK on Aurora PVC, which does one `hipHostMalloc` per bucket and asks for the device pointer of each field's interior pointer. `TestFixHostGetDevicePointerOffset` covers both the `hipHostMalloc` and the `hipHostRegister` paths.

Fixes #1505

### DIFF
--- a/src/CHIPBackend.cc
+++ b/src/CHIPBackend.cc
@@ -277,16 +277,20 @@ chipstar::AllocationTracker::getAllocInfoCheckPtrRanges(void *DevPtr) {
   // Note: This function is called from within a locked context
 
   // upper_bound gives the first entry with key > DevPtr; step back one to get
-  // the candidate whose start address is <= DevPtr, then range-check it.
+  // the candidate whose start address is <= DevPtr, then range-check it. The
+  // key is the start of either the host or the device range of the record, so
+  // the range is measured from the key: a hipHostRegister'ed record has a host
+  // range at a different address than its device range, and no device range
+  // at all until hipHostGetDevicePointer creates the backing.
   const auto It = PtrToAllocInfo_.upper_bound(DevPtr);
   if (It == PtrToAllocInfo_.cbegin())
     return nullptr;
 
-  chipstar::AllocationInfo *AllocInfo = std::prev(It)->second;
-  void *End = (char*) AllocInfo->DevPtr + AllocInfo->Size;
+  const auto &Candidate = *std::prev(It);
+  void *End = static_cast<char *>(Candidate.first) + Candidate.second->Size;
 
   if (DevPtr < End)
-    return AllocInfo;
+    return Candidate.second;
 
   return nullptr;
 }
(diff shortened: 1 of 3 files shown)


pvelesko:
Aurora PVC manual run: PASS. `check.py ./ dgpu level0` at head `214ef875`, 1049/1049 tests passed, 0 failed, 0 not run.

PR #1519, opened 2026-08-29
build: hand the SPIRV-Tools external its compilers as full paths

### LINKED ISSUE #1486: SPIRV-Tools-External installs to /usr/local when gcc resolves to a different path
Seen building chipStar 9c0e17dc on Aurora (SLES 15, cmake 3.31.11) with no SPIRV-Tools on CMAKE_PREFIX_PATH, so CMakeLists.txt:581 builds it via ExternalProject. `ninja` succeeds, `ninja install` fails:

```
CMake Error at source/opt/cmake_install.cmake:46 (file):
  file INSTALL cannot copy file ... to "/usr/local/lib64/libSPIRV-Tools-opt.a": Permission denied.
```

cfgcmd.txt has `-DCMAKE_INSTALL_PREFIX=<build>/external/spirv-tools -DCMAKE_INSTALL_LIBDIR=lib`, yet the external's CMakeCache.txt ends up with `CMAKE_INSTALL_PREFIX:PATH=/usr/local` and `CMAKE_INSTALL_LIBDIR:PATH=lib64`; configure-out.log shows a warm then a cold configure pass.

Mechanism: CMakeLists.txt:577 passes bare names `-DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++`. The custom UPDATE_COMMAND makes every ninja invocation re-run the external's configure, which resolves `gcc` through PATH again. If that gives a different path than the cached one (a different PATH in the `ninja install` shell), cmake prints `You have changed variables that require your cache to be deleted`, deletes CMakeCache.txt, restores only the two compiler entries and re-runs configure without the other CMAKE_ARGS (`cmake::HandleDeleteCacheVariables`). Reproduced on cmake 3.31.11 and 4.3.2 by re-running `ninja SPIRV-Tools-External` with a gcc symlink first on PATH.

Fix: `find_program(CHIPSTAR_C_COMPILER gcc REQUIRED)` (and g++) so cfgcmd carries full paths; verified.

Workaround: `-DCMAKE_PREFIX_PATH=<an installed SPIRV-Tools>`.

pvelesko:
When SPIRV-Tools is not found, the ExternalProject that builds it passes `-DCMAKE_C_COMPILER=gcc` by bare name; its configure step re-runs on every build, and a PATH that resolves `gcc` differently between `ninja` and `ninja install` makes cmake delete the external's cache and reconfigure with only the compiler entries, so `CMAKE_INSTALL_PREFIX` falls back to `/usr/local` and the install fails with a permission error. Found building chipStar on Aurora, where the two steps ran from shells with different modules loaded.

The ExternalProject block moves into `cmake/SpirvToolsExternal.cmake` and resolves `gcc`/`g++` once with `find_program`, so the argument is a stable absolute path. `TestFix1486SpirvToolsInstallPrefix` configures that module on its own, runs the generated configure step twice with a gcc symlink directory prepended to PATH the second time, and asserts the external's cache keeps the build-tree prefix; it fails without the fix and nothing is downloaded or installed.

Fixes #1486

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -578,55 +578,7 @@ add_subdirectory(bin)
 find_package(SPIRV-Tools)
 if (NOT SPIRV-Tools_FOUND)
   message(STATUS "spirv-tools not found, building...")
-  # Download and build SPIRV-Tools
-  include(ExternalProject)
-  
-  set(SPIRV_TOOLS_VERSION "main")
-  set(SPIRV_TOOLS_INSTALL_DIR "${CMAKE_BINARY_DIR}/external/spirv-tools")
-  
-  # Create the include directory before it's referenced
-  file(MAKE_DIRECTORY "${SPIRV_TOOLS_INSTALL_DIR}/include")
-  file(MAKE_DIRECTORY "${SPIRV_TOOLS_INSTALL_DIR}/lib")
-  
-  set(CHIPSTAR_C_COMPILER gcc)
-  set(CHIPSTAR_CXX_COMPILER g++)
-  message(STATUS "CHIPSTAR_C_COMPILER: ${CHIPSTAR_C_COMPILER}")
-  message(STATUS "CHIPSTAR_CXX_COMPILER: ${CHIPSTAR_CXX_COMPILER}")
-  ExternalProject_Add(SPIRV-Tools-External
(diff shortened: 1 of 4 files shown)


pvelesko:
Aurora PVC manual run: PASS. `check.py ./ dgpu level0` at head `49857dbb`, 1058/1058 tests passed, 0 failed, 0 not run.

PR #1520, opened 2026-08-29
devicelib: map the missing <cmath> C libcalls to OpenCL builtins and lower llvm.ldexp

### LINKED ISSUE #1437: llvm.ldexp.f32.i32 has no SPIR-V mapping, so powf(2.f, n) fails to compile (HeCBench mxfp4-hip)
chipStar 2026.07.31, IGC all-fixes-2026.07.30, Arc B570, OpenCL backend.

`mxfp4-hip` does not compile at all:

```
hipcc -std=c++17 -Wall -I../mxfp4-cuda -O3 -c main.cu -o main.o
InvalidFunctionCall: Unexpected llvm intrinsic:
 llvm.ldexp.f32.i32 [Src: SPIRV-LLVM-Translator/lib/SPIRV/SPIRVWriter.cpp:5382 false]
clang++: error: hipspv-link command failed with exit code 8
```

The intrinsic comes from ordinary `powf(2.f, scale_exp - FLOAT8_E8M0_MAX_EXP)` at `main.cu:137`, which clang recognises as a power of two and lowers to `llvm.ldexp.f32.i32`.

Repro:
```
cd src/mxfp4-hip && make
```

The SYCL port compiles and passes on the same GPU, so this is a translator gap rather than a benchmark defect. `OpenCL.std` has an `ldexp`, so mapping `llvm.ldexp` to it in the SPIR-V writer looks like the fix.

### LINKED ISSUE #1509: Device module fails to link: libstdc++ float <cmath> overloads emit tgammaf, asinhf, scalbnf and other C libcalls that c_to_opencl.def does not define
Found with `Kokkos_CoreUnitTest_HIP` (TestMathematicalFunctions.hpp, float variants) on Aurora PVC: 340 s of device compile, then the module fails to link and the runtime tries the multi-input relink fallback before failing (and then recompiles on the next API call, #1507):

```
CHIP warning : Missing definition for 'tgammaf'
error : unresolved external symbol tgammaf at offset 1644 in instructions segment #4336 (aka kernel : _ZN6Kokkos4ImplL32hip_parallel_launch_local_memory...)
CHIP warning : Module created with multiple inputs is unlinked. Attempting fallback: create separate modules and link.
```

libstdc++ implements `std::tgamma(float)` and friends as constexpr wrappers around `__builtin_tgammaf`. HIP makes constexpr functions implicitly host device, so in device code they beat chipStar's `extern "C++" __device__ float tgamma(float)` unless `sp_math.hh` re-exports it into `std` (it does for acosh, lgamma, log1p, nextafter, not for these). Builtins without an LLVM intrinsic become calls to C linkage names that nothing defines: `sp_math.hh` has only C++ linkage overloads and `bitcode/c_to_opencl.def` (the table mapping `expf`, `lgammaf` etc. to OpenCL builtins, linked into device code by the device `-cc1` via `-mlink-builtin-bitcode hipspv-spirv64.bc`) lacks them:

```
asinhf atanhf fdimf ilogbf logbf nexttowardf remainderf scalblnf scalbnf tgammaf
```

The same route exists for long double (a 64 bit double on spirv64, `tgammal` and 18 more) and for `__builtin_tgamma` etc. on double: 44 names in total. On the Intel CPU OpenCL runtime the imports are silently resolved against the host libm instead (long double results are garbage there because of the x87 ABI). Because the table is linked at compile time, applications must be recompiled after the fix.

Reproducer: `tests/runtime/TestFixStdCmathFloatLibcalls.hip` (branch 2026-08-28-cmath-float-libcalls, commit 42f18932dd89). Fix: commit 8f11501ae00f adds the mappings to `c_to_opencl.def` (`scalbn` is OpenCL `ldexp` since OpenCL C fixes `FLT_RADIX` at 2; `scalbln` clamps the exponent; `nexttowardf` gets a helper). Related: #880 (closed without a fix, same family), #1399, #1404. Not fixed here: `std::nanf("")`, `std::nexttoward(double, long double)` and `std::modf(long double, long double *)` do not compile in device code.

pvelesko:
libstdc++ implements its float and long double `<cmath>` overloads as constexpr wrappers around `__builtin_*f` and `__builtin_*l`, which HIP makes implicitly `__host__ __device__`, so in device code they beat chipStar's `__device__` overloads and compile to C99 libcalls (`tgammaf`, `asinhf`, `scalbnf` and 41 more) that the device library never defined, leaving dangling imports in the SPIR-V module and `ZE_RESULT_ERROR_INVALID_MODULE_UNLINKED` with `unresolved external symbol tgammaf` when `Kokkos_CoreUnitTest_HIP` was linked on Aurora PVC. This maps the missing float, double and long double names in `bitcode/c_to_opencl.def` to OpenCL builtins, adding two table shapes for the signatures the existing macros cannot express (`ilogb` returning int, `scalbn`, `scalbln` and `nexttoward` taking a mixed second operand) and small `devicelib.cl` helpers for the `scalbln` exponent clamp and the `nexttowardf` conversion rule. It also carries the `llvm.ldexp` lowering it depends on: clang emits `llvm.ldexp` for `std::ldexp` and `scalbn`, which no SPIR-V producer accepts (https://github.com/CHIP-SPV/chipStar/issues/1437), so `HipLowerRoundIntrinsics` now rewrites it to the OpenCL `ldexp` builtin. That lowering is a temporary workaround, not the real fix. Both upstream fixes have already merged, llvm/llvm-project#195402 https://github.com/llvm/llvm-project/pull/195402 for the in tree SPIR-V backend and KhronosGroup/SPIRV-LLVM-Translator#3608 https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3608 for the translator, but neither has reached a version chipStar pins: #195402 landed on `main` after the `llvmorg-23.1.0-rc2` tag and is not on `release/23.x`, and the LLVM 21 lanes will never receive it. The block is delimited in the source by `WORKAROUND` and `END WORKAROUND` comments and comes out when those propagate; https://github.com/CHIP-SPV/chipStar/issues/1476 holds the removal condition for the whole pass. `TestFixStdCmathFloatLibcalls` evaluates every `<cmath>` overload against the host and fails on any `Missing definition for` warning; `TestStdLdexp` and the `mathIntrinsics` opt fixtures cover the lowering.

Fixes #1437
Fixes #1509

### DIFF
--- a/bitcode/c_to_opencl.c
+++ b/bitcode/c_to_opencl.c
@@ -22,6 +22,10 @@
 
 // See c_to_opencl.def for details.
 
+// The C library prototypes of the *l functions take long double, which is a
+// 64-bit double on spirv64 (see c_to_opencl.def).
+#define C2OCL_LONG_DOUBLE long double
+
 #define DEF_UNARY_FN_MAP(FROM_FN_, TO_FN_, TYPE_)                              \
   extern TYPE_ __chip_c2ocl_##FROM_FN_(TYPE_);                                 \
   TYPE_ FROM_FN_(TYPE_ x) { return __chip_c2ocl_##FROM_FN_(x); }
@@ -29,6 +33,14 @@
 #define DEF_BINARY_FN_MAP(FROM_FN_, TO_FN_, TYPE_)                             \
   extern TYPE_ __chip_c2ocl_##FROM_FN_(TYPE_, TYPE_);                          \
   TYPE_ FROM_FN_(TYPE_ x, TYPE_ y) { return __chip_c2ocl_##FROM_FN_(x, y); }
+
+#define DEF_UNARY_FN_MAP_RET(FROM_FN_, TO_FN_, RET_TYPE_, TYPE_)               \
+  extern RET_TYPE_ __chip_c2ocl_##FROM_FN_(TYPE_);                             \
+  RET_TYPE_ FROM_FN_(TYPE_ x) { return __chip_c2ocl_##FROM_FN_(x); }
+
+#define DEF_BINARY_FN_MAP_MIXED(FROM_FN_, TO_FN_, TYPE_, TYPE2_)               \
+  extern TYPE_ __chip_c2ocl_##FROM_FN_(TYPE_, TYPE2_);                         \
+  TYPE_ FROM_FN_(TYPE_ x, TYPE2_ y) { return __chip_c2ocl_##FROM_FN_(x, y); }
 
 #include "c_to_opencl.def"
 
--- a/bitcode/c_to_opencl.def
+++ b/bitcode/c_to_opencl.def
@@ -26,6 +26,15 @@
 // definitions for mapping the C math calls to corresponding OpenCL
 // built-ins.
 //
+// The C names arrive from several directions: HIP device built-ins in
+// include/hip/devicelib that expand to __builtin_*, libstdc++'s <cmath>
+// float and long double overloads (constexpr wrappers around __builtin_*f
+// and __builtin_*l, which HIP makes implicitly __host__ __device__ so they
+// win overload resolution in device code), and calls the optimizer
+// synthesises itself (pow(2, i) -> ldexp(1, i)). Only the built-ins that
+// have no LLVM intrinsic reach a C name; the rest become intrinsics that
+// the SPIR-V producer maps to OpenCL.std instructions directly.
+//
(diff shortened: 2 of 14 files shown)


pvelesko:
Aurora PVC manual run: PASS. `check.py ./ dgpu level0` at head `b26c3ede`, 1062/1062 tests passed, 0 failed, 0 not run.

PR #1522, opened 2026-08-29
hipGraph: fixes needed by Kokkos::Graph on chipStar

### LINKED ISSUE #1490: Stream capture runs work eagerly: captureIntoGraph() returns false unconditionally
Kokkos hip.graph_capture (core/unit_test/TestGraph.hpp:1109) fails on Aurora PVC with the Level Zero backend:

```
TestGraph.hpp:1109: Failure
Value of: contains(exec_right, data_1, 0)
  Actual: false (0 is not in data used in the captured kernel, got 5)
```

The kernel captured between hipStreamBeginCapture and hipStreamEndCapture has already run before the graph is launched (5 is GraphNodeKind::Capture, the value the captured kernel adds).

Mechanism: `chipstar::Queue::captureIntoGraph()` in `src/CHIPBackend.hh:2312` starts with an unconditional `return false; // TODO: fix this in graphs refactor` (commit c75a8674c33e), so every kernel launch, memcpy, memset and event op on a capturing stream executes immediately and `hipStreamEndCapture` hands back an empty graph. `initCaptureGraph()` in `src/CHIPBackend.cc:2193` also keeps `LastNode_` from the previous capture, so a second capture on the same stream chains its first node behind a node of another graph.

Backend independent, reproduces on the Intel CPU OpenCL runtime:

```
OCL_ICD_VENDORS=<dir with the intel cpu .icd> CHIP_BE=opencl CHIP_DEVICE_TYPE=cpu ./build/tests/runtime/TestFixKokkosGraphCaptureEagerLaunch
FAIL: kernel ran during capture, value is 5
```

Note that `tests/known_failures.yaml:6-7` masks `.*[Gg]raph.*` and `.*[Cc]aptur.*` wholesale, which is why CI never saw this.

Reproducer test and fix are on branch `2026-08-28-hipgraph-kokkos` (commits 1ae7863f65ce, c1e64e5e49f7).

### LINKED ISSUE #1491: hipGraphLaunch fails with hipErrorTbd (removeDependency) on graphs with redundant edges
Kokkos hip_graph.when_all_cycle fails on Aurora PVC with the Level Zero backend:

```
CHIP error : hipErrorTbd (Failed to find) in src/CHIPGraph.hh:151:removeDependency
C++ exception ... hip_graph_launch_wrapper( m_graph_exec) error( hipErrorTbd)
```

The graph is root -> f1 -> f2 -> f3, agg1 = when_all(f2, f3) -> f4, agg2 = when_all(f1, f4, f3) -> f5. The edges agg2->f1, agg2->f3 and agg1->f2 are implied by longer chains.

Mechanism: `CHIPGraphExec::pruneGraph_()` in `src/CHIPGraph.cc:267` enumerates every leaf to root path once, then for each nested (path, sub path) pair `unchainUnnecessaryDeps()` (`src/CHIPGraph.cc:235`) removes the first diverging edge. The pairs (P3, P1) and (P2, P1) both single out agg2->f1; the second call reaches `CHIPGraphNode::removeDependency()` after the edge is gone and it throws `Failed to find`, which `hipGraphLaunch` returns as `hipErrorTbd`. Any `when_all()` over nodes that a longer chain already orders produces this shape.

Backend independent, reproduces on the Intel CPU OpenCL runtime:

```
OCL_ICD_VENDORS=<dir with the intel cpu .icd> CHIP_BE=opencl CHIP_DEVICE_TYPE=cpu ./build/tests/runtime/TestFixKokkosGraphWhenAllRedundantEdges
FAIL: hipGraphLaunch(GraphExec, 0) -> hipErrorTbd
```

Reproducer test and fix are on branch `2026-08-28-hipgraph-kokkos` (commits 96d6c8de8f29, 86dcbfc032f7).

### LINKED ISSUE #1492: hipGraphGetNodes and hipGraphGetRootNodes segfault on a null array and miscount roots
Kokkos_CoreUnitTest_HIPInterOpGraph segfaults on Aurora PVC (Level Zero) in hip.graph_instantiate_and_debug_dot_print:

```
[ RUN      ] hip.graph_instantiate_and_debug_dot_print
1/1 Test #12: Kokkos_CoreUnitTest_HIPInterOpGraph ...***Exception: SegFault  0.25 sec
```

The test calls `hipGraphGetNodes(graph, nullptr, &num_nodes)` to size its query, the documented count only form, before it reaches hipGraphDebugDotPrint.

Mechanism: `hipGraphGetNodes` and `hipGraphGetRootNodes` in `src/CHIPBindings.cc:1215` and `:1227` do `*nodes = *(Nodes.data());` without checking `nodes`, so a null array is written through (SIGSEGV) and an empty graph dereferences an empty vector. A non null array only ever receives its first slot, and `hipGraphGetRootNodes` reports `getNodes().size()`, the total node count, instead of the number of roots.

Backend independent, reproduces on the Intel CPU OpenCL runtime:

```
OCL_ICD_VENDORS=<dir with the intel cpu .icd> CHIP_BE=opencl CHIP_DEVICE_TYPE=cpu ./build/tests/runtime/TestFixKokkosGraphGetNodesNullBuffer
(exit 139, SIGSEGV)
```

Reproducer test and fix are on branch `2026-08-28-hipgraph-kokkos` (commits 6b66ca257d72, 647b838e9d8b). The same Kokkos test then needs `hipGraphDebugDotPrint`, reported separately.

### LINKED ISSUE #1493: hipGraphDebugDotPrint is unimplemented (hipErrorNotSupported)
Once the hipGraphGetNodes segfault is fixed, Kokkos hip.graph_instantiate_and_debug_dot_print (core/unit_test/hip/TestHIP_InterOp_Graph.cpp) aborts because `hipGraphDebugDotPrint(graph, path, hipGraphDebugDotFlagsVerbose)` returns `hipErrorNotSupported`. The test expects a DOT file whose kernel node labels contain the kernel name, matching `[A-Za-z0-9_]+Increment[A-Za-z0-9_]+RangePolicy` or `Increment<.*RangePolicy<`.

Mechanism: `hipGraphDebugDotPrint` in `src/CHIPBindings.cc:225` is `UNIMPLEMENTED(hipErrorNotSupported)`.

Reproduces on the Intel CPU OpenCL runtime:

```
OCL_ICD_VENDORS=<dir with the intel cpu .icd> CHIP_BE=opencl CHIP_DEVICE_TYPE=cpu ./build/tests/runtime/TestFixKokkosGraphDebugDotPrint
FAIL: hipGraphDebugDotPrint(Graph, DotPath.c_str(), hipGraphDebugDotFlagsVerbose) -> hipErrorNotSupported
```

Reproducer test and an implementation (DOT digraph, one box per node with its type, kernel nodes named via `chipstar::Kernel::getName()`, launch dimensions and memset, host and event details behind the flag bits, child graphs as clusters, one edge per dependency) are on branch `2026-08-28-hipgraph-kokkos` (commits 49e85e099ff5, d43a6feb8a6f).

### LINKED ISSUE #1494: hipGraphLaunch fails with hipErrorTbd on graphs containing a child graph node
`hipGraphLaunch` on a graph holding a `hipGraphAddChildGraphNode` node returns `hipErrorTbd`:

```
CHIP error : hipErrorTbd (Attemped to execute GraphNode) in src/CHIPGraph.hh:427:execute
```

Kokkos wraps every stream captured region in a child graph node (`GraphNodeCaptureImpl<HIP>::capture` in core/src/HIP/Kokkos_HIP_GraphNodeKernel.hpp), so hip.graph_capture cannot submit its graph once stream capture records nodes.

Mechanism: `CHIPGraphNodeGraph::execute()` in `src/CHIPGraph.hh:426` only throws. `CHIPGraphExec::ExtractSubGraphs_()` in `src/CHIPGraph.cc:392` was meant to inline child graphs but edits a local copy of the node list that `compile()` never reads, wires the edges in the wrong direction, and reads `Nodes[i + 1]` past the end whenever the child graph node is the last node.

Reproduces on the Intel CPU OpenCL runtime:

```
OCL_ICD_VENDORS=<dir with the intel cpu .icd> CHIP_BE=opencl CHIP_DEVICE_TYPE=cpu ./build/tests/runtime/TestFixKokkosGraphChildGraphNodeLaunch
FAIL: hipGraphLaunch(GraphExec, 0) -> hipErrorTbd
```

Reproducer test and fix (the node instantiates and runs its child graph at its place in the schedule, `ExtractSubGraphs_` removed) are on branch `2026-08-28-hipgraph-kokkos` (commits 1772f9e95ada, 7dd9e749d8e6).

### LINKED ISSUE #1495: N-th hipGraphLaunch of one hipGraphExec_t executes the graph N times
Launching the same `hipGraphExec_t` repeatedly re-executes every previous launch's schedule in front of the new one: a graph with a single kernel node runs that kernel 6 times over 3 launches (1 + 2 + 3). Kokkos hip_graph.repeat_chain submits one graph ten times; its results stay correct because the chain starts by resetting its counter, but it pays 55 node executions instead of 10.

Mechanism: `CHIPGraphExec::compile()` in `src/CHIPGraph.cc:330` runs on every `hipGraphLaunch` and pushes the new levels onto the member `ExecQueues_` without clearing it; `launch()` then drains the whole queue.

Reproduces on the Intel CPU OpenCL runtime:

```
OCL_ICD_VENDORS=<dir with the intel cpu .icd> CHIP_BE=opencl CHIP_DEVICE_TYPE=cpu ./build/tests/runtime/TestFixKokkosGraphRepeatedLaunch
FAIL: kernel node ran 6 times over 3 launches
```

Reproducer test and fix are on branch `2026-08-28-hipgraph-kokkos` (commits 2496adf430a8, 40419a9c2432).

### LINKED ISSUE #1496: hipGraphNodeSetEnabled and hipGraphNodeGetEnabled are unimplemented (hipErrorNotSupported)
Both entry points are stubs at `src/CHIPBindings.cc:287` and `:297` and return `hipErrorNotSupported`.

Kokkos `Kokkos_CoreUnitTest_HIPInterOpGraph`, test `hip.interact_with_hip_node` (core/unit_test/hip/TestHIP_InterOp_Graph.cpp on develop), disables the kernel node of an instantiated graph, submits, expects the kernel not to have run, re-enables it, submits again and expects it to have run. The first call fails:

```
CHIP error : hipErrorNotSupported (Called a function which is not implemented) in src/CHIPBindings.cc:293:hipGraphNodeGetEnabled
```

Expected per HIP (hip_runtime_api.h: "Disabled nodes are functionally equivalent to empty nodes", "only kernel, memset and memcpy nodes are supported"): the flag belongs to the `hipGraphExec_t`, is addressed through the original graph node handle, a disabled node is a no-op at `hipGraphLaunch`, and `hipGraphNodeGetEnabled` reports 1 or 0. Other node types, null handles, a null output pointer and a node the graph did not hold at instantiation return `hipErrorInvalidValue`.

Reproducer test and implementation are on branch `2026-08-28-hipgraph-node-enabled` (commits 40eda0d05508, b35111dd05c8): the flag lives on the exec graph's cloned node, reached through `CHIPGraphExec::getExecNode()`, and `launch()` skips disabled nodes. Related: #1490 to #1495 (same Kokkos graph work).

### LINKED ISSUE #1498: hipGraphExec*NodeSetParams fail on any graph not made by hipGraphClone and update the original node
The hipGraphExec{Kernel,Memcpy,MemcpyToSymbol,MemcpyFromSymbol,Memset,Host}NodeSetParams and hipGraphExec{EventRecord,EventWait}NodeSetEvent bindings in `src/CHIPBindings.cc` resolve the node with `EXEC(hGraphExec)->getOriginalGraphPtr()->nodeLookup()`, that is in the clone map of the original graph. That map is only populated when the graph was itself produced by `hipGraphClone`, so on a normally built graph the Kernel and FromSymbol variants fail with `hipErrorTbd` and the others with `hipErrorInvalidValue`. Where a node is found, most of them modify the original node rather than the copy held by the executable graph, and `CHIPGraphNodeKernel::setParams` only stored the struct while the exec item kept the arguments bound at construction, so even a correct lookup would not change what runs.

hipamd resolves the node through the executable graph (`GetClonedNode` in `hipamd/src/hip_graph.cpp`) and updates that copy: the original graph is untouched and the next `hipGraphLaunch` uses the new parameters.

Repro: `tests/runtime/TestFixGraphExecKernelNodeSetParams.hip` (one kernel node, instantiate, launch, `hipGraphExecKernelNodeSetParams` with a new argument, launch again, then re-instantiate the original graph; same for a memset node). Before the fix:

```
FAIL: hipGraphExecKernelNodeSetParams(Exec, Kernel, &ParamsB) -> hipErrorTbd
```

Found while implementing `hipGraphNodeSetEnabled` (#1496). Fix on branch `2026-08-28-hipgraph-exec-setparams` (commits 04c3d01cdde0, 10458b7b76a1, ed904b1667ca). `hipGraphExecUpdate` and `hipGraphExecChildGraphNodeSetParams` share the same lookup and are not covered by it. The `Unit_hipGraphExec*NodeSetParams*` entries in `tests/known_failures.yaml` should be retried once this lands.

### LINKED ISSUE #1502: hipLaunchHostFunc on a capturing stream runs the host function immediately instead of recording a host node
`hipLaunchHostFunc` (`src/CHIPBindings.cc:964`) never checks the stream's capture status, so between `hipStreamBeginCapture` and `hipStreamEndCapture` the host function is enqueued as a stream callback and runs right away, while the kernels around it are recorded into the graph. A captured chain of kernel launches and host functions runs its host steps during capture and the graph from `hipStreamEndCapture` replays only the kernels. Kernel launches, memcpys, memsets and event ops all go through `captureIntoGraph()`; `hipLaunchHostFunc` is the one enqueue that does not, so it stays eager even with #1490 fixed.

Backend independent, reproduces on the Intel CPU OpenCL runtime with a K H K H K H chain where each step appends its digit to a `hipMallocManaged` counter:

```
OCL_ICD_VENDORS=<dir with the intel cpu .icd> CHIP_BE=opencl CHIP_DEVICE_TYPE=cpu ./build/tests/runtime/TestFixKokkosGraphMixedHostDeviceNodes
FAIL: captured work ran during capture, counter is 222
```

Found while chasing Kokkos `hip.mixed_then_host_device_nodes` (TestGraph.hpp:1326, 4 instead of 6 on Aurora PVC). That test builds its chain with `hipGraphAddHostNode`, which passes on the CPU runtime, so this is adjacent to the Kokkos failure rather than its cause.

Reproducer and fix (record the call as a `CHIPGraphNodeHost` when the stream is capturing) on branch `2026-08-28-hipgraph-mixed-host-nodes` (commits fdc806099c95, 84b5af748aa2).

pvelesko:
Found running the Kokkos::HIP unit tests (Kokkos::Graph, stream capture) on Aurora PVC. Stream capture executed work eagerly, `hipGraphLaunch` failed on redundant edges and on child graph nodes, repeated launches re-ran the graph N times, `hipGraphGetNodes` segfaulted on a null array, `hipGraphDebugDotPrint` and `hipGraphNodeSetEnabled` were unimplemented, `hipGraphExec*NodeSetParams` updated the wrong graph, and `hipLaunchHostFunc` under capture ran immediately. Each fix has its reproducer test committed before it. Three follow-up commits from review of the series define the kernel node destructor where `chipstar::ExecItem` is complete, make `hipGraphExecHostNodeSetParams` reject null host params, and make child graph nodes own a clone of the child graph as the HIP header documents, which stops a graph added as its own child from recursing at launch.

Fixes #1490
Fixes #1491
Fixes #1492
Fixes #1493
Fixes #1494
Fixes #1495
Fixes #1496
Fixes #1498
Fixes #1502

### DIFF
--- a/src/CHIPBackend.cc
+++ b/src/CHIPBackend.cc
@@ -2191,7 +2191,12 @@ void chipstar::Queue::updateLastNode(CHIPGraphNode *NewNode) {
   LastNode_ = NewNode;
 }
 
-void chipstar::Queue::initCaptureGraph() { CaptureGraph_ = new CHIPGraph(); }
+void chipstar::Queue::initCaptureGraph() {
+  CaptureGraph_ = new CHIPGraph();
+  // The first node recorded into the new graph is a root; a node left over
+  // from an earlier capture on this stream belongs to another graph.
+  LastNode_ = nullptr;
+}
 
 std::shared_ptr<chipstar::Event>
 chipstar::Queue::RegisteredVarCopy(chipstar::ExecItem *ExecItem,
--- a/src/CHIPBackend.hh
+++ b/src/CHIPBackend.hh
@@ -2309,7 +2309,6 @@ public:
    */
   template <class GraphNodeType, class... ArgTypes>
   bool captureIntoGraph(ArgTypes... ArgsPack) {
-    return false; // TODO: fix this in graphs refactor
     if (getCaptureStatus() == hipStreamCaptureStatusActive) {
       auto Graph = getCaptureGraph();
       auto Node = new GraphNodeType(ArgsPack...);
--- a/src/CHIPBindings.cc
+++ b/src/CHIPBindings.cc
@@ -284,13 +284,46 @@ hipError_t hipGraphMemFreeNodeGetParams(hipGraphNode_t node, void *dptr) {
   CHIP_CATCH
 }
 
+/// Resolve hNode to its copy in hGraphExec for hipGraphNodeGetEnabled and
+/// hipGraphNodeSetEnabled. The enabled switch is defined for kernel, memcpy
+/// and memset nodes only; any other type, and a node that was not part of the
+/// graph when hGraphExec was instantiated, is hipErrorInvalidValue.
+static CHIPGraphNode *findEnableableExecNode(hipGraphExec_t hGraphExec,
+                                             hipGraphNode_t hNode) {
+  if (!hGraphExec || !hNode)
+    CHIPERR_LOG_AND_THROW("Null hipGraphExec_t or hipGraphNode_t",
+                          hipErrorInvalidValue);
+
+  auto *ExecNode = EXEC(hGraphExec)->getExecNode(NODE(hNode));
+  if (!ExecNode)
+    CHIPERR_LOG_AND_THROW("Node is not part of the instantiated graph",
+                          hipErrorInvalidValue);
+
+  switch (ExecNode->getType()) {
+  case hipGraphNodeTypeKernel:
+  case hipGraphNodeTypeMemcpy:
+  case hipGraphNodeTypeMemset:
+    break;
+  default:
+    CHIPERR_LOG_AND_THROW(
+        "Only kernel, memcpy and memset nodes can be enabled or disabled",
+        hipErrorInvalidValue);
+  }
+  return ExecNode;
+}
+
 hipError_t hipGraphNodeGetEnabled(hipGraphExec_t hGraphExec,
                                   hipGraphNode_t hNode,
                                   unsigned int *isEnabled) {
   CHIP_TRY
   LOCK(ApiMtx);
   CHIPInitialize();
-  UNIMPLEMENTED(hipErrorNotSupported);
+
+  if (!isEnabled)
+    RETURN(hipErrorInvalidValue);
+
+  *isEnabled = findEnableableExecNode(hGraphExec, hNode)->isEnabled() ? 1 : 0;
+  RETURN(hipSuccess);
   CHIP_CATCH
 }
 
@@ -300,7 +333,9 @@ hipError_t hipGraphNodeSetEnabled(hipGraphExec_t hGraphExec,
   CHIP_TRY
   LOCK(ApiMtx);
   CHIPInitialize();
-  UNIMPLEMENTED(hipErrorNotSupported);
+
+  findEnableableExecNode(hGraphExec, hNode)->setEnabled(isEnabled != 0);
+  RETURN(hipSuccess);
   CHIP_CATCH
 }
 
--- a/src/CHIPBindings.cc
+++ b/src/CHIPBindings.cc
@@ -227,7 +227,14 @@ hipError_t hipGraphDebugDotPrint(hipGraph_t graph, const char *path,
   CHIP_TRY
   LOCK(ApiMtx);
   CHIPInitialize();
-  UNIMPLEMENTED(hipErrorNotSupported);
+  if (!graph || !path)
+    RETURN(hipErrorInvalidValue);
+  std::ofstream Out(path);
+  if (!Out)
+    RETURN(hipErrorOperatingSystem);
+  GRAPH(graph)->writeDot(Out, flags);
+  Out.close();
+  RETURN(Out.good() ? hipSuccess : hipErrorOperatingSystem);
   CHIP_CATCH
 }
 
--- a/src/CHIPBindings.cc
+++ b/src/CHIPBindings.cc
@@ -972,6 +972,14 @@ hipError_t hipLaunchHostFunc(hipStream_t stream, hipHostFn_t fn,
 
   auto ChipQueue = Backend->findQueue(static_cast<chipstar::Queue *>(stream));
 
+  // On a capturing stream the call is recorded as a host node, which runs
+  // the function in dependency order when the graph is launched.
+  hipHostNodeParams Params = {};
+  Params.fn = fn;
+  Params.userData = userData;
+  if (ChipQueue->captureIntoGraph<CHIPGraphNodeHost>(&Params))
+    RETURN(hipSuccess);
+
   ChipQueue->launchHostFunc(fn, userData);
   RETURN(hipSuccess);
   CHIP_CATCH
--- a/src/CHIPBindings.cc
+++ b/src/CHIPBindings.cc
@@ -2302,8 +2302,10 @@ hipError_t hipGraphExecChildGraphNodeSetParams(hipGraphExec_t hGraphExec,
 
   auto CastNode = static_cast<CHIPGraphNodeGraph *>(node);
 
-  // Check if childGraph is not parent graph
-  if (CastNode->getGraph() != childGraph)
+  // The node holds its own clone, so compare topology rather than identity:
+  // the replacement must have the same number of nodes as the embedded graph.
+  if (CastNode->getGraph()->getNodes().size() !=
+      GRAPH(childGraph)->getNodes().size())
     RETURN(hipErrorInvalidValue);
 
   CastNode->setGraph(GRAPH(childGraph));
--- a/src/CHIPBindings.cc
+++ b/src/CHIPBindings.cc
@@ -1246,15 +1246,32 @@ hipError_t hipGraphGetEdges(hipGraph_t graph, hipGraphNode_t *from,
   CHIP_CATCH
 }
 
+/// Shared body of hipGraphGetNodes and hipGraphGetRootNodes. A null Out
+/// array asks for the count only; otherwise *Count is the array's capacity on
+/// entry and the number of entries written on return.
+static hipError_t copyGraphNodes(const std::vector<CHIPGraphNode *> &Nodes,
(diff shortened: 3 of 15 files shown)


pvelesko:
Aurora PVC manual run: PASS. `check.py ./ dgpu level0` at head `939d7391`, 1054/1054 tests passed, 0 failed, 0 not run.

PR #1529, opened 2026-08-29
Capture work forked onto other streams during stream capture

### LINKED ISSUE #1523: Stream capture does not follow event dependencies onto other streams (cross-stream fork/join capture)
Work forked onto another stream during stream capture is not captured. With capture active on `stream1`, `hipEventRecord(ev, stream1)` plus `hipStreamWaitEvent(stream2, ev)` should pull `stream2` into the capture so kernels launched on `stream2` (and joined back with another event) become graph nodes; chipStar records only the kernels launched on `stream1`.

Seen with the catch tests `Unit_hipStreamBeginCapture_nestedStreamCapture` and `Unit_hipStreamBeginCapture_streamReuse` (both listed in tests/known_failures.yaml): after https://github.com/CHIP-SPV/chipStar/pull/1522 stops capture from executing work eagerly (#1490), both fail with `REQUIRE( (*hostMem) == 7 ) with expansion: 2 == 7`. They passed before only because the forked work ran eagerly outside the graph, which was the bug.

Needs: `hipStreamWaitEvent` on a non-capturing stream waiting on an event recorded by a capturing stream must switch that stream into the same capture (CUDA graphs semantics: cross-stream capture via event dependencies), and joining back must add the edges. Found while porting Kokkos::Graph to chipStar on Aurora PVC.

pvelesko:
Stream capture now follows event dependencies onto other streams, following the ROCm clr implementation in `hip_event.cpp`, `hip_stream.cpp` and `hip_graph.cpp`. `hipEventRecord` on a capturing stream tags the event with the stream's current capture dependencies instead of adding a node, `hipStreamWaitEvent` on such an event joins the waiting stream into the same capture (or, for a stream already in it, merges the event's nodes into its dependencies, which is how a fork joins back), and `hipStreamEndCapture` returns `hipErrorStreamCaptureUnmatched` on a forked stream and `hipErrorStreamCaptureUnjoined` when a fork was never joined. The reproducer `TestFix1523CrossStreamCapture` forks two streams, launches one kernel on each of three streams, joins, replays the graph and checks the unjoined error; it failed with `work ran during capture, slots are 0 1 1` before the fix. On OpenCL CPU the fix also makes `Unit_hipStreamBeginCapture_nestedStreamCapture`, `Unit_hipStreamBeginCapture_streamReuse` and 12 further catch graph tests pass with no new failures against the base branch. Stacked on #1522.

Fixes #1523


pvelesko:
Superseded by #1565, which is the same branch rebased onto `main`. This PR was stacked on the now merged #1522 and GitHub refuses to retarget a stacked PR.

PR #1570, opened 2026-09-02
llvm-patches: permit SPV_INTEL_masked_gather_scatter on the HIPSPV device path

### LINKED ISSUE #1454: LLVM23+patches build results in a compile fail
An application on Aurora fails to compile with chipStar on top of LLVM23+patches. From pulling out the .bc of the compile, and running `llvm-reduce` to get a minimal IR, it looks like the issue is from the SPIR-V backend crashing on IR that stores a vector of pointers. 

## Reproducer
```llvm
target triple = "spirv64"

define spir_kernel void @k(ptr addrspace(1) %src, ptr addrspace(1) %dst) {
entry:
  %v = load <4 x ptr addrspace(4)>, ptr addrspace(1) %src, align 8
  store <4 x ptr addrspace(4)> %v, ptr addrspace(1) %dst, align 8
  ret void
}
```

## Compiling

``` bash
clang -cc1 -triple spirv64 -emit-obj bug_min.ll -o /dev/null
```

## Running

The fail looks like:
``` bash
> clang -cc1 -triple spirv64 -emit-obj bug_min.ll -o /dev/null
clang: /lus/flare/projects/Aurora_deployment/bertoni/chip-spv_source-20260807-23-release/llvm-project/llvm/lib/IR/Type.cpp:869: static llvm::FixedVectorType* llvm::FixedVectorType::get(llvm::Type*, unsigned int): Assertion `isValidEleme\
ntType(ElementType) && "Element type of a VectorType must " "be an integer, floating point, " "pointer type, or a valid target " "extension type."' failed.
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace and dumped files.
Stack dump:
0.      Program arguments: clang -cc1 -triple spirv64 -emit-obj bug_min.ll -o /dev/null
1.      Code generation
2.      Running pass 'SPIRV emit intrinsics' on module 'bug_min.ll'.
 #0 0x0000000002567dfa llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) (/home/bertoni/latest/clang-chip-spv/23-20260807-23-release/bin/clang-23+0x2567dfa)
 #1 0x0000000002564958 llvm::sys::RunSignalHandlers() (/home/bertoni/latest/clang-chip-spv/23-20260807-23-release/bin/clang-23+0x2564958)
 #2 0x0000000002564aab SignalHandler(int, siginfo_t*, void*) Signals.cpp:0:0
 #3 0x0000000002564b48 SignalHandlerTerminate(int, siginfo_t*, void*) Signals.cpp:0:0
 #4 0x00007fee8b0bf8c0 __restore_rt (/lib64/libpthread.so.0+0x168c0)
 #5 0x00007fee8a7d8c6b raise (/lib64/libc.so.6+0x4ac6b)
 #6 0x00007fee8a7da305 abort (/lib64/libc.so.6+0x4c305)
 #7 0x00007fee8a7d0c6a __assert_fail_base (/lib64/libc.so.6+0x42c6a)
 #8 0x00007fee8a7d0cf2 (/lib64/libc.so.6+0x42cf2)
 #9 0x0000000001eac08a llvm::FixedVectorType::get(llvm::Type*, unsigned int) (/home/bertoni/latest/clang-chip-spv/23-20260807-23-release/bin/clang-23+0x1eac08a)
#10 0x0000000000daeffd (anonymous namespace)::SPIRVEmitIntrinsics::deduceNestedTypeHelper(llvm::User*, llvm::Type*, llvm::SmallPtrSetImpl<llvm::Value*>&, bool) SPIRVEmitIntrinsics.cpp:0:0
#11 0x0000000000db7cf1 (anonymous namespace)::SPIRVEmitIntrinsics::insertPtrCastOrAssignTypeInstr(llvm::Instruction*, llvm::IRBuilder<llvm::ConstantFolder, llvm::IRBuilderDefaultInserter>&) SPIRVEmitIntrinsics.cpp:0:0
#12 0x0000000000dbbf15 (anonymous namespace)::SPIRVEmitIntrinsics::runOnFunction(llvm::Function&) SPIRVEmitIntrinsics.cpp:0:0
#13 0x0000000000dbe625 (anonymous namespace)::SPIRVEmitIntrinsics::runOnModule(llvm::Module&) SPIRVEmitIntrinsics.cpp:0:0
#14 0x0000000001e34852 (anonymous namespace)::MPPassManager::runOnModule(llvm::Module&) LegacyPassManager.cpp:0:0
#15 0x0000000001e34f6c llvm::legacy::PassManagerImpl::run(llvm::Module&) (/home/bertoni/latest/clang-chip-spv/23-20260807-23-release/bin/clang-23+0x1e34f6c)
#16 0x00000000027f1898 (anonymous namespace)::EmitAssemblyHelper::TimeCodegenPasses(llvm::function_ref<void ()>) BackendUtil.cpp:0:0
#17 0x000000000280574b (anonymous namespace)::EmitAssemblyHelper::emitAssembly(clang::BackendAction, std::unique_ptr<llvm::raw_pwrite_stream, std::default_delete<llvm::raw_pwrite_stream>>, clang::BackendConsumer*) BackendUtil.cpp:0:0
#18 0x0000000002805a80 clang::emitBackendOutput(clang::CompilerInstance&, clang::CodeGenOptions&, llvm::StringRef, llvm::Module*, clang::BackendAction, llvm::IntrusiveRefCntPtr<llvm::vfs::FileSystem>, std::unique_ptr<llvm::raw_pwrite_st\
ream, std::default_delete<llvm::raw_pwrite_stream>>, clang::BackendConsumer*) (/home/bertoni/latest/clang-chip-spv/23-20260807-23-release/bin/clang-23+0x2805a80)
#19 0x0000000002ec6682 clang::CodeGenAction::ExecuteAction() (/home/bertoni/latest/clang-chip-spv/23-20260807-23-release/bin/clang-23+0x2ec6682)
#20 0x0000000003229caa clang::FrontendAction::Execute() (/home/bertoni/latest/clang-chip-spv/23-20260807-23-release/bin/clang-23+0x3229caa)
#21 0x00000000031b20b0 clang::CompilerInstance::ExecuteAction(clang::FrontendAction&) (/home/bertoni/latest/clang-chip-spv/23-20260807-23-release/bin/clang-23+0x31b20b0)
#22 0x0000000003301d1e clang::ExecuteCompilerInvocation(clang::CompilerInstance*) (/home/bertoni/latest/clang-chip-spv/23-20260807-23-release/bin/clang-23+0x3301d1e)
#23 0x0000000000ce9530 cc1_main(llvm::ArrayRef<char const*>, char const*, void*) (/home/bertoni/latest/clang-chip-spv/23-20260807-23-release/bin/clang-23+0xce9530)
#24 0x0000000000cdf897 ExecuteCC1Tool(llvm::SmallVectorImpl<char const*>&, llvm::ToolContext const&, llvm::IntrusiveRefCntPtr<llvm::vfs::FileSystem>) driver.cpp:0:0
#25 0x0000000000ce44e9 clang_main(int, char**, llvm::ToolContext const&) (/home/bertoni/latest/clang-chip-spv/23-20260807-23-release/bin/clang-23+0xce44e9)
#26 0x0000000000c2d549 main (/home/bertoni/latest/clang-chip-spv/23-20260807-23-release/bin/clang-23+0xc2d549)
#27 0x00007fee8a7c324d __libc_start_main (/lib64/libc.so.6+0x3524d)
#28 0x0000000000cded4a _start /home/abuild/rpmbuild/BUILD/glibc-2.31/csu/../sysdeps/x86_64/start.S:122:0
Aborted (core dumped)
```

- We expect this to compile cleanly, as the code is valid IR, based on `llvm-as` and `opt -passes=verify` accepting it, and `llc -mtriple=x86_64-unknown-linux-gnu` compiles the IR cleanly.

- Note that this IR compile with LLVM22 and LLVM21 as well. However, during compile of the application, LLVM21/22 don't generate the vector of pointers in the IR, so thus they never hit this issue when building the application.

I can report this to LLVM if that's preferred, I wanted to do it here first in case possibly this is something chipStar-specific. Let me know if it's chipStar-specific or if I should report to LLVM instead. Thanks!

pvelesko:
Device code that materialises a vector of pointers cannot be emitted as SPIR-V without `SPV_INTEL_masked_gather_scatter`. The HIPSPV driver pins the allowed set with `--spirv-ext=-all,...`, so the extension was not merely absent but switched off, and llvm-spirv refused the module:

```
RequiresExtension: Feature requires the following SPIR-V extension:
 SPV_INTEL_masked_gather_scatter
NOTE: LLVM module contains vector of pointers, translation of which requires this extension
```

Found by building rocPRIM on Aurora against LLVM 23. `test/rocprim/test_device_reduce_by_key.cpp` triggers it, which fails the rocPRIM build and then cascades into rocThrust, rocSPARSE, hipSPARSE and hipMM, and into zeroRK because thrust headers are never installed. Six of fifteen components lost to one defect.

Isolated A/B/C against the reproducer, so the result does not depend on anything else in the toolchain:

```
upstream list                                   rc=18, RequiresExtension
upstream list + masked_gather_scatter           rc=0, spirv-val VALID
upstream list + an unrelated extension only     rc=18, RequiresExtension
```

IGC accepts the emitted SPIR-V on PVC (`ocloc compile -spirv_input -device pvc` reports `Build succeeded`).

End to end on Aurora PVC, LLVM `llvmorg-23.1.0-rc2`, `-DCHIP_LLVM_USE_INTERGRATED_SPIRV=OFF`, rebuilding the whole suite with the extension permitted:

```
components built      10/15 -> 14/15
rocPRIM               builds, 9/9 tests
rocThrust             builds, 76/76 tests
zeroRK                4/4 CPU vs GPU correctness checks
chipStar unit tests   1063/1063
```

This does NOT close https://github.com/CHIP-SPV/chipStar/issues/1454. The in-tree SPIR-V backend still aborts on the same input in `SPIRVEmitIntrinsics::deduceNestedTypeHelper` during type deduction, before any extension gating, so no flag reaches it. I confirmed the native configuration still fails identically with the extension permitted. The issue should stay open for the LLVM side.

The test has two halves because the driver does not expose its extension list through `-###` or `-v`, and no small HIP source reproduces the IR pattern. Adjacent pointer copies, indexed gathers and scatters and an array of pointers dereference were each compiled and none produced a vector of pointer type in the lowered device IR, so the `.ll` from the issue is used instead. Half one asserts the driver patch still permits the extension, which is what regresses. Half two runs llvm-spirv and spirv-val to show that this is still the extension that makes the difference, and fails if translation ever succeeds without it, so the test retires itself when the extension stops being needed.

One disclosure about the environment the end to end numbers came from. The shared LLVM source tree on Aurora used for those builds was edited by another party partway through, adding `+SPV_INTEL_cache_controls` to the same two lists. That edit is not in this PR and is not in chipStar main. The A/B/C isolation above was run separately through llvm-spirv precisely so the claim does not rest on that build.

### DIFF
--- a/llvm-patches/llvm-23/llvm/0001-hipspv-in-tree-spirv-backend.patch
+++ b/llvm-patches/llvm-23/llvm/0001-hipspv-in-tree-spirv-backend.patch
@@ -124,7 +124,12 @@ index d6900c7..98bec18 100644
 +                       ",+SPV_INTEL_function_pointers"
 +                       ",+SPV_INTEL_subgroups"
 +                       ",+SPV_KHR_bit_instructions"
-+                       ",+SPV_EXT_shader_atomic_float_add");
++                       ",+SPV_EXT_shader_atomic_float_add"
++                       // A vector of pointers (e.g. rocPRIM's
++                       // test_device_reduce_by_key) is untranslatable without
++                       // this: llvm-spirv refuses with RequiresExtension. IGC
++                       // accepts the result (ocloc -device pvc).
++                       ",+SPV_INTEL_masked_gather_scatter");
 +
 +      // Preserve debug info in the NonSemantic.Shader.DebugInfo form.
 +      // These flags are passed unconditionally instead of gating on -g: in
@@ -170,7 +175,8 @@ index d6900c7..98bec18 100644
 +                      ",+SPV_KHR_bit_instructions"
 +                      ",+SPV_EXT_shader_atomic_float_add"
 +                      ",+SPV_KHR_non_semantic_info"
-+                      ",+SPV_INTEL_optnone");
++                      ",+SPV_INTEL_optnone"
++                      ",+SPV_INTEL_masked_gather_scatter");
 +
 +    Cc1Args.push_back(TempFile);
 +    Cc1Args.push_back("-o");
--- a/llvm-patches/llvm-23/llvm/0001-hipspv-in-tree-spirv-backend.patch
+++ b/llvm-patches/llvm-23/llvm/0001-hipspv-in-tree-spirv-backend.patch
@@ -84,7 +84,7 @@ index d6900c7..98bec18 100644
  void HIPSPV::Linker::constructLinkAndEmitSpirvCommand(
      Compilation &C, const JobAction &JA, const InputInfoList &Inputs,
      const InputInfo &Output, const llvm::opt::ArgList &Args) const {
-@@ -73,44 +102,100 @@ void HIPSPV::Linker::constructLinkAndEmitSpirvCommand(
+@@ -73,44 +102,106 @@ void HIPSPV::Linker::constructLinkAndEmitSpirvCommand(
    tools::constructLLVMLinkCommand(C, *this, JA, Inputs, LinkArgs, Output, Args,
                                    TempFile);
  
@@ -224,7 +224,7 @@ index d6900c7..98bec18 100644
  
    InputInfo TrInput = InputInfo(types::TY_LLVM_BC, TempFile, "");
    SPIRV::constructTranslateCommand(C, *this, JA, Output, TrInput, TrArgs);
-@@ -150,6 +235,11 @@ HIPSPVToolChain::HIPSPVToolChain(const Driver &D, const llvm::Triple &Triple,
+@@ -150,6 +241,11 @@ HIPSPVToolChain::HIPSPVToolChain(const Driver &D, const llvm::Triple &Triple,
    getProgramPaths().push_back(getDriver().Dir);
  }
  
--- a/tests/compiler/CMakeLists.txt
+++ b/tests/compiler/CMakeLists.txt
@@ -172,6 +172,7 @@ add_shell_test(TestPrintfStaticString.bash)
 # fails. This skips when the in-tree SPIR-V backend is used
 # (LLVM_SPIRV=NOT_NEEDED), since the patch targets the external translator.
 add_shell_test(TestSpirvDuplicatePhiHip.bash)
+add_shell_test(TestFix1454VectorOfPointers.bash)
 
 add_shell_test(TestTextureUnsupportedMixedDim.bash)
 
--- a/tests/compiler/TestFix1454VectorOfPointers.bash
+++ b/tests/compiler/TestFix1454VectorOfPointers.bash
@@ -1,0 +1,69 @@
+#!/bin/bash
+# Regression test for CHIP-SPV/chipStar#1454.
+#
+# Device IR that materialises a vector of pointers can only be translated to
+# SPIR-V when SPV_INTEL_masked_gather_scatter is permitted. chipStar pins the
+# allowed set in the HIPSPV driver with `--spirv-ext=-all,...`, so the
+# extension is not merely absent by default, it is switched off, and
+# llvm-spirv rejects such a module with
+#   RequiresExtension: ... SPV_INTEL_masked_gather_scatter
+# rocPRIM's test/rocprim/test_device_reduce_by_key.cpp is a real trigger:
+# without the extension the whole rocPRIM build fails and rocThrust,
+# rocSPARSE, hipSPARSE, hipMM and zeroRK fail after it.
+#
+# The driver does not expose its extension list through -### or -v, and no
+# small HIP source reliably produces a vector of pointers (adjacent pointer
+# copies, indexed gathers and scatters and an array-of-pointers dereference
+# were all tried and none vectorised), so the guard has two halves:
+#   1. the chipStar patch that sets the list must still list the extension
+#   2. the extension must still be what makes such a module translatable
+set -eu
+
+SRC_DIR="@CMAKE_CURRENT_SOURCE_DIR@"
+LLVM_SPIRV="@LLVM_SPIRV@"
+# llvm-as sits beside llvm-spirv in the same LLVM bin directory.
+LLVM_AS="$(dirname "@LLVM_SPIRV@")/llvm-as"
+SPIRV_VAL="@CMAKE_BINARY_DIR@/external/spirv-tools/bin/spirv-val"
+PATCH="@CMAKE_SOURCE_DIR@/llvm-patches/llvm-@LLVM_VERSION_MAJOR@/llvm/0001-hipspv-in-tree-spirv-backend.patch"
+OUT="@CMAKE_CURRENT_BINARY_DIR@/@TEST_NAME@.d"
+
+# Part 1: the driver patch must still permit the extension.
+if [ ! -f "${PATCH}" ]; then
+  echo "no HIPSPV driver patch for LLVM @LLVM_VERSION_MAJOR@; skipping"
+  exit 0
+fi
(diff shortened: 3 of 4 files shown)


pvelesko:
Correction to the IGC claim in the description, found by reproducing this on a local Arc B570 box rather than only on Aurora.

The LLVM half is not Aurora specific at all. The same reproducer on a local machine with the same clang 23.1.0-rc2 (`561093d94eb7`) gives byte identical results:

```
in-tree backend        Assertion `isValidElementType' in 'SPIRV emit intrinsics'
llvm-spirv, no ext     RequiresExtension: SPV_INTEL_masked_gather_scatter
llvm-spirv, with ext   rc=0, 404 byte module (same size as on Aurora)
```

The IGC half does differ, and not in the direction the description implies. `ocloc compile -spirv_input` on the translated module:

```
Aurora PVC, IGC 2.11.29 (agama 25.18)   Build succeeded
local, IGC 2.38.2, -device bmg          InvalidModule: ... uses extension 'SPV_INTEL_masked_gather_scatter' which were disabled by --spirv-ext option
local, IGC 2.38.2, -device dg2          same
local, IGC 2.38.2, -device adl-s        same
```

So the older IGC on Aurora accepts the extension and the current IGC rejects it. Practical effect of this PR by platform:

on Aurora PVC, device code holding a vector of pointers now compiles and runs, which is what took the suite from 10 of 15 components to 14 of 15

on current IGC, such code still does not work, but the failure moves from a translation error at build time to an InvalidModule from IGC. That is arguably worse for diagnosis even though nothing that previously worked is affected, since permitting an extension does not make anything use it

I am leaving the change as is because it is a strict improvement for PVC and neutral elsewhere, but reviewers should know the extension is not universally consumable, and a follow up may want to gate it or to ask IGC why a newer version dropped support.

pvelesko:
Closing this in favour of a portable fix.

Permitting `SPV_INTEL_masked_gather_scatter` works on Aurora, but it is not the right fix, because the extension is not consumable everywhere. Measured with `ocloc compile -spirv_input` on the module this PR makes translatable:

```
Aurora PVC, IGC 2.11.29    -device pvc     Build succeeded
local, IGC 2.38.2          -device bmg     InvalidModule: uses extension 'SPV_INTEL_masked_gather_scatter' which were disabled by --spirv-ext option
local, IGC 2.38.2          -device dg2     same
local, IGC 2.38.2          -device adl-s   same
```

So on current IGC the module still fails, only later and with a worse diagnostic: translation succeeds and IGC then refuses the binary. That is not an improvement for anything except the one older IGC on Aurora.

The better fix is to stop producing a vector of pointers at all. LLVM 23's SROA folds rocPRIM's struct of three pointers into `<3 x ptr addrspace(4)>`, and a vector of integers is ordinary SPIR-V needing no extension, so the values can be carried as `<3 x i64>` instead. That works on every target rather than one, and it is the same class of fix as the existing `HipLowerOverflowIntrinsics`, `HipPromoteInts` and `HipFunctionPointerAS` passes.

Replacement PR to follow, along with a tracking issue for reverting it once the SPIR-V backend handles this itself. The underlying LLVM 23 regression stays tracked in https://github.com/CHIP-SPV/chipStar/issues/1454.

PR #1585, opened 2026-09-03
hip_fp16: stop specializing std::is_floating_point for _Float16

### LINKED ISSUE #1582: spirv_hip_fp16.h specializes std::is_floating_point, which libc++ >= 20 rejects (6 test failures)
## Summary

`include/hip/spirv_hip_fp16.h:80` opens `namespace std` and specializes
`std::is_floating_point` for `_Float16`. libc++ marks that trait
`[[_Clang::__no_specializations__]]`, which makes the specialization a hard error. Six
tests fail as a result.

**This is not macOS-specific.** The attribute is upstream libc++, not an Apple addition —
`libcxx/include/__type_traits/is_floating_point.h` carries it on `release/20.x`,
`release/21.x` and `main`, and not on `release/19.x`. And in `libcxx/include/__config` it is
gated on compiler support alone, with no platform condition:

```cpp
#if __has_cpp_attribute(_Clang::__no_specializations__)
#  define _LIBCPP_NO_SPECIALIZATIONS \
     [[_Clang::__no_specializations__("Users are not allowed to specialize this standard library entity")]]
#else
#  define _LIBCPP_NO_SPECIALIZATIONS
#endif
```

So the trigger is **libc++ >= 20 plus a clang that supports the attribute, on any
platform** — a Linux build with `-stdlib=libc++` would hit it too. It surfaces on macOS
first only because libc++ is the default there. GNU libstdc++ has no equivalent diagnostic
(zero occurrences in its `<type_traits>`), which is why the usual Linux configuration stays
quiet — but the code is UB either way: `std::is_floating_point` is not a customization
point, and `[namespace.std]` permits specializing a standard template only for
program-defined types.

The specialization is unguarded, chipStar passes no suppression flag of its own, and none
of the six tests are listed in `tests/known_failures.yaml` — so I don't think this is
currently visible to chipStar CI.

## The code

`include/hip/spirv_hip_fp16.h`, ~line 73-81:

```cpp
#if defined(__cplusplus)
  #if !defined(__HIPCC_RTC__)
    #include "spirv_math_fwd.h"
    #include "spirv_hip_vector_types.h"
  #endif
    namespace std
    {
        template<> struct is_floating_point<_Float16> : std::true_type {};   // line 80
    }
```

libc++ (`__type_traits/is_floating_point.h:30`, upstream and in the SDK copy I tested):

```cpp
struct _LIBCPP_NO_SPECIALIZATIONS is_floating_point : ... {};
```

where `_LIBCPP_NO_SPECIALIZATIONS` expands to
`[[_Clang::__no_specializations__("Users are not allowed to specialize this standard library entity")]]`
whenever the compiler supports the attribute (see the `__config` excerpt above).

## Error

```
include/hip/spirv_hip_fp16.h:80:31: error: 'is_floating_point' cannot be specialized:
  Users are not allowed to specialize this standard library entity [-Winvalid-specialization]
   80 |             template<> struct is_floating_point<_Float16> : std::true_type {};
      |                               ^
```

## Failing tests

`scripts/check.py ./ pocl opencl` → 7 failures / 378. Six share this cause:

- `hipcc-TestFix797ShflHalf`
- `TestHipccHalfConversions`
- `TestHipccHalfOperators`
- `hipcc-TestFix1248HalfHostOps`
- `hipcc-TestHalfDeviceMathFuncs`
- `TestHipccFp16Include`

## Reproduction

Note a bare `hipcc -x hip <file>` against an *installed* tree does not reproduce — the
include paths differ. Use the build tree, i.e. the same invocation ctest uses:

```bash
R=<chipStar source dir>
$LLVM/bin/clang++ \
  -include $R/include/hip/spirv_fixups.h -I/$R/build/include \
  -D__HIP_PLATFORM_SPIRV__= --offload=spirv64v1.3-unknown-chipstar \
  -nohipwrapperinc --hip-path=$R/build \
  --target=arm64-apple-darwin25.6.0 --sysroot=$(xcrun --show-sdk-path) \
  -Xarch_device -I$R/include/hip/devicelib/macOS \
  -I$R/include -I$R/HIP/include -I$R/build/include \
  -Xoffload-compiler-spirv64v1.3-unknown-chipstar -fno-integrated-objemitter \
  -x hip $R/tests/compiler/inputs/testfp16include.cpp -o /dev/null
```

That emits the error; appending `-Wno-invalid-specialization` clears it and the compile
proceeds to link.

## Why CI probably doesn't see it

Two reasons, and only the first is macOS-shaped:

1. **The macOS runner's libc++ is likely too old.** The attribute is present at
   `_LIBCPP_VERSION 210106` (SDK 26.5) and absent at `190102` (SDK 15.4), so the runner in
   `.github/workflows/unit-tests-macos-pocl.yml` probably has a pre-26 SDK. Check on that
   host with:

   ```bash
   grep -c _LIBCPP_NO_SPECIALIZATIONS \
     "$(xcrun --show-sdk-path)/usr/include/c++/v1/__type_traits/is_floating_point.h"
   ```

   Non-zero means it would reproduce.

2. **The Linux runners use libstdc++**, which has no such diagnostic — so they stay green
   regardless of libc++ version. They would start failing if anyone built with
   `-stdlib=libc++` against libc++ >= 20.

That combination is why this looks platform-specific when the underlying problem is not.

## Why `-Wno-invalid-specialization` isn't the fix

It works for chipStar's own build (I currently pass it locally), but it does not reach
tests or user code that drive `hipcc` directly — which is exactly why the six tests still
fail. It also suppresses a diagnostic that is telling the truth: `std::is_floating_point`
is not a customization point, and specializing it is UB.

### (1) stop specializing the standard trait, and use a chipStar-local one at the call
sites that depend on it:

```cpp
namespace chipstar {
template <class T> struct is_floating_point : std::is_floating_point<T> {};
template <> struct is_floating_point<_Float16> : std::true_type {};
}
```

### (2) Minimal alternative, if there are too many call sites to change safely:

```cpp
#if !defined(_LIBCPP_NO_SPECIALIZATIONS)
namespace std { template<> struct is_floating_point<_Float16> : std::true_type {}; }
#endif
```

Note the guard alone changes behaviour on new libc++ (the trait then reports `false` for
`_Float16`), so whatever depends on it needs checking — which is the argument for the
local-trait approach.

## Environment

- macOS 26.6.2, Apple Silicon (M4)
- Xcode 26.6 (Build 17F113) — note its **SDK is 26.5**
  (`xcrun --show-sdk-version` → `26.5`)
- **The relevant version is the SDK's libc++: `_LIBCPP_VERSION 210106`**, in
  `MacOSX.sdk/usr/include/c++/v1`. That is what carries the attribute.
- LLVM 23.1.0-rc2 (`561093d9`), built via `scripts/configure_llvm.sh --version 23`, translator variant
- chipStar `main` @ `8f44c604`, PoCL `main` @ `515308fb`

The sysroot in use is pinned three independent ways, so there is no ambiguity about which
libc++ produced the error: the build's `clang.cfg`/`clang++.cfg`, the `--sysroot` chipStar
bakes into `hip-config.cmake` from `CMAKE_OSX_SYSROOT` (which is what `hipcc` actually
passes), and the compiler's own `note:` line in the failure output — all name
`.../Xcode.app/.../SDKs/MacOSX.sdk`.

Not Xcode-specific: the Command Line Tools `MacOSX26.5.sdk` ships the same libc++
(`_LIBCPP_VERSION 210106`) and behaves identically. Only an older SDK differs — CLT's
leftover `MacOSX15.4.sdk` has `_LIBCPP_VERSION 190102` and no such attribute, and the error
does not occur there.

pvelesko:
`spirv_hip_fp16.h` opened `namespace std` and specialized `std::is_floating_point<_Float16>`. That is UB: `[namespace.std]` allows specializing a standard template only for a program-defined type, and `_Float16` is not one. libc++ >= 20 marks the trait `[[clang::no_specializations]]`, which turns it from UB into a hard error, so any build against libc++ >= 20 fails on any platform. It shows up on macOS first only because libc++ is the default there.

The specialization is replaced with a chipStar-local trait in `hip_impl`, and the three `__half` members that consumed it (the converting constructor, `operator=`, and `operator T()`) now use that trait. The standard trait is left exactly as the standard library defines it.

Reproducer `TestFix1582StdTraitSpecialization.hip` asserts that `std::is_floating_point<_Float16>` is unspecialized and that the `__half` conversions the specialization used to enable still compile. It fails on the tests-only commit and passes on the fix commit. Asserting on the trait rather than on the libc++ diagnostic makes it fire on every platform, including the libstdc++ CI lanes, which the diagnostic itself never reaches.

Verified on Intel Arc B570, LLVM 22.1.0, level0 backend. Full `check.py igpu level0` run on the baseline header and on the fixed header gives byte-identical failure sets (445 vs 444 of 1671); the single difference is the new test, which fails on baseline and passes with the fix. The six tests named in the issue all pass here: `hipcc-TestFix797ShflHalf`, `TestHipccHalfConversions`, `TestHipccHalfOperators`, `hipcc-TestFix1248HalfHostOps`, `hipcc-TestHalfDeviceMathFuncs`, `TestHipccFp16Include`.

No libc++ is installed on the development machine, so the `-Winvalid-specialization` error itself was not reproduced locally, only the root cause it stems from. Confirmation on the macOS lane with an SDK carrying `_LIBCPP_VERSION >= 200000` is still wanted.

One behavior consequence worth naming: `is_valid_type` in `spirv_hip_cooperative_groups_helper.h` uses `std::is_floating_point<T>`, so a cooperative-groups reduction over a raw `_Float16` was accepted when `hip_fp16.h` happened to be included first, and is now rejected. That acceptance was already include-order dependent and `__half` was never accepted there, so it is left alone rather than widened in this PR.

Fixes #1582

### DIFF
--- a/include/hip/spirv_hip_fp16.h
+++ b/include/hip/spirv_hip_fp16.h
@@ -75,8 +75,15 @@ THE SOFTWARE.
         #include "spirv_math_fwd.h"
         #include "spirv_hip_vector_types.h"
       #endif
-        namespace std
+        // std::is_floating_point is not a customization point: [namespace.std]
+        // only allows specializing a standard template for a program-defined
+        // type, and libc++ >= 20 marks the trait
+        // [[clang::no_specializations]], which turns the specialization into a
+        // hard error. Use a chipStar-local trait instead (issue #1582).
+        namespace hip_impl
         {
+            template<typename T>
+            struct is_floating_point : std::is_floating_point<T> {};
             template<> struct is_floating_point<_Float16> : std::true_type {};
         }
 
@@ -103,7 +110,7 @@ THE SOFTWARE.
                 __half(decltype(data) x) : data{x} {}
                 template<
                     typename T,
-                    Enable_if_t<std::is_floating_point<T>{}>* = nullptr>
+                    Enable_if_t<hip_impl::is_floating_point<T>{}>* = nullptr>
                 __HOST_DEVICE__
                 __half(T x) : data{static_cast<_Float16>(x)} {}
             #endif
@@ -162,7 +169,7 @@ THE SOFTWARE.
             #if !defined(__HIP_NO_HALF_CONVERSIONS__)
                 template<
                     typename T,
-                    Enable_if_t<std::is_floating_point<T>{}>* = nullptr>
+                    Enable_if_t<hip_impl::is_floating_point<T>{}>* = nullptr>
                 __HOST_DEVICE__
                 __half& operator=(T x)
                 {
@@ -232,7 +239,7 @@ THE SOFTWARE.
             #if !defined(__HIP_NO_HALF_CONVERSIONS__)
                 template<
                     typename T,
-                    Enable_if_t<std::is_floating_point<T>{}>* = nullptr>
+                    Enable_if_t<hip_impl::is_floating_point<T>{}>* = nullptr>
                 __HOST_DEVICE__
                 operator T() const { return data; }
             #endif
--- a/tests/compiler/CMakeLists.txt
+++ b/tests/compiler/CMakeLists.txt
@@ -130,6 +130,8 @@ add_hipcc_test(TestFix1248HalfHostOps.hip HIPCC_OPTIONS -fsyntax-only)
 
 add_hipcc_test(TestHalfDeviceMathFuncs.hip HIPCC_OPTIONS -fsyntax-only)
 
+add_hipcc_test(TestFix1582StdTraitSpecialization.hip HIPCC_OPTIONS -fsyntax-only)
+
 add_hipcc_test(TestRocmVersionHeader.hip HIPCC_OPTIONS -fsyntax-only)
 
 add_hipcc_test(TestZeroLenArrayTypes.hip TEST_NAME TestZeroLenArrayTypes-O0
--- a/tests/compiler/TestFix1582StdTraitSpecialization.hip
+++ b/tests/compiler/TestFix1582StdTraitSpecialization.hip
@@ -1,0 +1,31 @@
+// Reproduces issue #1582: hip_fp16.h must not specialize std::is_floating_point
+// for _Float16. That is UB ([namespace.std] allows specializing a standard
+// template only for program-defined types) and libc++ >= 20 marks the trait
+// [[clang::no_specializations]], turning it into a hard error.
+#include <hip/hip_fp16.h>
+#include <type_traits>
+
+// The standard trait must be left exactly as the standard library defines it.
+static_assert(!std::is_floating_point<_Float16>::value,
+              "chipStar must not specialize std::is_floating_point<_Float16>");
+
+// The __half conversions that the specialization used to enable must keep
+// working through a chipStar-local trait.
+__host__ __device__ void half_conversions() {
+    _Float16 f16 = 1.5;
+    __half h{f16};
+    h = f16;
+    _Float16 back = h;
+
+    __half hf{1.5f};
+    hf = 2.5f;
+    float f = hf;
+
+    __half hd{1.5};
+    hd = 2.5;
+    double d = hd;
+
+    (void)back; (void)f; (void)d;
+}
+
+int main() { return 0; }


pvelesko:
Verified on macOS ARM64 (pastrami, Apple Silicon, macOS 15.3.1, LLVM 22.1.0, PoCL/OpenCL).

That machine's SDK is 15.5, `_LIBCPP_VERSION 190102`, which has no `_LIBCPP_NO_SPECIALIZATIONS` at all, so it is the "runner's libc++ is too old" case predicted in the issue and it cannot hit the error on its own. No SDK newer than 15.5 and no in-tree libc++ is installed on it.

To get real coverage of the libc++ >= 20 condition, the SDK's libc++ 19 header tree was copied and `__type_traits/is_floating_point.h` was patched with the same attribute upstream libc++ added between `release/19.x` and `release/20.x`, then used with `-nostdinc++ -isystem`. The emulation was validated first on a standalone TU: specializing the trait under it produces the diagnostic, and clang 22 reports `__has_cpp_attribute(_Clang::__no_specializations__)` as supported.

Baseline header against that libc++, through the build tree's `hipcc`, gives exactly the failure in the issue report:

```
/tmp/cs1582/include/hip/spirv_hip_fp16.h:80:31: error: 'is_floating_point' cannot be specialized:
  Users are not allowed to specialize this standard library entity [-Winvalid-specialization]
   80 |             template<> struct is_floating_point<_Float16> : std::true_type {};
      |                               ^
```

With the fix, the same command is clean, and all six tests named in the issue compile against the patched libc++:

```
TestFix1582StdTraitSpecialization.hip    PASS
TestFix797ShflHalf.hip                   PASS
TestFix1248HalfHostOps.hip               PASS
TestHalfDeviceMathFuncs.hip              PASS
inputs/testfp16include.cpp               PASS
TestHipccHalfConversions.cc -D__HIP_NO_HALF_CONVERSIONS__=1   PASS
TestHipccHalfConversions.cc -D__HIP_NO_HALF_OPERATORS__=1     PASS
```

Separately, against the stock SDK libc++ 19, the tree configures and builds clean on macOS and the fp16 test set was run on the baseline header and on the fixed header in the same build. Baseline: 5 of 11 fail. Fixed: 4 of 11 fail. The only difference is the new test. The four constant failures are `fp16`, `fp16_math`, `fp16_half2_math` and `TestBFloat16AsCasts`, all of which execute a kernel and fail in PoCL's JIT with `ld: library 'System' not found`, a pre-existing environment problem on that host unrelated to this change.

Noerr:
Confirmed on a real libc++ >= 20 SDK — the case your emulation was standing in for. Thank
you for the very fast turnaround on both this and #1584, and for going to the trouble of
patching a copy of the libc++ 19 header to get coverage; the emulation was faithful, the
results match exactly.

**Environment** — macOS 26.6.2, Apple Silicon (M4), Xcode 26.6, SDK **26.5**,
`_LIBCPP_VERSION 210106` (the genuine `_LIBCPP_NO_SPECIALIZATIONS`, not a patched header).
LLVM 23.1.0-rc2, PoCL `main`, OpenCL backend, translator lane.

**All seven pass**, including the new reproducer:

```
hipcc-TestFix1582StdTraitSpecialization   Passed
hipcc-TestFix797ShflHalf                  Passed
TestHipccHalfConversions                  Passed
TestHipccHalfOperators                    Passed
hipcc-TestFix1248HalfHostOps              Passed
hipcc-TestHalfDeviceMathFuncs             Passed
TestHipccFp16Include                      Passed
100% tests passed out of 7
```

**Full suite, same tree:**

| | tests | result |
|---|---|---|
| before (chipStar `8f44c604`) | 378 | 98% — 7 failures |
| after (`9ad35e7c`, #1584 + #1585) | 385 | **100% — 0 failures** |

The six fp16 tests account for six of the seven. The seventh,
`Unit_Device_Complex_hipCfma_Device_Sanity_Positive - hipDoubleComplex_ComplexTest`
(SEGFAULT), also no longer fails; I had it flagged as possibly flaky and never triaged it,
so I would not credit it to either PR without more evidence.

Two notes that may be useful to you:

- Your `ld: library 'System' not found` failures on pastrami are the macOS sysroot problem
  — PoCL's JIT invokes the `clang` driver at runtime and a custom-built LLVM has no default
  sysroot. Writing `--sysroot=$(xcrun --show-sdk-path)` into `clang.cfg` *and* `clang++.cfg`
  next to the clang binaries fixes it (both are needed: PoCL's JIT uses `clang`, hipcc uses
  `clang++`). See llvm#137352 for the upstream discussion.
- The cooperative-groups consequence you named — `is_valid_type` no longer accepting a raw
  `_Float16` — did not surface anywhere in this run.

Issue #1594, opened 2026-09-04
macOS PoCL: hipMemcpy segfaults inside pocl_svm_memcpy_common under parallel ctest


pvelesko:
On the macOS PoCL lane, `Unit_Device_Complex_hipCfma_Device_Sanity_Positive` segfaults intermittently. It failed the run below at 384 of 385 tests passing, and it also passed the identical tree twice on the same machine, so it is load dependent rather than deterministic.

Failing job: https://github.com/CHIP-SPV/chipStar/actions/runs/33891841952/job/101085122078

```
The following tests FAILED:
	171 - Unit_Device_Complex_hipCfma_Device_Sanity_Positive - hipFloatComplex_ComplexTest (SEGFAULT)
```

The crash is inside PoCL, reached from a plain `hipMemcpy`:

```
libpocl.2.16.0.dylib   pocl_svm_memcpy_common
libpocl.2.16.0.dylib   clEnqueueSVMMemcpy
libOpenCL.1.0.0.dylib  clEnqueueSVMMemcpy
libCHIP.dylib          CHIPQueueOpenCL::memCopyAsyncImpl(void*, void const*, unsigned long, hipMemcpyKind)
libCHIP.dylib          chipstar::Queue::memCopy(void*, void const*, unsigned long, hipMemcpyKind)
libCHIP.dylib          hipMemcpyInternal(void*, void const*, unsigned long, hipMemcpyKind)
libCHIP.dylib          hipMemcpy
ComplexTest            void ComplexFunctionTernaryDeviceTest<HIP_vector_type<double, 2u>>(...)
ComplexTest            Catch::TestCase::invoke() const
```

macOS records it as `EXC_BAD_ACCESS`, `KERN_INVALID_ADDRESS at 0x0000000000000070`, a null dereference at offset `0x70`. The process died 48 ms after launch.

Reproduction on the runner, from the CI build tree:

```
ctest -R "hipCfma_Device_Sanity_Positive" -j8 --repeat until-fail:40
```

That fails within a few dozen iterations, on either the `hipFloatComplex` or the `hipDoubleComplex` variant. The same test run serially passes 10 times out of 10, so concurrency is what exposes it.

[...]

PR #1595, opened 2026-09-05
tests: mask the PoCL SVM memcpy crash on pastrami

pvelesko:
Both `hipCfma` device sanity variants segfault inside PoCL's `pocl_svm_memcpy_common` on a plain `hipMemcpy`, about one process in sixty, which has reddened the macOS lane on three consecutive runs and blocks every pull request. The fault is a null dereference inside PoCL on input the OpenCL spec allows, not a chipStar call error. Issue #1594 carries the backtrace, the reproduction and the spec quote.

The mask is scoped to the `pastrami` host so the Linux PoCL lane still reports the test. Delete both entries when the PoCL fix lands.

### DIFF
--- a/tests/known_failures.yaml
+++ b/tests/known_failures.yaml
@@ -931,3 +931,15 @@ chipstar-rusticl: # AMD Radeon Pro W6400 via rusticl/radeonsi (self-hosted runne
   OPENCL_CPU:
   OPENCL_GPU:
   OPENCL_POCL:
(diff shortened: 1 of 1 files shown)


pvelesko:
Superseded: the known_failures entries moved into https://github.com/CHIP-SPV/chipStar/pull/1593 so the disk reap change and its mask land together.

